PR #266, opened 2016-10-21
Use https in links were possible

mbeijen:
I updated zammad to use https links to zammad.org where possible.

There were links to forums which do not exist. I changed them to links to the /participate part of the web site.

The download link to the tarball in the OS X documentation was broken, I fixed that.
But I also noticed the software download - where you'd really want it - is served from ftp.zammad.com which does not have https. Can you maybe enable that?

### DIFF
--- a/README_Install_Mac.md
+++ b/README_Install_Mac.md
@@ -15,9 +15,9 @@ Get Zammad
 ----------
 
 ````shell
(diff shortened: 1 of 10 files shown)


martini:
Nice! ❤️

martini:
Thanks, https://ftp.zammad.com/ is now also available.

PR #288, opened 2016-10-24
Add ruby version  as recommended in the docs.

Libermentix:
I added a .ruby-version file for everyone that wants to use rbenv. It doesn't hurt so I though i send a pull request.


martini:
Thank you! ❤️

PR #291, opened 2016-10-24
add support for unicorn

Libermentix:
i have added support for unicorn. 

also up for discussion would be the use of rbenv-vars. This would make configuration management a lot easier (the only file that needs to be deployed is a .rbenv-vars in the root folder of the application).

If you want i can also add that pattern to the database part, because imho it makes live so much easier ;-)


martini:
@Libermentix thanks for the pull request, currently we are just busy with bug fixing and next release. So I'll take a closer look later.

monotek:
I guess adding a database.yml would brake database creation of our packages because the contrib/packager.io/postinstall.sh checks for this file and creates/updates the db when found/not found.

monotek:
@martini 

tried unicorn but removed the rbenv and database.yml files from this pull: 

* https://github.com/monotek/zammad/tree/unicorn

start via:

* bundle exec unicorn -p 3000 -c config/unicorn.rb

works for me but i'm still not sure whats de benefit of unicorn @Libermentix?

if i google for puma vs unicorn performance  comparison puma wins.

Libermentix:
@monotek i think the answer would be choice for deployment. since some prefer unicorn over puma to simplify/unify their deployment scripts. but again, that is also personal flavor.

monotek:
What i saw several times now is, that after finishing the installation wizard the wizard is shown again.
Reloading page with F5 helps.
Did you experienced the same?

Libermentix:
it worked fine for me. i installed it, we went through the configuration screens and then started using it.

monotek:
@Libermentix 
Saw this several times. Maybe you could test it again too?

@martini 
Is Zammad thread safe out of the box?

monotek:
I've added the unicorn support in https://github.com/zammad/zammad/commit/8c29d9fb550eed337429f79b17ba10e56af01750 now, without the rbenv stuff.

Issue #304, opened 2016-10-26
miss package gawk on Debian


NeverMin:
Updating systemzammad password required for 'apt-get --quiet --yes update': 
.....
Installing required packages: gawk......

monotek:
@NeverMin 
At which step do you needed gawk?
Which Debian Version did you used?

NeverMin:
Hi @monotek 
rvm install 2.3.1 on Debian 7.

monotek:
Thanks for the info.
Would be nice you could rename to file to "README_Install_Debian_Ubuntu.md" and also change the title to "Installation on Ubuntu 16.04 Server / Debian 7"

NeverMin:
Got it and thank you for your suggestion, I will do that

monotek:
No need to close you pull request. You can just push to your repo and the changes are automaticly committed to you pull request ;-)

NeverMin:
@monotek 
done. thanks.

NeverMin:
Sorry :-(

2016-10-26 17:57 GMT+08:00 André Bauer <notifications@github.com>:

PR #305, opened 2016-10-26
fixed miss package gawk on Debian

NeverMin:
miss package gawk on Debian #304


NeverMin:
fixed miss package gawk on Debian #305 again.

PR #472, opened 2016-11-29
SLES12, postgresql with sockets, possiblity to use "rake db:drop and rake db:create"

martini:
- Some SLES12 fixes.
- Replaed \t with spaces.
- Postgres: Use socket connections (user changes on pg_hba.conf needed).
- Postgres: Added zammad user as owner to zammad database of be able to use “rake db:drop and rake db:create” (in you want to to a re-setup)
- Moved from config/database.yml.dist ro config/database.yml.pkgr
- Renamed database from zammad_production to zammad (since we only have one zammad database user)

### DIFF
--- a/.pkgr.yml
+++ b/.pkgr.yml
@@ -17,7 +17,7 @@ targets:
       - postgresql|mysql-server|mariadb-server|sqlite
   sles-12:
     dependencies:
-      - nginx|apache2
+      - nginx
       - postgresql-server
 before:
   - uname -a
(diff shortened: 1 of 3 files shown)


martini:
@monotek QA pls  (I have tested it on CentOS7 and SLES12)

monotek:
This stuff should be done by the postinstall refactoring commits...

monotek:
Closing. Imho everything is already done by the postinstall.sh refactoring...

PR #479, opened 2016-11-29
Clean up CTI controller

corny:
I started to clean up code and moved some logic from the SipgateController to models. There is still more work to do.

### DIFF
--- a/app/controllers/integration/sipgate_controller.rb
+++ b/app/controllers/integration/sipgate_controller.rb
@@ -4,14 +4,11 @@ require 'builder'
 
 class Integration::SipgateController < ApplicationController
 
+  before_filter :check_configured
+
   # notify about inbound call / block inbound call
   def in
-    http_log_config facility: 'sipgate.io'
-    return if !configured?
-
     if params['event'] == 'newCall'
-
-      config = Setting.get('sipgate_config')
       config_inbound = config[:inbound] || {}
       block_caller_ids = config_inbound[:block_caller_ids] || []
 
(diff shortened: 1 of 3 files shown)


martini:
Wow! 👍

corny:
By the way: If I add a route and extend the `CtiController` by the the following snippet, I am able to use my [asterisk integration](https://github.com/digineo/zammad-asterisk/tree/ari).
```ruby
  def create
    Cti::Log.process(params)
    head :created
  end
```

martini:
Great work! It seems that you have more experiences with ruby? I'm really interested what you are doing here! :)

corny:
I've been doing Ruby on Rails since 2007. Currently we are using OTRS and Request Tracker in my company. Unfortunately I have a small aversion to perl software from the previous century and that's why I want to give Zammad a chance. Asterisk integration was pretty easy. Next step is to optimize the workflow for incoming calls like:
* Creating tickets by clicking on caller logs.
* Creating users by clicking on caller logs or adding the number to an existing user.

martini:
It there any chance to see the asterisk integration? I'm not familiar with Asterisk so I do not know how to set it up with https://github.com/digineo/zammad-asterisk Maybe a small introduction?

corny:
It's basically what is written in the README. The rest depends on the very individual Asterisk installation. Without a working asterisk server you won't be able to test it. Currently only incoming calls are handled.

If there are other people out there that have Asterisk and Zammad running we can provide help on the integration.

martini:
Any hint for a asterisk mini howto (sip provider for inbound and outbound should be sipgate)?

corny:
Follow the instructions on the [sipgate page](https://teamhelp.sipgate.de/hc/de/articles/203565222-Wie-konfiguriere-ich-Asterisk-f%C3%BCr-sipgate-team-), install Go 1.7, build the program, create the configuration files an run it :+1:

corny:
To use the REST interface of asterisk, you need a really recent version of Asterisk. Ubuntu 16.04 and Debian 7 are shipped with older versions. With older versions you can only use the manager API.
I could rewrite my interface program to support both APIs.

corny:
You need Asterisk 13.6.0 or newer for the REST API.

PR #482, opened 2016-11-30
Fix race condition in tests

corny:
The order was random if objects were created in the same second.

### DIFF
--- a/test/unit/ticket_overview_test.rb
+++ b/test/unit/ticket_overview_test.rb
@@ -614,7 +614,7 @@ class TicketOverviewTest < ActiveSupport::TestCase
     assert(result[2][:tickets].empty?)
 
     overview2.order = {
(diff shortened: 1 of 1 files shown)


corny:
1736ac9a70f614df12cea693a802ef72ed056e08 does not fix it

martini:
IMO mysql is sort incorrectly (IMO mysql has millisecond issue).

monotek:
Switching to postgres for travis tests is not an option?

martini:
IMO not, to test both would IMO better. :)

monotek:
OK, i'll add modifications to your travis-ci-tests branch.

martini:
@corny Thanks for the hint. It's now done with rails test helper.

travel_to Time.zone.now + 1.second # because of mysql millitime issues

PR #485, opened 2016-11-30
Small refactoring of logical expressions

corny:


app/models/user.rb
```diff
@@ -154,18 +154,7 @@ def longname
...
-        next if !role_name.include?(role.name)
-      elsif role.name != role_name
-        next
-      end
-      result = true
-      break
-    }
-    result
+    roles.where(name: role_name).any?
   end
```
corny: I'm afraid there are other constructs like this hidden in the code ...

thorsteneckel:
Thanks - great!

PR #486, opened 2016-11-30
Move calendar mapping to config directory

corny:


monotek:
Whats the reason to move this to a file?
Imho it's easier for the user if he is able to do most of the configuration through the frontend.

corny:
Because it's not a part of the (business) logic and blows up the model. Files of several hundred lines of code should be also splitted in mid-term. [ActiveSupport::Concern](http://api.rubyonrails.org/classes/ActiveSupport/Concern.html) is the way to go.

corny:
This commit does not change logic. It just removes the data from the `model` directory.

martini:
@monotek you still can use the ui for edit/changing -> It's just the ical feed list for the different countries.

PR #490, opened 2016-11-30
Use deliver_method :test for tests

corny:
This avoids flooding the screen with warnings on systems without MTA:

Please install an MTA on this system if you want to use sendmail!
Please install an MTA on this system if you want to use sendmail!
Please install an MTA on this system if you want to use sendmail!


thorsteneckel:
Hi @corny - thanks! I merged it with slight changes in https://github.com/zammad/zammad/commit/2ed9ed5897f5ee559f02797b0a64bb1cac85cf30

corny:
Thanks for the change.

PR #495, opened 2016-12-02
Refactor Cti::CallerId

corny:
- execute rebuild in a single transaction
- reduce queries
- DRY up code
- satisfy RuboCop

### DIFF
--- a/app/models/cti/caller_id.rb
+++ b/app/models/cti/caller_id.rb
@@ -16,22 +16,17 @@ module Cti
 =end
 
     def self.maybe_add(data)
-      records = Cti::CallerId.where(
+      record = find_or_initialize_by(
         caller_id: data[:caller_id],
-        level: data[:level],
-        object: data[:object],
-        o_id: data[:o_id],
-        user_id: data[:user_id],
+        level:     data[:level],
+        object:    data[:object],
+        o_id:      data[:o_id],
+        user_id:   data[:user_id],
       )
-      return if records[0]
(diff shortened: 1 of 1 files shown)


corny:
Tests failed: `Mysql2::Error: Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)`
`rspec spec/models/cti/caller_id_spec.rb` works for me.

monotek:
Seems to be a travis Bug:

https://github.com/travis-ci/travis-ci/issues/6961

corny:
Ok, please test it locally.

app/models/cti/caller_id.rb
```diff
@@ -23,10 +23,9 @@ def self.maybe_add(data)
         o_id:      data[:o_id],
         user_id:   data[:user_id],
       )
-      if record.new_record?
-        record.comment = data[:comment]
-        record.save!
-      end
+      return if !record.new_record?
+      record.comment = data[:comment]
+      record.save!
```
corny: I prefer the previous version that always returns the record. If you want to keep you version better write `return record record if record.persisted?` or `return record unless record.new_record?`

thorsteneckel: Thanks - done

corny:
Looks fine for me. But I still prefer `unless` instead of `if !` ;-)

corny:
Merge? :)

PR #497, opened 2016-12-02
Fix and refactor Cti::CallerId.parse_text

corny:
- It did non recognize mobile numbers like '0170 12345678'.
- Support for numbers without country/area prefixes dropped.
- `get_comment_preferences` has to normalize the number too.


app/models/cti/caller_id.rb
```diff
@@ -2,6 +2,8 @@ module Cti
   class CallerId < ApplicationModel
     self.table_name = 'cti_caller_ids'
 
+    DefaultCountryId = '49'
```
thorsteneckel: I think we should define this via the settings / depending on the location of of zammad instance / admin like the translations. What do you think?

app/models/cti/caller_id.rb
```diff
@@ -197,40 +199,21 @@ def self.config
...
-        number.gsub!(/(\s|\-|\(|\))/, '')
-        if !Phony.plausible?(number)
-          if number =~ /^0/
-            number.gsub!(/^0/, default_country_id)
-          else
-            number = "#{default_country_id}#{number}"
-          end
-          next if !Phony.plausible?(number)
+      # see specs for example
+      text.scan(/([\d|\s|\-|\(|\)]{6,26})/).map do |match|
```
thorsteneckel: I think the yielded block should be moved into a separate function since here are (at least) two functionalities in one method:

1. looping over a text extracting numbers
2. working up numbers to be in a specific format

You can see this even clearer in the tests where those two get mixed together.

Would be great if you could split this up. This would make the tests and API more clearer and splits the responsibilities as they should be. In the long term both functions could be used independently for various use cases.

spec/models/cti/caller_id_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+RSpec.describe User do
+
+  describe 'parse_text' do
+    # can be anything
+    it { expect(Cti::CallerId.parse_text('5754321')).to eq ['5754321'] }
+    it { expect(Cti::CallerId.parse_text('622 32281')).to eq ['62232281'] }
+
+    it { expect(Cti::CallerId.parse_text("some text\ntest 123")).to eq [] }
+    it { expect(Cti::CallerId.parse_text('0049 1234 123456789')).to eq ['491234123456789'] }
```
thorsteneckel: number cleaning starts here (see comment in the parse_text method)

spec/models/cti/caller_id_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    it { expect(Cti::CallerId.parse_text('092213212')).to eq ['4992213212'] }
+    it { expect(Cti::CallerId.parse_text('(09)1234321')).to eq ['4991234321'] }
+    it { expect(Cti::CallerId.parse_text('+49 30 53 00 00 000')).to eq ['4930530000000'] }
+    it { expect(Cti::CallerId.parse_text('+49 160 0000000')).to eq ['491600000000'] }
+    it { expect(Cti::CallerId.parse_text('+49 (0) 30 60 00 00 00-0')).to eq ['4930600000000'] }
+    it { expect(Cti::CallerId.parse_text('0043 (0) 30 60 00 00 00-0')).to eq ['4330600000000'] }
+    it { expect(Cti::CallerId.parse_text('0043 30 60 00 00 00-0')).to eq ['4330600000000'] }
+    it { expect(Cti::CallerId.parse_text('1-888-407-4747')).to eq ['18884074747'] }
+    it { expect(Cti::CallerId.parse_text('Lorem ipsum dolor sit amet, consectetuer +49 (0) 30 60 00 00 00-0 adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. Donec pede justo, fringilla vel.')).to eq ['4930600000000'] }
```
thorsteneckel: text parsing starts here (see comment in the parse_text method)

spec/models/cti/caller_id_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.describe User do
```
corny: This is an copy and paste error.

app/models/cti/caller_id.rb
corny: :+1: but adding features should be another pull request. This PR is only a fix.

app/models/cti/caller_id.rb
corny: Good idead. I will split this up.

spec/models/cti/caller_id_spec.rb
corny: I removed Phony because the only method used was `Phony.plausible?` and I could satisfy all test cases without Phony. `Phony.split` could be useful to display the number in a readable format.

app/models/cti/caller_id.rb
thorsteneckel: Agree

corny:
Done, please squash on merge.

corny:
`test/integration/sipgate_controller_test.rb` runs without failures when I run it isolated.

thorsteneckel:
Squash merged with https://github.com/zammad/zammad/commit/f39f66fe96ae0d93f8008f3c77893a908599b4d5

PR #517, opened 2016-12-07
improved call log workflow

TeaMoe:
We wanted to be able to create a Ticket from an "inidentified" call log entry. 
To do this i linked unknown phone numbers in caller log and bound click event to open new user modal.
When user is saved, the user is taken to to the new ticket page with the new customer pre filled.

### DIFF
--- a/app/assets/javascripts/app/controllers/cti.coffee
+++ b/app/assets/javascripts/app/controllers/cti.coffee
@@ -1,7 +1,7 @@
 class App.CTI extends App.Controller
   events:
     'click .js-check': 'done',
-    'click [data-type=new]': 'new_user'
+    'click .js-new-user': 'new_user'
 
   constructor: ->
     super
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/app/controllers/cti.coffee
```diff
@@ -1,6 +1,7 @@
 class App.CTI extends App.Controller
   events:
-    'click .js-check': 'done'
+    'click .js-check': 'done',
+    'click [data-type=new]': 'new_user'
```
thorsteneckel: Please use 'js-' prefixed classes (like in the line above) to identify elements that take action(s) in JS instead of data attributes.

TeaMoe:
done

thorsteneckel:
Thanks! We will review this and come back to you. Might be necessary to add some tests for this.

martini:
@TeaMoe Nice work too! ❤️

PR #518, opened 2016-12-08
Adding translations to source (update via contrib/translation-download.sh)

monotek:
packager.io is constantly fails building new packages because of the translation download, so i put the translations to the source.

There is a ne file contrib/translation-download.sh which downloads translations using a temporary sqlite db. This script would need to be run before a Zammad release to update all translations.

P.S: No need to merge. I'll put it in git.znuny.com if everybody is ok with it.


martini:
If I got it correctly, you want to put the translation download files (.yml) into the git repo?

So you do not want to call contrib/localisation-download.sh via packager.io, only for the packages?

monotek:
correct. imho this process needs to long in packager.io. 
imho thats the reason the builds are failing.
therefore it would be better to have this already in the sources.

thorsteneckel:
I don't think that the Zammad source repo is the right place for the translations. This will lead to a lot of side efffects from pull requests to manual syncs etc... What concrete leads to the aborted processes? Can we maybe increase a timeout somewhere?

monotek:
packager.io is kind of a blackbox for us.
Not exactly sure whats the problem.
The database dies with:

```
WARNING:  terminating connection because of crash of another server process
       DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
       HINT:  In a moment you should be able to reconnect to the database and repeat your command.
       rake aborted!
```

i'll try to switch to sqlite instead of postgresql. Maybe it helps.

If not, what to do? Separate (sub)repo for translations?

martini:
> If not, what to do? Separate (sub)repo for translations?

I'll try 10 lines ruby today.

monotek:
Switch to sqlite does not work. Postgresql is used anyway...

monotek:
Any news on this topic?

Imho it would be better to be able to fetch this stuff via wget instead needing a running rails environment & database...

martini:
@monotek here you go https://github.com/zammad/zammad/commit/f9c104093033b08a530d9ced36f70cdd2febb3ee it a ruby only (without rails) implementation.

monotek:
Great! Now we have really nice & fast GitHub to packager.io to dockerhub buildchain :-)

martini:
Package build also looks to be stable again!  🤓

monotek:
On small issue. New package size is more than 40MB larger :-/

Before around 65MB:
https://packager.io/gh/zammad/zammad/build_runs/1428

Now around 107MB:
https://packager.io/gh/zammad/zammad/build_runs/1439

Maybe there could be done some cleanup in the after script?
I guess the translations are not 40 MB big?

martini:
du -sh config/translations/
3.6M	config/translations/

No, only 3,6 MB. 

If I compare
https://packager.io/gh/zammad/zammad/build_runs/1326 -> (91.65 MiB)

and

https://packager.io/gh/zammad/zammad/build_runs/1439 -> (107.49 MiB)

seems to be fine for me.

If I install the cents 7 package I see the following:

[root@localhost ~]# du -sh /opt/zammad/
667M	/opt/zammad/
[root@localhost ~]# du -sh /opt/zammad/tmp/cache/assets/sprockets/v3.0/
423M	/opt/zammad/tmp/cache/assets/sprockets/v3.0/
[root@localhost ~]# du -sh /opt/zammad/vendor/
200M	/opt/zammad/vendor/
[root@localhost ~]# 

Main content is in:
 -> sprockets -> Precompiled assets -> is ok - precompile assets will build faster next time btw. contrib/packager.io/postinstall.sh is faster
 -> vendor -> no chance to make it smaller

-=> For me it seems that packager io has changed something in the build process.

Win: My cents 7 installation was up and running within 2 min. because of already Precompiled assets
 -=> So for me the big package looks better as the small one.

monotek:
Not sure if its true that packager.io changed something. 
The bigger filesize showed up first time when the switch to the new download method was made.
I guess because you removed "rm -rf tmp/cache*" from the after script and thats a difference of 1,2GB uncompressed.

Has this dir something to do with the language files?
If not, is it realy needed?

martini:
> Has this dir something to do with the language files?

No.

> If not, is it realy needed?

It speed up the package installation from about 7 min to 2 min :-o

martini:
> The bigger filesize showed up first time when the switch to the new download method was made.

No it was "sometimes"  🤔 also before:
https://packager.io/gh/zammad/zammad/build_runs/1326 -> (91.65 MiB)

monotek:
Yes. I just compared to the version before.

The question to me is: Do we need tmp/cache in the package or not?

martini:
> The question to me is: Do we need tmp/cache in the package or not?

Yea. It speed up the package installation from about 7 min to 2 min :-o

config/locales.yml
```diff
@@ -0,0 +1,145 @@
...
+  active: true
+- locale: vi
+  alias: ''
+  name: Vietnam (ViɆt Nam)
+  active: true
+- locale: zh-cn
+  alias: ''
+  name: Chinese (Sim.) (简体中文)
+  active: true
+- locale: zj-tw
```
NeverMin: Hi @martini , How to fix it, maybe it is `locale: zh-tw` ?

PR #595, opened 2017-01-06
Chinese for the Chat widget

NeverMin:
Added Chinese translation for the Chat widget.


martini:
@NeverMin JFI, there was also a bug in chat.coffee to use xx-xx locales. It's no fixed. Thanks for the translation!

Issue #395, opened 2016-11-11
Auto-assign ticket owner on first reply


jacotec:
I'd like the option that the first agent, who sends an answer to the customer on a new ticket, gets auto-assigned to the ticket in case there is no assignment yet.

This would prevent to forget to manually select the assignment in the menu bar in addition to the answer.

If an assignment is already done it should not be changed by further responses of other agents.

thorsteneckel:
Hi @jacotec - Thanks for your proposal. Do I get you right:

If opening a ticket in the detail view that is new/has yet no owner the owner selection should be automatically set to yourself / the current agent. This will automatically set the owner for responses and saves some clicks. Additionally it will keep the possibility to any other agent or reset to no agent - if you / the current agent wants to inform all agents of the group about his reply.

monotek:
Imho autoasign only should take place after sending an answer to the customer and not when agent reads the ticket.

jacotec:
Hi @thorsteneckel ,

nearly ;-)

The auto assignment of the owner should not be done when opening the ticket (otherwise you can't just open to read without being assigned!)

The auto assignment would be useful in this case:

- The ticket has no owner yet
- The agent **replies to the customer** (just an internal note will not assign it!)
- The agent sends out his reply to the customer by the "update" button. (That's the point where the assignment triggers. If the agent has assigned someone else during that action, the selected assignment would be used)

This would prevent from answered tickets being unassigned to an owner.

jacotec:
@monotek Exactly! You've been 33 seconds faster ;-)

thorsteneckel:
Yes, thats what I meant. Just preselected with the change indication next to the field. Thanks!

jacotec:
For now I'm doing it with a Macro (Close and send). This assigns the ticket to the agent and closes it. But the auto assignment as stated above would be generally useful as any answer from the customer will directed back to the agent who wrote the reply by default. Otherwise if the agent does not assign itself it'll run into the group globally.

olivergrahl:
Hmm, so I tried to work around this with a trigger, but I probably don't understand how triggers work, as it has no effect at all?

![grafik](https://cloud.githubusercontent.com/assets/4621219/20622264/4ce2f88a-b302-11e6-9957-286cdc11cdf4.png)

Taps7734:
GMTA... I tried the same trigger (I think.  I tried it in English, but I believe we used the same options) and it also did nothing.

Any help on this?

[...]

PR #601, opened 2017-01-10
Fix an error condition in Trigger

olivergrahl:
Checking for `@item[:changes][attribute]` before making sure `@item[:changes]` is valid 
may result in an error (NoMethodError: undefined method `[]' for nil:NilClass).


olivergrahl:
Okay, looks like this makes some test fail. Basically, I was trying to solve issue #395, but it looks like it's more complicated than that. Any ideas?

PR #624, opened 2017-01-15
Update the call for participation in JS console

dereckson:
Previously, if opened in developer mode, an hello banner
advertised job opportunities at Zammad.

As this isn't a priority anymore, switch to Zammad Commnity
participate invitation instead to serve a 404 error page.


martini:
@dereckson Thanks!

PR #630, opened 2017-01-16
add sync setting 'track_retweets' to twitter channel

### LINKED ISSUE #627: Option to not create ticket on retweet
It would be nice to turn off ticket creation on retweets. Quotes are fine in most cases but Zammad currently also creates a ticket for every single retweet.

schurig:
Hopefully resolves #627

### DIFF
--- a/app/assets/javascripts/app/controllers/_channel/twitter.coffee
+++ b/app/assets/javascripts/app/controllers/_channel/twitter.coffee
@@ -269,9 +269,9 @@ class AccountEdit extends App.ControllerModal
       else
         search.push params.search
     if params.track_retweets
-      params.track_retweets = 1
+      params.track_retweets = true
     else
-      params.track_retweets = 0
+      params.track_retweets = false
     params.search = search
     @channel.options.sync = params
     @ajax(
--- a/app/assets/javascripts/app/controllers/_channel/twitter.coffee
+++ b/app/assets/javascripts/app/controllers/_channel/twitter.coffee
@@ -268,6 +268,10 @@ class AccountEdit extends App.ControllerModal
           position += 1
       else
         search.push params.search
+    if params.track_retweets
+      params.track_retweets = 1
+    else
+      params.track_retweets = 0
     params.search = search
     @channel.options.sync = params
     @ajax(
--- a/app/assets/javascripts/app/views/twitter/account_edit.jst.eco
+++ b/app/assets/javascripts/app/views/twitter/account_edit.jst.eco
@@ -22,7 +22,7 @@
           <%- @Icon('plus-small') %>
     </tfoot>
   </table>
-  
+
   <h3><%- @T('Mentions Group') %></h3>
   <p class="description"><%- @T('Choose which group %s will get added to.', 'mentions') %></p>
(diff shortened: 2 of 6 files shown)


thorsteneckel:
Nice 🚀 I'll review this tomorrow and let you know.

app/assets/javascripts/app/controllers/_channel/twitter.coffee
```diff
@@ -268,6 +268,10 @@ class AccountEdit extends App.ControllerModal
           position += 1
       else
         search.push params.search
+    if params.track_retweets
+      params.track_retweets = 1
```
thorsteneckel: Shouldn't we use true and false for these boolean typed variables?

app/assets/javascripts/app/views/twitter/account_edit.jst.eco
```diff
@@ -31,4 +31,8 @@
   <p class="description"><%- @T('Choose which group %s will get added to.', 'direct messages') %></p>
   <div class="js-directMessagesGroup"></div>
 
+  <h3>Retweets</h3>
```
thorsteneckel: I think this headline should be translatable, too.

app/assets/javascripts/app/views/twitter/account_edit.jst.eco
```diff
@@ -31,4 +31,8 @@
...
+  <p class="description"><%- @T('Choose if %s should also be converted to tickets.', 'retweets') %></p>
```
thorsteneckel: Shouldn't be 'retweets' in the translation string to it could get translated in e.g. mandarin?

app/assets/javascripts/app/views/twitter/list.jst.eco
```diff
@@ -60,6 +60,13 @@
         <% end %>
       </div>
     </div>
+    <div class="action-flow action-flow--row">
+      <div class="action-block">
+        <h3><%- @T('Retweets') %></h3>
+        <% if channel.options.sync.track_retweets: %><%- @T('%s are converted to Tickets', 'retweets') %>.<% end %>
```
thorsteneckel: If we use boolean variables we should use an if / else block here.

app/assets/javascripts/app/controllers/_channel/twitter.coffee
schurig: I was about to do that but putting booleans there makes Zammad save them as string in the database. Can you tell me if they get converted back to booleans when I access the channel in the backend?

app/assets/javascripts/app/views/twitter/account_edit.jst.eco
schurig: @thorsteneckel what is a good alternative in your opinion?

schurig:
@thorsteneckel pushed some changes. Can you help me with [this](https://github.com/zammad/zammad/pull/630#discussion_r96628852) please?

thorsteneckel:
Hi @schurig - had to make some changes and merged it via squash so your credits got lost 😞 Nevertheless: Big thanks for your support and effort!

PR #681, opened 2017-01-27
bugfix: Keep modal open on call log event

corny:
Otherwise the modal closes on hangup.
I hope it does not have any unwanted side effects.

### DIFF
--- a/app/assets/javascripts/app/controllers/cti.coffee
+++ b/app/assets/javascripts/app/controllers/cti.coffee
@@ -181,7 +181,7 @@ class App.CTI extends App.Controller
       genericObject: 'User'
       item:
         phone: phone
-      container: @el.closest('.content')
+      container: @el.closest('#app')
       callback: @ticketNew
     )
 


martini:
Thanks @corny @TeaMoe for this PR. 

To change the DOM position of the modal would have side effects (e. g. switching to other tab will not be possible because the cti new customer modal is blocking the whole app).

-=> To get this issue fixed we split the cti content (cti page + caller log in there) into two elements. We only rerender the caller log (not the while cti page). This change will keep the local modal dialog for the cti page. 🤓

Thanks a lot for the contribution (also if it's not 1:1 taken over).

Here the commit as reference: https://github.com/zammad/zammad/commit/4d7f3c1e6a1ee783bb99c619543fd34fb7062900

 -Martin

corny:
That's a nice solution. Thanks a lot for your work!

martini:
Thanks for reporting and providing solutions!

PR #752, opened 2017-02-09
Czech email template localisation

RAD-X:


martini:
@RAD-X Thanks for great work! It would be also really cool if you could translate the missing string in Zammad too. 

Just do the following:
1) Login with an admin account
2) Select Czech in Profile -> Language
3) Go to Admin -> Translations 

Translate strings and at the end, submit your changes (on the blue button at top right corner). Then it's available for everybody. ❤️

RAD-X:
Of course I did so. :-) For now about 80% translated. :sunglasses:

martini:
You are my man! 🍾

Issue #698, opened 2017-01-31
Some text hard to read because of low contrast


johan12345:
### Infos:

* Used Zammad version: 1.2
* Operating system: Windows 10
* Browser + version: Chrome 55

### Actual behavior:

* On one of my computers, I have a cheap, somewhat old LCD screen whose contrast ratio is not that great. When I look at a ticket overview in Zammad on this screen, the columns that don't have a link (-> use gray text instead of blue) are quite hard to read because of the low contrast ratio between the text and the background
* The colors used seem to be #f6f7f8 (background) and #aabdc5 (text). Their contrast ratio is lower than the W3C recommendation according to [this tool](http://leaverou.github.io/contrast-ratio/#%23aabdc5-on-%23f6f7f8), so this is probably not only a problem for people with crappy screens, but also an accessibility issue.
* The same color combination is also used in other places, e.g. for the timestamps below each entry in a ticket, but not in detailed search results (there, the text is #585856 instead, which is perfectly readable).

martini:
Is this issue still open? Or already done with PR #783  ?

johan12345:
@martini Well, there are some more places where #aabdc5 and other rather light colors are used as the text color that I haven't changed in PR #783, e.g.
- Ticket number in the upper left corner
- Number of tickets in an overview
- Details of the stats on the front page
- ...

nsrosenqvist:
The sidebar is really hard to read as well in 2.2.0 when an item isn't selected. I have to adjust my position in order to look at the screen at a good angle just to be able to see what it says

MrGeneration:
I believe that this directly covered by this issue:
https://github.com/zammad/zammad/issues/3067

As it's covering more than your issue (at least this is what I believe :x) I will close this issue as a duplicate.

Thank you very much for your time!

PR #783, opened 2017-02-18
Some adjustments with regard to #698

johan12345:
This replaces some very light text colors with the default #585856 color, which passes the [W3C recommendation](http://leaverou.github.io/contrast-ratio/) for text readability.

This is not a complete fix of the problem, there are still places where the text is too light, but the two places where I changed it are the ones I mentioned in #698 (ticket overview table, ticket entry timestamps).

### DIFF
--- a/app/assets/stylesheets/zammad.scss
+++ b/app/assets/stylesheets/zammad.scss
@@ -121,7 +121,7 @@ a.create {
 }
 
 small {
-  color: #c6c6c5;
+  color: hsl(198,4%,56%);
   font-size: 12px;
 }
 
@@ -889,7 +889,7 @@ table {
 }
 
 .table--light {
-  color: hsl(198,19%,72%);
+  color: hsl(198,4%,56%);
 }
 
 .table-fluid {


thorsteneckel:
Thanks @johan12345 ! @mrflix can you take a look at it? 🤓

mrflix:
Hey Johan! Thanks for participating!
I find the switch to the default text color too harsh. In both cases the text is grayed out on purpose because the information it holds is secondary. I suggest something in the middle:
ticket
overview
The gif is not perfectly representable of the change. Click on it to see it in bigger size.
The new color is `hsl(198,4%,56%)` in both cases.
Can you test this on the cheap LCD and report back?

johan12345:
Hi,

yes, I understand that the text is grayed out on purpose, even though then I don't know why the default text color in the search results tables.

The compromise you suggested will probably be enough to make it more readable on the cheap LCD (though I can't test that at the moment - I'll be back at that computer on thursday).

But if you want to follow the W3C [accessibility guidelines](https://www.w3.org/TR/WCAG/#visual-audio-contrast), this only leads to a [contrast ratio of 3](http://leaverou.github.io/contrast-ratio/#hsl%28198%2C4%25%2C56%25%29-on-%23f6f7f8), which is not enough for small text (< 18 pt). For passing the AA accessibility level, you'd need at least [`hsl(198,4%,44.5%)`](http://leaverou.github.io/contrast-ratio/#hsl%28198%2C4%25%2C44.5%25%29-on-%23f6f7f8) and [`hsl(198,4%,33%)`](http://leaverou.github.io/contrast-ratio/#hsl%28198%2C4%25%2C33%25%29-on-%23f6f7f8) (which is even darker than the default color) for the AAA level.

mrflix:
> even though then I don't know why the default text color (is used) in the search results tables.

Good point. I'll check that out. Most probably because thats a part that wasn't designed as a mockup so it's not that pixel perfect.

> But if you want to follow the W3C accessibility guidelines, [...]

I do know that it is only a ratio of 3.4 and I think its okay. 44.5% and 33% brightness is just not looking good. At least not on a Mac. The gamma curve on Windows is different but I hope we can find a compromise for both.

If its okay on the cheap LCD and on Windows in general (@martini please verify the contrast as well)  I'd suggest that you change the colors in the commit to the color I suggested because in the end it was your idea and then I'll merge the pull request.

johan12345:
Okay, I'll test it on the cheap LCD on Thursday and then update the commit accordingly. On the other Windows PC I am currently at (with newer and more expensive Dell screens) it looks good with the color you suggested.

johan12345:
I have tested it on the cheap LCD: The `hsl(198,4%,56%)` is already much better.
I have now updated the PR accordingly.

mrflix:
Great, thanks once again!

jancborchardt:
@mrflix awesome to see you’re working on this! :) We use Zammad at Nextcloud and I was just impressed by the nice interface. :)

PR #784, opened 2017-02-18
Allow to send email notifications from scheduled jobs

nidico:
Surprisingly, this just worked. Is there any reason, email notifactions aren't enabled for scheduled jobs (and macros, too)?


martini:
Hi @nidico 

> Surprisingly, this just worked. Is there any reason, email notifactions aren't enabled for scheduled jobs 

No specific reason. The use case was just not in mind.

 -Martin

nidico:
Cool!

Issue #203, opened 2016-09-19
Mailprocessing: Reply-To header are ignored


hanneshal:
If I send a mail via a generic form mailer like this:

```
X-PHP-Originating-Script: 33:form-handler.php
From: foobar@bar.com
MIME-Version: 1.0
Content-type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable
Reply-To: hannes@myspecialdomain.com
Message-Id: <20160919130314.34EEB1DA1B2@bar.com>
Date: Mon, 19 Sep 2016 15:03:14 +0200 (CEST)
X-znuny-MailScanner-Information: Please contact the ISP for more information
X-znuny-MailScanner-ID: 5E64620C3DE3.A1C7C
X-znuny-MailScanner: Found to be clean
X-znuny-MailScanner-From: www-data@bar.com
X-Spam-Status: No
```

The Reply-To header, which is this customer address, is ignored.

olivergrahl:
Any objections against [this change](https://github.com/olivergrahl/zammad/commit/d7b35a9367379596be7388fc9eb0f949add044c1) to ignore the From header in favour of the Reply-To header? Would be happy to send a PR.

thorsteneckel:
Hi @olivergrahl - I hadn't time yet to take a deeper look but looks promising. Just one question so far: Could you please use a more speaking variable name instead of `from_from`? Maybe something like `reply_to`?

olivergrahl:
Okay, `from_from` was probably neither creative nor very self explanatory. :-)  I renamed it to `from_full` which is possibly not very self explanatory either. `from_full` holds the complete address with name and email, while `from_email`, `from_domain`, `from_display_name` hold its parsed sub elements. All of them are based on either `reply-to`, `from` or `return-path`. Let me know if you come up with a better name.

olivergrahl:
Let me add that this code works well here, but the GUI displays the "Reply All" button, which would reply to the `reply-to` (VON below) and the `to` address (AN below, the email channel inbound address), and I believe that doesn't make too much sense. I'm not sure how this could be prevented, though. 
![grafik](https://cloud.githubusercontent.com/assets/4621219/21815285/57606fda-d75c-11e6-8e6d-4beb21e3a8de.png)

rolfschmidt:
Hi all,
we decided to create an own column to save the reply to information. It will be displayed now in the detail information of the article:

[...]

PR #943, opened 2017-04-11
Feat use reply to header

chschenk:
#203


martini:
Hi @olivergrahl 

question, the goal is to to replace from by reply-to (if given)? Or show from and for auto responders + email reply use reply-to?

 -Martin

olivergrahl:
Hi Martin,

I think when there's a "reply-to" address, any replies (auto-responders and email) should by default go to that address, ignoring the "from" address. I guess Zammad should behave the same way most email client do.

My implementation above is more of a quick & dirty fix, as it completely replaces the "from" with the "reply-to". It works perfectly here (we couldn't even work without it), but it kind of omits the original "from", so that it cannot be seen anywhere anymore. A better solution would probably be to hold the "reply-to" in a separate db field, display both in the ticket details (between "from" and "to", after clicking a message), and defaulting to using "reply-to" instead of "from" when it exists. 

This would probably raise more questions, like which address to use for the user account, etc. 
So, not sure if this extra work is really required.

Ah, and here's a use case: Our software has the ability to send bug reports from a crash screen. User can optionally enter their email address. The software sends those emails through our own SMTP with a "from" like "noreply@ourdomain.com", and the user's address in the "reply-to". We could not set the user's address as the "from", because then it would look like we're spoofing their mail, and the messages would not make it through any filters. I guess this is a typical use on 3rd party contact forms (that send out email) as well. 

HTH,
-Oliver

rolfschmidt:
Fixed. Please take a look at #203 for more information.

PR #977, opened 2017-04-23
hidden form on small screen/smartphone

chrisdecker1201:
I noticed that on small screen/smartphone the modal of the contact form does not show up. So I added max-width to the css class.


martini:
Thanks a lot!

PR #993, opened 2017-04-25
Localization on templates

gilsonmontanhajr:
New file templates for Brazilian Portuguese.


martini:
Thanks a lot @gilsonmontanhajr !

Issue #806, opened 2017-02-27
Implement inbound SMTP listener for E-Mail


siepkes:
I was thinking that it would be really nice if Zammad could behave as an SMTP server (MTA) with the help of for example the [Ruby mini smtp server](https://github.com/aarongough/mini-smtp-server). The benfit would be that one can configure his main MTA (postfix, sendmail, etc.) to just relay mail to Zammad. Zammad could then function as the endpoint for the email.

This has advantages over using an IMAP server in between:
*  Less moving parts, simple delivery directly to Zammad. No need to maintain IMAP boxes which are only used as an intermediary.
* MTA can better respond to Zammad downtime. For example bounce message after 2 days with an undelivered message if it can't be relayed to Zammad when Zammad is down. If the mail is sitting in an IMAP spool for a couple of days the MTA will have already communicated to the sending MTA that delivery was successful while in fact it was not.

I was thinking that in the "Email Inbound" selection box the option SMTP could be selected and a port could be provided on which Zammad would then start listening. 

What do you guys think of such an idea? (I might be interested in implementing this feature so I would like some feedback).

martini:
Hi @siepkes 

from the technical point of view it's really nice. I really like it. On the other hand, I see some missing features in Ruby mini smtp server like ssl or spam protections and many other config options which postfix, sendmail, .. has.

So another idea comes into my mind. How about an public service to create customisable Zammad appliances.

Let's say:
o Webserver
 o Apache|Nginx
 o ssl

o Database
 o PostgreSQL|MySQL

o MTA
 o Sendmail|Postfix|Exim
 o accept direct SMTP connections
  o certain spam options

o Network
 o IP
 o Firewall Rules
 o ssh keys

o Elasticsearch

And so on, after 10 min. you will get an download link with your Zammad appliance.

Feedback is welcome!

 -Martin

siepkes:
Hi @martini ! Thanks for you quick reply!

> like ssl or spam protections and many other config options which postfix, sendmail

[...]

PR #999, opened 2017-04-26
Added possibility of creating an inbound smtp channel, which makes Za…

MaartenR:
As mentioned in #806 we (@ServicePlanet) like to see a feature which adds the possibility of creating inbound smtp email channels. Zammad would then listen on a port for smtp connections and incoming emails would be processed through this channel. As mentioned in the issue the idea is that this feature allows a frontend MTA to relay mails directly to Zammad. 

This pull request contains the current implementation of this feature and i was wondering whether you guys would consider merging this, and if not, if there are any changes that can be made to make you reconsider.

Currently it works by creating an email account and selecting SMTP as the type. After the account is created Zammad starts an smtp server that listens for incoming connections on the configured port. The host and user are still configured, because they seem to function like channel identifiers. The password has become optional because it is not used here. I feel that ideally fields would be optional or mandatory depending on the adapter type that was selected, but currently that is not the case. What do you think?

Any kind of feedback is certainly appreciated. Information about the problem this is supposed to solve can be found in issue #806.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 
 # Ignore Rubymine config
 /.idea
+*.iml
 
 # Ignore .project files
 /.project
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,8 @@ gem 'em-websocket'
 
 gem 'diffy'
 
+gem 'midi-smtp-server'
+
 # Gems used only for develop/test and not required
 # in production environments by default.
 group :development, :test do
--- a/app/assets/javascripts/app/controllers/_channel/email.coffee
+++ b/app/assets/javascripts/app/controllers/_channel/email.coffee
@@ -565,31 +565,47 @@ class App.ChannelEmailAccountWizard extends App.WizardModal
       { name: 'adapter',            display: 'Type',     tag: 'select', multiple: false, null: false, options: @channelDriver.email.inbound },
       { name: 'options::host',      display: 'Host',     tag: 'input',  type: 'text', limit: 120, null: false, autocapitalize: false },
       { name: 'options::user',      display: 'User',     tag: 'input',  type: 'text', limit: 120, null: false, autocapitalize: false, autocomplete: 'new-password', },
-      { name: 'options::password',  display: 'Password', tag: 'input',  type: 'password', limit: 120, null: false, autocapitalize: false, autocomplete: 'new-password', single: true },
+      { name: 'options::password',  display: 'Password', tag: 'input',  type: 'password', limit: 120, null: true, autocapitalize: false, autocomplete: 'new-password', single: true },
       { name: 'options::ssl',       display: 'SSL',      tag: 'boolean', null: true, options: { true: 'yes', false: 'no'  }, default: true, translate: true, item_class: 'formGroup--halfSize' },
       { name: 'options::port',      display: 'Port',     tag: 'input',  type: 'text', limit: 6,   null: true, autocapitalize: false,  default: '993', item_class: 'formGroup--halfSize' },
       { name: 'options::folder',    display: 'Folder',   tag: 'input',  type: 'text', limit: 120, null: true, autocapitalize: false },
     ]
 
-    showHideFolder = (params, attribute, attributes, classname, form, ui) ->
+    showHideOptions = (params, attribute, attributes, classname, form, ui) ->
       return if !params
-      if params.adapter is 'imap'
-        ui.show('options::folder')
-        return
-      ui.hide('options::folder')
+      if params.adapter is 'smtp'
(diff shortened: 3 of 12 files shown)


thorsteneckel:
Hi @MaartenR - Thanks for this really interesting solution and sorry for the late response. We took our time to discuss this and decided against merging it into the Zammad core. The changes will increase the maintenance costs and attack vector. However we would love to see this as an Zammad extension 😍 for those who can profit from it. Zammad already has a build in extension system. We are happy and ready to support you building the first public Zammad extension 🚀 Just contact us at support@zammad.com and refer to me in case you are interested. Closing now.

siepkes:
@thorsteneckel We discussed this internally and decided that the extension infrastructure Zammad currently offers is too immature for us to maintain this. The extension infrastructure doesn't offer a stable extension API. Only the internal (private) API of Zammad is available so our extension could (and probably will) break with each new Zammad release. 

Maintaining a fork and keeping that in sync with Zammad is actually easier for us then maintaining an external (though partly internal) extension.

We are however still interested in providing a method in Zammad which allows for real time bulk insertion with low over head. Would you guys be open to a contribution adding a  REST API `/api/v1/bulk_insert` endpoint? Which would allow a raw mail to be inserted in to Zammad which Zammad would then process just like mail scrapped from IMAP? This endpoint would obviously require the user to have a special "bulk_insert" permission.

thorsteneckel:
Hi @siepkes - We totally understand your concerns.

Does it help you to process mails [via STDIN](https://github.com/zammad/zammad/blob/develop/app/models/channel/driver/mail_stdin.rb#L9)?

siepkes:
@thorsteneckel Our main problem with the stdin method is (and correct me if I'm wrong); Every time the MTA (postfix in our case) delivers a single mail via stdin the whole Ruby interpreter is started, which starts rails, creates DB connections, processes a single mail and then closes the DB connections and stops the interpreter. And this cycle happens for every single mail we deliver in to Zammad. IMHO the overhead of delivering a single mail that way is just too big.

thorsteneckel:
Hi @siepkes - thanks for clearing things up. One possible solution would be to introduce a new Channel backend (just like `Channel::Driver::MailStdin`) which instead of reading STDIN reads out a defined directory where you put your mails in. There would be an additional Scheduler job which is performed on a regular base which starts the Channel backend. In that way you won't have to start the whole application because the Scheduler is always running. Only downside would be that mails will be processed with the offset till the next scheduler job execution. Do you think this is a possible solution?

siepkes:
@thorsteneckel **Short story:** After some internal discussion we agreed the embedded SMTP server might be a bit too complex and understand why you are hesitant to include it. So we have made an addition to the Zammad REST API which is called `/api/v1/insert_mail` which allows for the very simple use case of the MTA inserting a mail in to Zammad. It is basically `Channel::Driver::MailStdin` via HTTP REST. 

The advantages over batch processing are:

* It's realtime, no waiting for the batch processing.
* Less moving parts. If delivery fails the MTA can alert the sender the proper way.
* No extra staging area needed. No extra IO overhead.

Would you be open to a pull request for that functionality?

**The long story why I personally don't like batch jobs for this**:

* They introduce a delay. For example sometimes agents have the customer on the phone and ask them to mail something; People don't expect a 5 minute delay in delivery (even though I realize mail is not instant messaging, people tend to expect a speedy delivery).
* Batch jobs are extra moving parts which would require monitoring; If the batch job breaks down there would be no immediate sign of the breakdown. The MTA will tell the sender the mail has been delivered yet it is actually sitting unprocessed in some directory.
* When using Zammad in a container there is also a small chance on container redeployment mail is still in this staging area when the container is destroyed and recreated.

thorsteneckel:
Hi @siepkes - We don't think this is a feature for the majority of users. One of Zammad's main goals is to be slim. That's why we would be happy to see an extension offering this feature. We are happy to support you in creating an extension.

siepkes:
@thorsteneckel I'm a bit confused by what you see as an extension; Do you mean if we implement the Zammad functionality to register it as an extension (which allows one to enable and disable the functionality) you would accept a PR for it? As far as I can tell all extension live in the main code base (sipgate, etc)?

thorsteneckel:
Hi @siepkes - It's possible to extend Zammad via packages in the admin area:
localhost_3000_ 1

You can add or modify any files you want. Note that I recommend to only add files and discuss needed changes with us. There might be a general need for an API to allow plugging in via an extension that we can provide in the core.

So you can add your own route file and controller, put it into an package and you should be good to go and version stable.
If you are interested I can provide you some information on how to proceed. There is currently no official documentation for this but it would be great to provide it for everyone.

Issue #295, opened 2016-10-25
Add BCC to Email-Ticket


tobibe:
Sometimes we like to inform a person about an answer without knowlege of customer. So it would be nice, if we could send it by bcc. Both customer and the person have no access to Zammad.

thorsteneckel:
Hey there! We still think that BCC is only a workaround for many use cases and should be avoided. We haven't seen a use case yet that shouldn't be solved in a different way. Therefore I'll lock this issue. Please let us know your use case over at the [community board](http://community.zammad.org/) in a dedicated thread or re-use [the existing thread](https://community.zammad.org/t/bcc-for-outgoing-emails/3289). Attention: Please make sure to use  the search before creating a new thread, we might already have a solution for your use case.

PR #1040, opened 2017-05-04
Add BCC field

MarkusHarmsen:
**BCC:** We had the need of an BCC field for emails, so that as been added. This has already been filed as an issue #295.

I could not figure out how the zammad test "policy" is or where a possible spec could be implemented. Any hints on that topic?

### DIFF
--- a/app/assets/javascripts/app/controllers/agent_ticket_create.coffee
+++ b/app/assets/javascripts/app/controllers/agent_ticket_create.coffee
@@ -24,7 +24,7 @@ class App.TicketCreate extends App.Controller
       @buildScreen(params)
     @bindId = App.TicketCreateCollection.one(load)
 
-    # rerender view, e. g. on langauge change
+    # rerender view, e. g. on language change
     @bind 'ui:rerender', =>
       return if !@authenticateCheck()
       @render()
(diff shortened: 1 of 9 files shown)


monotek:
Please add the vagrantfile to the vagrantfile repo.
Maybe modify the existing one to change zammad environment by variables...

MarkusHarmsen:
@monotek good point, I removed it from this pull request.

MarkusHarmsen:
Sorry there is a failing spec - I will fix that tomorrow!

MarkusHarmsen:
The Mysql version is still failing - it seems that not all migrations have been executed. Speaking about migrations: the error was this line I added to the new migration `return if !Setting.find_by(name: 'system_init_done')` which prevented an execution of the needed migration.
It seems that everything is expected to be in the `CreateTicket` migration - where is the right place to put the `bcc` migration?

thorsteneckel:
Hi @MarkusHarmsen - sorry for the delay. We took our time to discuss this internally. After a lot of back and forth we decided to not include BCC support on the current stage of Zammad. The cause of this is that Zammad is currently designed around processing email communication which will change very soon. The BCC implementation will cause of lot of new special email code and conditions in the current stage of Zammad that will slow us down while migrating and is error prone.

This does not mean that there will be never a BCC functionality in Zammad. It's only for now.

Thanks a lot for making the effort and a lot of <3

Closing now.

aparcar:
Is it possible to set a default BCC somewhere so all send mail get stored somewhere in IMAP readable format?

dkehne:
BCC would be a feature for us to integrate Zammad into our CRM-processes and this is currently the only barrier to implement zammad in our company.  

Feature request: Global BCC per User - or individual BCC.

thorsteneckel:
Hey there! We still think that BCC is only a workaround for many use cases and should be avoided. We haven't seen a use case yet that shouldn't be solved in a different way. Therefore I'll lock this issue. Please let us know your use case over at the [community board](http://community.zammad.org/) in a dedicated thread or re-use [the existing thread](https://community.zammad.org/t/bcc-for-outgoing-emails/3289). Attention: Please make sure to use  the search before creating a new thread, we might already have a solution for your use case>

PR #1118, opened 2017-05-18
No longer raise exception when scheduler job fails to run

MaartenR:
We have been experiencing some problems with errors occurring in scheduler jobs. We would for instance encounter something like this in our logs:

```
/opt/zammad/app/models/scheduler.rb:114:in `rescue in _start_job': STOP thread for HttpLog.cleanup after 10 tries (RuntimeError)
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:112:in `rescue in _start_job'
        from /opt/zammad/app/models/scheduler.rb:86:in `_start_job'
        from /opt/zammad/app/models/scheduler.rb:73:in `block in start_job'
```

After something like this happens the scheduler thread exits and future background jobs are no longer executed. This has happened quite a few times (not always the same error) since we have gone live with Zammad. Therefore we think it would be nice if the scheduler would no longer exit on an error occurring in a single job, but instead would only stop trying to start this specific job. This pull request contains a proposition to no longer raise an exception when an error occurs, but instead record the error and stop all attempts to start this job. The user is informed about failing jobs through the monitoring tool. To make it possible for an administrator to attempt to restart the job without having to edit database columns a button was added that appears when issues. Clicking this button will mark dead jobs for a restart if there are any.

So this pull request consists of two parts. The first is no longer throwing an exception when an error occurs in a scheduler to prevent the scheduler thread from exiting, which is our primary concern. The second is the button to restart dead jobs.

I'm curious to hear your thoughts on this.

### DIFF
--- a/app/assets/javascripts/app/controllers/monitoring.coffee
+++ b/app/assets/javascripts/app/controllers/monitoring.coffee
@@ -4,6 +4,7 @@ class Index extends App.ControllerSubContent
   events:
     'click .js-resetToken': 'resetToken'
     'click .js-select': 'selectAll'
+    'click .js-restartDeadJobs': 'restartDeadJobs'
 
   constructor: ->
     super
@@ -29,7 +30,10 @@ class Index extends App.ControllerSubContent
     )
 
   render: =>
-    @html App.view('monitoring')(data: @data)
+    @html App.view('monitoring')(
+      data: @data
+      job_restart_count: @job_restart_count
+    )
 
   resetToken: (e) =>
     e.preventDefault()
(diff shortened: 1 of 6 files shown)


rolfschmidt:
Hi @MaartenR ,
thank you very much for your support and your pull request! Your pull request has been merged.

Best regards,
Rolf

PR #1174, opened 2017-06-09
Add subcategories at the overview

Jonske:
It would be awesome, if there is a way to add some subcategories under an overview-list. So you can create an Overview for example for a company and add some subcategories for the departments.


thorsteneckel:
Hi @Jonske - Seems there went something wrong while creating the pull request. Please do the following:

- Update your Zammad fork so that develop is at the same point as this repository.
- Create a new branch with your changes. 
- Create your changes.
- Push your branch back to your Zammad fork.
- Create a new pull request for your changes.

Thanks for contributing 😍

PR #1249, opened 2017-07-13
Do not fail Zendesk import if subject is not set

luqasn:
Fall back to description or hard-coded title "No title" instead if ticket has no subject in Zendesk (which is allowed).

### DIFF
--- a/lib/import/zendesk/ticket.rb
+++ b/lib/import/zendesk/ticket.rb
@@ -39,7 +39,7 @@ module Import
 
         {
           id:                       ticket.id,
-          title:                    ticket.subject,
+          title:                    ticket.subject || ticket.description || "No title",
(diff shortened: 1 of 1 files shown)


luqasn:
Should probably add a test for this, but I think you need to create a title without a title in your zendesk testing account. Not sure if that is possible via the normal UI, we created these tickets without subject via the API.

luqasn:
Not sure if you want the hard-coded english string "No title" there, but I figured an ugly last resort is better than an import failing completely, especially because you have to wipe your database in order to start over.

thorsteneckel:
Hi @luqasn - Nice! I'll have a look at it this week.

thorsteneckel:
Hi @luqasn - Thanks! I merged your commit and changed the default title to a `-` as Zammad usually does.

PR #1300, opened 2017-07-26
fix #1292 quick&dirty

### LINKED ISSUE #1292: Don't use Proxy on localhost Address.
### Infos:

* Used Zammad version: 1.6.1
* Used Zammad installation source: (source, package, ...) debian repo
* Operating system: Debian
* Browser + version: irrelevant

### Expected behavior:
In example of running elasticsearch on localhost and using localhost or 127.0.0.1 as Adress, no proxy should be used because localhost mustn't be connected over Proxy.

* 

### Actual behavior:
It connects via my Proxy Server from which i got errors, because 127.0.0.1 is not a good Ip for using Proxy.
* 

Set Proxy under System -> Proxy and try something on localhost via zammad, e.g. using elasticsearch or setting up elasticsearch

*

jaykijay:
This will fix issue #1292 proxy-localhost quick& dirty. Not the nice way, but works.
Happy to see commits to make this look good.

### DIFF
--- a/lib/user_agent.rb
+++ b/lib/user_agent.rb
@@ -280,8 +280,11 @@ returns
       if proxy_password.blank?
         proxy_password = nil
       end
-
-      http = Net::HTTP::Proxy(proxy_host, proxy_port, proxy_username, proxy_password).new(uri.host, uri.port)
+                if (!uri.host.include?("127.0.0.1") && !uri.host.include?("localhost"))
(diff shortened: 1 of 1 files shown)


martini:
Wunderfull! ❤️

martini:
We reworked it and the outcome is the following: https://github.com/zammad/zammad/commit/5568d599ba0b0cba940c5550a298a64de439ce23

Thanks for the hint!

PR #1380, opened 2017-08-31
Zammad systemd support for websocket-server.rb and scheduler.rb

HLFH:


chocochrispi:
Great! thanks

HLFH:
@rkaldung I prefer to separate the _zammad_ user settings of the _zammad_ git project itself.
So, with these new commits, the unique location of the `zammad` user & project is /srv/http/zammad.

> The /srv/ directory contains site-specific data served by a GNU/Linux operating system. This directory gives users the location of data files for a particular service, such as FTP, SFTP or HTTP.

But it you prefer `/home/zammad` to `/srv/http/zammad` or `/var/www/zammad`, please tell me.

Here, using /srv/http is an ArchLinux tradition: it is to me the most rational choice.

monotek:
Home directory should reflect source install docs: https://docs.zammad.org/en/latest/install-source.html

Also i don't know where you take the .rbenv stuff from?
Its not part of zammads git repo and will not work out of the box.

HLFH:
`/opt` directory could be recommended for desktop applications but not for webapps.
In a production environment, you cannont consider using directly `ruby` since you need more than one ruby version and more than one rails version. So, `rbenv` is the best tool to manage your app's Ruby environment.

The [install doc](https://github.com/zammad/zammad-documentation/blob/master/install-source.rst) should be updated with a better location for the webapp: either `var/www` either `/srv/http`.

monotek:
Even our DEB & RPM packages are installed to /opt so we wont change this without good reason.

rbenv is something which should be discussed in a separate pull request.

I also don't like to have several systemd units to start.
Imho there should be:

zammad 
zammad-rails
zammad-scheduler
zammad-websocket

When zammad unit is used the other 3 units should be started / stoped, like in our DEB/RPM packages.

HLFH:
Your DEB & RPM packages are badly optimized regarding `systemd`/ and the install folder.

> rbenv is something which should be discussed in a separate pull request.

You only need zammad-scheduler.service and zammad-websocket.service. Zammad is handled by either `apache` or `nginx` web server, with [Phusion Passenger](https://www.phusionpassenger.com/library/).

You need better `nginx` and `apache` contrib files because it's not optimized.

Good luck.

monotek:
We also don't support phusion passenger so adding systems files which only fits your needs does not make a lot of sense to us. 

We still like the idea to have systemd units which are usable with our standards.

HLFH:
> which only fits your needs

You don't support production servers. We got it.

Don't hesitate to add `systemd` units that meet your development needs.

monotek:
Sure we do. With our official DEB & RPM packages.

Using them also means not to have a compiler one your production machine, which is seen as s good thing among people running production machines ;-)

monotek:
Added own implementation now: https://github.com/zammad/zammad/commit/7c24141cdf18971b8d7624b08d950c014ffd0160

Issue #656, opened 2017-01-20
Shortcuts in different keyboard layout


tojaskiba:
* Used Zammad version: 1.2.0
* Operating system: Windows 7 Proffesional
* Browser + version: Chromium 57
* Ticket-ID: #1048406, #1076974, #1081532, #10106475, #10115149

When using Polish keyboard, using special character like ę with alt+e triggers logout shortcut.
Maybe zammad should use ctrl + shift?

skibson:
What is the best way to turn off shortcuts system wide? Maybe such option should appear also in every user's profile settings menu?

maege:
Same Problem by typing @ in swiss german layout -> (Ctrl + Alt + 2) will result in h2 title.

skibson:
i fixed this in our installation by changing line 127 in app/assets/javascripts/app/lib/base/jquery.contenteditable.js
from e.altKey to e.shiftKey

martini:
The current idea is to move to different shortcuts for windows/mac:

mac: ctrl+alt
windows: ctrl+shift

Any feedback about this? @johan12345

johan12345:
Yeah, changing to Ctrl+Shift would probably be more conflict-free solution than the current one. Some shortcuts could even be done just with ctrl (for example, Ctrl+B, Ctrl+U, Ctrl+I are quite common for text formatting, and the same thing on Mac with Cmd+B, Cmd+U, Cmd+I).

The most flexible way would of course be to make the shortcuts user-configurable, but that's also more time-consuming to implement.

monotek:
@martini 
Why not "ctrl + shift" everywhere?

martini:
> for example, Ctrl+B, Ctrl+U, Ctrl+I are quite common for text formatting, and the same thing on Mac with Cmd+B, Cmd+U, Cmd+I

Thanks a good idea! Done.

> Why not "ctrl + shift" everywhere?

There are some native mac shortcuts on mac with ctrl+shift which conflict then.

Shortcuts are now dependent on OS (in develop branch). Feedback if there are still conflicts would be good!

Mac:
<img width="746" alt="screen shot 2017-04-10 at 00 01 45" src="https://cloud.githubusercontent.com/assets/110226/24841283/722263c6-1d81-11e7-8b19-a8b1ef014dd2.png">

Win:
[...]

PR #1393, opened 2017-09-08
fixed shortcut key typo in translation

NeverMin:
ctrl+alt+t -> shift+ctrl+t


monotek:
shift+ctrl+t is a chrome shortcut too.
It reopens closed tabs.

thorsteneckel:
Hi @NeverMin - Thanks for making the effort! There is currently a bigger issue with the keyboard shortcuts - namely #656 . Since we have no solution for all platforms there is no benefit from changing from one issue to the other. So I'm closing the PR.

PR #1435, opened 2017-09-17
add auth0 authentication

muhammadn:
This is the code to include the auth0 authentication.

You will need to run `rake db:seed` again to create the necessary data for Auth0 in the settings page.


Gemfile.lock
```diff
@@ -277,13 +279,14 @@ GEM
     omniauth-twitter (1.4.0)
       omniauth-oauth (~> 1.1)
       rack
+    parallel (1.12.0)
     parser (2.4.0.0)
       ast (~> 2.2)
-    pg (0.20.0)
+    pg (0.21.0)
```
thorsteneckel: Please don't perform `bundle update` since some dependency versions (e.g. `pg`) are currently broken 😞

.rubocop.yml
```diff
@@ -62,7 +62,7 @@ Style/LeadingCommentSpace:
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#hash-space'
   Enabled: false
 
-Style/MethodCallParentheses:
+Style/MethodCallWithoutArgsParentheses:
```
thorsteneckel: Changes to this file won't be necessary after the previous `rubocop` version is used.

Gemfile.lock
muhammadn: Okay, sorry about that. Do you want me to make another new PR instead?

.rubocop.yml
muhammadn: will remove that.

muhammadn:
I have created an isolated temporary endpoint here (which i will remove after you have tested):

site: https://taqisystems.auth0.com
app id: ujG5FREzJFykqLo4Z9osd6ASal4qcIS4
secret: 4Eoex-JyaB1UJUX4-CziKX-ysRllzCxlLi564kLwBRTWhW0Iwf7r-K6yXMNsAxpS

username: test@zammad.com
password: password123

thorsteneckel:
Hi @muhammadn - Sorry for the late response. I discussed this internally and we came to a similar conclusion as for another Pull Request I answered earlier. This is clearly a similar place where we need an interface to extend the functionality. Pull requests would be great for this!

Even if auth0 and OpenID are used occasionally, this strategy is not widely used. In addition, this means additional dependencies and maintenance that cannot be borne.

We must therefore refrain from this function in the standard system. However, an extension would be exactly the right application. For this, however, the extensibility described above must first be implemented.

xeruf:
Any progress on the extensibility front?

Is this tracked anywhere?

PR #1443, opened 2017-09-18
Show customer and agent tickets respectively when creating new ticket…

### LINKED ISSUE #1354: Create Ticket Button not intuitive
### Infos:

* Used Zammad version: 1.6.1
* Used Zammad installation source: (source, package, ...)
* Operating system: 
* Browser + version: 

### Expected behavior:
The label of the new ticket buttons should distinguish.
Right now both a labled identical
* 

### Actual behavior:
Zammad is used for different groups in the company.
i.e. IT-helpdesk and Human ressources.
All users are member of the ldap group "all_users" and that group is associated with the role "customer".
The users that are responsible for IT-helpdesk are also member of the ldap group "zammd-it-helpdesk" which is associated to the role "IT-Helpdesk agent".

If these it-helpdesk agents press the "+" button to create a new ticket, they get two buttons called "new ticket".
One button links to  the URL: 
http://zammad/#ticket/create
The other button links to the URL:
http://zammad/#customer_ticket_new

It would be useful to have different names for the buttons.

* 

clipboard02

*

muhammadn:
…. Fixes #1354

### DIFF
--- a/app/assets/javascripts/app/controllers/agent_ticket_create.coffee
+++ b/app/assets/javascripts/app/controllers/agent_ticket_create.coffee
@@ -573,4 +573,4 @@ App.Config.set('ticket/create/:ticket_id/:article_id', Router, 'Routes')
 App.Config.set('ticket/create/id/:id/:ticket_id/:article_id', Router, 'Routes')
 
 # set new actions
-App.Config.set('TicketCreate', { prio: 8003, parent: '#new', name: 'New Ticket', translate: true, target: '#ticket/create', permission: ['ticket.agent'], divider: true }, 'NavBarRight')
+App.Config.set('TicketCreate', { prio: 8003, parent: '#new', name: 'New Agent Ticket', translate: true, target: '#ticket/create', permission: ['ticket.agent'], divider: true }, 'NavBarRight')
--- a/app/assets/javascripts/app/controllers/customer_ticket_create.coffee
+++ b/app/assets/javascripts/app/controllers/customer_ticket_create.coffee
@@ -189,4 +189,4 @@ class Index extends App.ControllerContent
       )
 
 App.Config.set('customer_ticket_new', Index, 'Routes')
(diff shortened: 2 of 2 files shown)


thorsteneckel:
Hi @muhammadn ! Great! Thanks for creating this pull request. I think it's one step in the right direction but not the right solution. It doesn't fit for regular agents/tickets that are only in one role. In this case it's to much information and loses the simple elegance.

The behavior should be:

- Agent: `Create Ticket`
- Customer: `Create Ticket`
- Agent + Customer: `Create agent Ticket` and `Create customer Ticket`

Currently I can't see a straight way of doing this since the check for a certain permission is [locked in the frontend ApplicationController](https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/controllers/_application_controller.coffee#L239) which is not accessible from the outer scope.

I would recommend to extract this somewhere into the `App.` scope so it can be accessed. 
ApplicationController should then tunnel calls to this extracted method. What do you think @martini ?

muhammadn:
I think your idea is the way to go and i have to pick up Spine client-side MVC (i came from AngularJS/React and spine is totally new thing)

But i get what you mean. Currently the code runs in App.Controller and i have to extract   `permissionCheck: (key) -> ....` into `App.` so it somehow become in another word, "global" to be accessed from anywhere.

thorsteneckel:
Sorry for the late response. There should be a method called `permitted` on the [Permission model](https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/models/permission.coffee). You can define it like [e.g. App.Group.accesses()](https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/models/group.coffee#L39).

The content should be the same like in [permissionCheck](https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/controllers/_application_controller.coffee#L239). Actually `permissionCheck` should then call `App.Permission.permitted` internally.

If there are any questions feel free to ask. I am very happy about your support! 🚀

thorsteneckel:
Hi @muhammadn - are there any updates on this?

muhammadn:
Was busy lately with my day job. Sorry about that. I think i will close this first to work on it, don't want to pile up the pull requests, will re-open when i get to the result.

thorsteneckel:
As stated in #1354 this is currently no use case. Closing now.

Issue #1352, opened 2017-08-19
additional attribute type when creating objects


gwingend:
### Infos:

* Used Zammad version: 
* Used Zammad installation source: (source, package, ...)
* Operating system: 
* Browser + version: 

### Expected behavior:
When creating attributes it would be useful to have a text attribute (or any other attribute) with the IP-address format for IPv4 and IPv6
I.e. if you want the customer to make a chance request for an firewall rule by ticket, source and destination addresses should be entered. 
With the choice IP-address also a field-validation should be implemented, so only valid ip addresses could be entered.

* 

### Actual behavior:
Right now there is the choice between "email", "phone", "text" and "URL".
* 

### Steps to reproduce the behavior:

*

hanneshal:
Hi @gwingend 

small update. We just rely on the HTML 5 input types. Sadly there is no IP address type in the spec. So a custom validation would be needed. Which is much more effort.

https://www.w3schools.com/html/html_form_input_types.asp
> HTML5 added several new input types:
> 
> color
> date
> datetime-local
> email
> month
> number
> range
> search
> tel
> time
> url
> week
>

muhammadn:
@hanneshal @gwingend Can i know where is the form at?  Which version of zammad is this?

Sample code which i had written is here using HTML5 pattern.
```
<html>
  <form>
    <input id="ip" type="string" pattern="((^|\.)((25[0-5])|(2[0-4]\d)|(1\d\d)|([1-9]?\d))){4}$|((^|:)([0-9a-fA-F]{0,4})){1,8}$" required>
    <button>Submit</button>
  </form>
  <script>
    var ipAddress = document.getElementById("ip");

    ipAddress.oninvalid = function(event) {
      event.target.setCustomValidity('Enter a valid IPv4 or IPv6 address');
    }
  </script>
</html>
```

PR #1447, opened 2017-09-19
add IP address attribute with input validation for IPv6 and IPv4 addr…

muhammadn:
…esses #1352

This code is about implementing pattern IP address attribute with pattern matching for input validation.


thorsteneckel:
Hi @muhammadn - Thanks for this nice Pull Request. Unfortunately, it shows the limitation of extensibility. We should first focus on making it easier to add new field types without changing existing files. Pull requests would be great for this!

Even if IP addresses are a comprehensible application, this applies to a small range of installations. We must therefore refrain from this function in the standard system. However, an extension would be exactly the right application. For this, however, the extensibility described above must first be implemented.

PR #1451, opened 2017-09-20
Openid connect

muhammadn:
This is the integration of OpenID Connect gem (this gem is not published so i had to use the github version at https://github.com/sdoxsee/omniauth-oidc) 

After that you can configure OpenID Connect as follows to test which is a isolated temporary setup:

site: https://taqisystems.auth0.com/
app_id : ujG5FREzJFykqLo4Z9osd6ASal4qcIS4
secret: 4Eoex-JyaB1UJUX4-CziKX-ysRllzCxlLi564kLwBRTWhW0Iwf7r-K6yXMNsAxpS
authorize_url: /authorize
token_url: /oauth/token


thorsteneckel:
Hi @muhammadn - Sorry for the late response. I discussed this internally and we came to a similar conclusion as for another Pull Request I answered earlier. This is clearly a similar place where we need an interface to extend the functionality. Pull requests would be great for this!

Even if auth0 and OpenID are used occasionally, this strategy is not widely used. In addition, this means additional dependencies and maintenance that cannot be borne.

We must therefore refrain from this function in the standard system. However, an extension would be exactly the right application. For this, however, the extensibility described above must first be implemented.

Issue #757, opened 2017-02-13
Integrating zammad into our android app


KirkBushman:
Hi, just a quick info:
I'm trying to integrate Zammad into our internal ticketing system (an android app - Java), and I already did part of the work, but there is a bit that I dont understand.

How do I List all the tickets, filtered by states?

I can use a simple http request to /api/v1/tickets, but then I receive all the tickets in my groups.

What if I wanted to use a tab layout of states and when one is pressed I GET the json only for the "open" ones? or "closed successfully" ones? how?

In case we have a lot of tickets, it might take a lot to download the json for all tickets every time (even more on a mobile connection)...
Can I order them by Creation Date and then get the ones, that I dont already have in memory? how?

thanks in advance

thorsteneckel:
Sorry to tell @KirkBushman but there is currently no way to perform this action properly since the self made ElasticSearch method doesn't provide it. Nevertheless I agree with you that this should be possible. Therefore I classify this issue as a bug.

muhammadn:
@KirkBushman I have checked the current development code already has paging so not all tickets will be shown. I just added to endpoints /api/v1/tickets/open (for open tickets) and /api/v1/tickets/closed (for closed tickets) if this is what you are looking for.

thorsteneckel:
Hi @KirkBushman - It's been a while. Since it's not noted here I just want to make sure and ask: Have you tried using the `/api/v1/tickets/search` endpoint and use the `conditions` parameter? See: https://github.com/zammad/zammad/blob/develop/app/controllers/tickets_controller.rb#L445

KirkBushman:
I forgot that this was here...
Yes, I used that to integrate Zammad in our app, and it's working really well.

thorsteneckel:
Awesome to hear! Thanks for the fast feedback. I'll close this now.

PR #1461, opened 2017-09-21
add new endpoints tickets/open and tickets/closed so user has a choic…

muhammadn:
…e to get open or closed tickets #757

Add new endpoints for either open or closed tickets, sorted using closed_at column by checking NULL value.

/api/v1/tickets/closed for closed tickets
/api/v1/tickets/open for open tickets

### DIFF
--- a/app/controllers/concerns/ticket_sort.rb
+++ b/app/controllers/concerns/ticket_sort.rb
@@ -1,0 +1,45 @@
+module TicketSort
+  extend ActiveSupport::Concern
+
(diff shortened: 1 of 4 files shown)


thorsteneckel:
Hi @muhammadn - Interesting implementation! As I tried to describe in the issue, we should improve the search interface of our Elasticsearch connection. Here, pull requests would be great.

This would enable us to cover such and many other requirements and to keep the API manageable and clean.

PR #1463, opened 2017-09-23
Adds customer_ticket_create requirement to enable form

### LINKED ISSUE #1339: Deny ticket creation over webinterface for customers doesn't work
### Infos:

* Used Zammad version: 1.5.0-1501623470.50f2d496.jessie
* Used Zammad installation source: debian package
* Operating system: Debian 8.9
* Browser + version: latest FX and Chrome

### Expected behavior:

* Customer is not allowed to create a ticket over webinterface

### Actual behavior:

* Customer IS allowed to create a ticket over webinterface while config is set to "no" under channels --> web - activate ticket creation (german translation: Kanäle -> Web -> Aktivierung der Ticketerstellung)
Also mixing around with the second option, which defines groups for which customer can create tickets, doesn't have any effect. 
At last i miss short info that settings are saved when pressing the submit-button.

* deny ticket creation over webinterface for customers. login with customer credentials and see if you can create a ticket.

bwowbagger:
Fixes #1339 

Enable ticket creation at Admin -> Channels -> Web now enables/disables the web form.
Requires `customer_ticket_create` to be `true` to consider the Web form enabled.

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -241,7 +241,7 @@ class FormController < ApplicationController
 
   def enabled?
     return true if params[:test] && current_user && current_user.permissions?('admin.channel_formular')
-    return true if Setting.get('form_ticket_create')
+    return true if Setting.get('form_ticket_create') && Setting.get('customer_ticket_create')
     response_access_deny
     false
   end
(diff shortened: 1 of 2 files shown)


bwowbagger:
I'm assuming Enable Ticket creation at Admin -> Channels -> Web has the same expected behavior as the form enable switch at Admin -> Channels -> Form.

thorsteneckel:
Merged with applied rubocop. Would be great if you could apply rubocop before creating a pull request 💪

PR #1464, opened 2017-09-24
added style to ticket_create

adhamkurniawan:
I added some default style css for related issue "Notification E-Mail Template" in ticket_create

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,7 @@ gem 'zendesk_api'
 gem 'viewpoint'
 gem 'rubyntlm', git: 'https://github.com/wimm/rubyntlm.git'
 gem 'autodiscover', git: 'https://github.com/thorsteneckel/autodiscover.git'
+gem 'wysiwyg-rails'
 
 # event machine
 gem 'eventmachine'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -151,6 +151,8 @@ GEM
       rake
     figaro (1.1.1)
       thor (~> 0.14)
+    font-awesome-sass (4.7.0)
+      sass (>= 3.2)
     formatador (0.2.5)
     github_changelog_generator (1.14.3)
       activesupport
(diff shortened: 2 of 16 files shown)


app/assets/javascripts/app/views/agent_ticket_create.jst.eco
```diff
@@ -30,7 +30,7 @@
             <input type="hidden" name="form_id" value="<%= @form_id %>"/>
 
             <div class="ticket-form-top"></div>
-            <div class="article-form-top"></div>
+            <div id="editor" class="article-form-top"></div>
```
martini: We can't work with id here. Because the ticket create screen can be opened many times as you want.

Gemfile
```diff
@@ -85,6 +85,7 @@ gem 'zendesk_api'
 gem 'viewpoint'
 gem 'rubyntlm', git: 'https://github.com/wimm/rubyntlm.git'
 gem 'autodiscover', git: 'https://github.com/thorsteneckel/autodiscover.git'
+gem 'wysiwyg-rails'
```
martini: Since Zammad is a web app, we try to do all dependencies for the web app not in rails context (we add no gems for adding java script libs). The reason is, in one of our next steps we want to separate the REST store from the web app.

Gemfile.lock
```diff
@@ -151,6 +151,8 @@ GEM
       rake
     figaro (1.1.1)
       thor (~> 0.14)
+    font-awesome-sass (4.7.0)
```
martini: We do not use font-awesome right now. We do not want to blow Zammad right now with additional forms.

app/assets/javascripts/application.js
```diff
@@ -42,6 +42,8 @@
 //= require ./app/lib/rangy/rangy-textrange.js
 //= require ./app/lib/rangy/rangy-highlighter.js
 
+//= require froala_editor.min.js
+
```
martini: Since Zammad is open source and Froala Editor is not (https://www.froala.com/wysiwyg-editor/pricing) we can't use Froala Editor anyway.

martini:
> I added some default style css for related issue "Notification E-Mail Template" in ticket_create

That is also not clear to me, is it related to an existing issue?

PR #1466, opened 2017-09-24
Add code to support regex: for icinga and nagios email address

muhammadn:
Code to check for regex:
Check for 'regex:' field, if available, set regex = true and set the match_rule on the second regex code (after the regex: field)

If there is no regex:, try to match **exact** email address.

Then lastly check for regex == true and if true,  match the regex for the match_rule regex code.

### DIFF
--- a/app/models/channel/filter/monitoring_base.rb
+++ b/app/models/channel/filter/monitoring_base.rb
@@ -26,6 +26,7 @@ class Channel::Filter::MonitoringBase
     return if !session_user_id
 
     # check if sender is monitoring
+    regex = false
     if sender =~ /^(regex:)(.+?)$/ # check if sender setting has regex:
       regex = true
       match_rule = $2 # assign the second regex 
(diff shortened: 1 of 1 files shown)


muhammadn:
@martini I have made some changes that instead of having the .match method in Channel::Filter::Database, i feel i should extract that out into a separate concern so it won't look messy.

below code EmailRegex is actually email_regex.rb in models/concern with the .match method code.

What do you think about it?

```
 class Channel::Filter::MonitoringBase
-
+  include EmailRegex
   # according
 
   # Icinga
@@ -27,7 +27,7 @@ class Channel::Filter::MonitoringBase
     return if !session_user_id
 
     # check if sender is monitoring
-    return if !Channel::Filter::Database.match(mail[:from], sender, true, true)
+    return if !EmailRegex.match(mail[:from], sender, true, true)
```

martini:
> @martini I have made some changes that instead of having the .match method in Channel::Filter::Database, i feel i should extract that out into a separate concern so it won't look messy.

That's good! We could also create separat tests for it. I like it!

PR #1468, opened 2017-09-25
[WIP] Bug fix/block customer reopen

### LINKED ISSUE #933: Don't allow reopen a closed ticket over the gui
### Infos:

* Used Zammad version: newest
* Operating system: debian
* Browser + version: firefox

### Expected behavior:

* when a ticket is closed we won't that customer can reopen ticket over the web gui. Is there planed anything?

### Actual behavior:

* 

*

bwowbagger:
Fixes #933 

Validates if customer is opening a closed ticket and renders a 401. 

screenshot 2017-09-24 23 47 37

The UX should also be changed to filter out the `open` state option which makes this a WIP.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -5,6 +5,7 @@ class TicketsController < ApplicationController
   include TicketStats
 
   prepend_before_action :authentication_check
+  before_action :reopen_role_validation, only: :update
 
   # GET /api/v1/tickets
   def index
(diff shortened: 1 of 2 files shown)


muhammadn:
not that too important but I think you should return a 422 instead of 401?

thorsteneckel:
Hi @bwowbagger - Are you planning to work on this? I would like to start reviewing it before the branches diverge to much. A short feedback about the current state and further plans would be great.

bwowbagger:
@thorsteneckel sure, I can send an update over. The current status is to find the best way to filter the open option in the front end. Either find the best option in the Backend to not serve this state or have the front end not request it. What do you think? I'll also change the error as per @muhammadn suggestion

thorsteneckel:
Thanks for your fast reply @bwowbagger ! We discussed this and described the desired solution in the issue #933 so it's more accessible.

Since this implementation derives from the desired solution I'll have to close this pull request. Feel free to discuss in the issue and/or create a new pull request. Would be great!

PS: Never use direct ID checks since Zammad is configurable. If the admin changes names and/or types this will break. Always check for types which and fetch those by name. These are static.

PR #1476, opened 2017-09-26
fixed #1475

### LINKED ISSUE #1475: login button
login button

NeverMin:
HI @mrflix , this is little changed, but will be change all `.btn` style, maybe create other one, e.g. `btn-login`? please let me know, thank you for your time.

login button now

### DIFF
--- a/app/assets/stylesheets/zammad.scss
+++ b/app/assets/stylesheets/zammad.scss
@@ -366,7 +366,7 @@ pre code.hljs {
 .btn {
   display: inline-block;
   font-size: 14px;
-  padding: 10px 24px 9px;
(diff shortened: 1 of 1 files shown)


mrflix:
Hey @NeverMin. Thanks for contributing to Zammad! Its a great idea to align the login buttons. As you suggest creating a specific style for this case makes more sense than changing the button padding globally. The `.auth-provider` buttons have a width of `calc(33% - 6px)` (one third of the width minus some padding) so I adjusted the css to give the login button the same minimal width (it could get bigger depending on the translation and we don't want to limit that).

NeverMin:
:+1:  Nice.

Issue #1439, opened 2017-09-18
Icinga/Nagios Integration: No longer accepts regex in "Sender"


hiaz:
### Infos:

* Used Zammad version: 2.0
* Used Zammad installation source: package
* Operating system: Ubuntu 16.04

### Expected behavior:

If I use nagios@.* it should work for mails from nagios@customer**A**.com and nagios@customer**B**.com

It worked very well in 1.5. 

### Actual behavior:

In 2.0 it only works for exact matches.

The 1.5 "feature" was very useful for us. If regex is not possible we need the ability to enter more than one sender e-mail address.

natseg:
Hi,

Is this issue not resolved by simply using the wildcard alone? 
<img width="531" alt="screen shot 2017-09-19 at 16 00 19" src="https://user-images.githubusercontent.com/6449818/30599263-d4040b10-9d53-11e7-80f0-8d3364773515.png">
` nagios@*` instead of `nagios@.*`

martini:
@natseg in Zammad under Admin -> Channel -> E-Mail -> Filter (or also for integrations -> icinga or nagios) your can use regular expressions too. 

For Zammad 1.5 and below is was enabled by default. It looked like:

<img width="651" alt="screen shot 2017-09-19 at 18 51 15" src="https://user-images.githubusercontent.com/110226/30604839-47ae4ac2-9d6c-11e7-8d98-5147830fa0d9.png">

The only problem was, many people used simple * for a wildcard. But technically it was used for a regex which was broken then (because with regex you need to use .*).

Anyway, for Zammad 2.0 and higher we changed the syntax it support * as wildcard per default (95% of the users will expect it). If you still want to use regular expressions you need to prefix it with **regex:** in the search field:

<img width="642" alt="screen shot 2017-09-19 at 18 51 45" src="https://user-images.githubusercontent.com/110226/30604984-c9d9fe42-9d6c-11e7-81ef-ef5344e920a8.png">

@hiaz for icinga and nagios integration we missed the support of "regex:" prefix. -> Fix is on the way.

martini:
@hiaz now it's fixed for Zammad 2.0 (just update the package). You can use regex again just enter `regex:nagios@(customerA.com|customerB.com)`

hiaz:
@martini Just did the update. Seems to work fine! Thank you.

martini:
@hiaz Thanks for feedback!

PR #1483, opened 2017-09-27
move out the regex matching code to better readability #1439

muhammadn:
Move out the regex filter code from Channel::Filter::Database into a separate concern with tests included


app/models/concerns/email_regex.rb
```diff
@@ -0,0 +1,28 @@
+module EmailRegex
```
thorsteneckel: Two things:
- Modules should be used for mixins. This case is an isolated logic task which should be a class.
- `EmailRegex` is a to generic name for this specific use case. Please use the namespace `Channel::Filter::Match::EmailRegex` instead.

app/models/concerns/email_regex.rb
```diff
@@ -0,0 +1,28 @@
...
+  extend ActiveSupport::Concern
```
thorsteneckel: Why are you using a concern here?

app/models/concerns/email_regex.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def self.match(value, match_rule, _should_match, check_mode = false)
```
thorsteneckel: Two things:
- Why is there an argument `_should_match` which is never used?
- We should use keyword arguments. A rule of thumb: More than two arguments should always be keyword arguments. This increases the readability a lot.

test/unit/email_regex_test.rb
```diff
@@ -0,0 +1,127 @@
+# encoding: utf-8
+# rubocop:disable all
```
thorsteneckel: Don't disable Rubocop in new files.

test/unit/email_regex_test.rb
```diff
@@ -0,0 +1,127 @@
...
+require 'test_helper'
+
+class EmailRegexTest < ActiveSupport::TestCase
```
thorsteneckel: Just an information: If you prefer RSpec we would gladly see RSpec tests in the future over Minitest.

app/models/concerns/email_regex.rb
muhammadn: yeah. i was using positional arguments and i think keyword arguments will increase readablity. _should_match came from the original method which i forgot to exclude from earlier.

app/models/concerns/email_regex.rb
muhammadn: Shouldn't be there. you are right.

app/models/concerns/email_regex.rb
muhammadn: Yeap. i will namespace that module. will see the new updated code soon.

test/unit/email_regex_test.rb
muhammadn: Sure. i am beginning to like rubocop. Never used it but it is really nice to have the code to be in order.

app/models/channel/filter/database.rb
```diff
@@ -18,12 +18,12 @@ def self.run(_channel, mail)
           match_rule = meta['value']
           min_one_rule_exists = true
           if meta[:operator] == 'contains not'
-            if value.present? && match(value, match_rule, false)
+            if value.present? && Channel::Filter::Match::EmailRegex.match(value: value, match_rule: match_rule, check_mode: false)
```
thorsteneckel: Looks like the change of attributes has introduced a behavior change. The third parameter `false` was previously the now removed `_should_match` but is now `check_mode`.

app/models/channel/filter/database.rb
```diff
@@ -18,12 +18,12 @@ def self.run(_channel, mail)
...
               all_matches_ok = false
               Rails.logger.info "  matching #{key.downcase}:'#{value}' on #{match_rule}, but shoud not"
             end
           elsif meta[:operator] == 'contains'
-            if value.blank? || !match(value, match_rule, true)
+            if value.blank? || !Channel::Filter::Match::EmailRegex.match(value: value, match_rule: match_rule, check_mode: true)
```
thorsteneckel: Looks like the change of attributes has introduced a behavior change. The third parameter `true` was previously the now removed `_should_match` but is now `check_mode`.

app/models/postmaster_filter.rb
```diff
@@ -14,11 +14,7 @@ def validate_condition
       raise Exceptions::UnprocessableEntity, 'operator invalid, ony "contains" and "contains not" is supported' if meta['operator'].blank? || meta['operator'] !~ /^(contains|contains not)$/
       raise Exceptions::UnprocessableEntity, 'value invalid/empty' if meta['value'].blank?
       begin
-        if meta['operator'] == 'contains not'
-          Channel::Filter::Database.match('test content', meta['value'], false, true)
-        else
-          Channel::Filter::Database.match('test content', meta['value'], true, true)
-        end
+        Channel::Filter::Match::EmailRegex.match(value: 'test content', match_rule: meta['value'], check_mode: true)
```
thorsteneckel: nice spot!

thorsteneckel:
Merged with the squash commit. Thanks a lot @muhammadn 🚀

PR #1489, opened 2017-09-29
Tell Internet Explorer to display content in the highest mode available.

IAmWebSA:
This shall solve issues like Company Profiles setting to a lower document mode per default, that prevents the page to be loaded  for e.g.  IE document mode 7 etc.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -3,6 +3,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=1024">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <title><%= Setting.get('product_name') %></title>
   <%= stylesheet_link_tag "application", :media => 'all' %>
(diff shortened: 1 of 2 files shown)


thorsteneckel:
Hi @mrflix / @martini - as stated [here](https://stackoverflow.com/a/6771584/7900866) there should be no need for this if we are only supporting IE 11 and Edge. As far as I can remember we also have the support for IE 10, right? What do you think?

mrflix:
@IAmWebSA what's your use case? Are you working in a company where you use IE10 with a document mode of IE7? We support IE >= 10.

IAmWebSA:
Yes exactly, the mode is set for compatibility reasons in IE11. So telling the mode for IE would be a great addition. To Auto use the latest mode.

IAmWebSA:
I have fixed the intention. Thanks for integrating the change

thorsteneckel:
Merged with the commits above. Thanks for your contribution!

Issue #1481, opened 2017-09-27
Error while receiving email


tobibe:
### Infos:

* Used Zammad version: 2.0.0
* Used Zammad installation source: package
* Operating system: Ubuntu 16.04.03

Zammad could not parse an email and append it to a ticket.
Error in production log and email attached.

Zammad fetches the email from Exchange server 2013 which usually works.

[5592ab0b7df1c71cb8a8b1da9c923195.txt](https://github.com/zammad/zammad/files/1337117/5592ab0b7df1c71cb8a8b1da9c923195.txt)
[production.txt](https://github.com/zammad/zammad/files/1337118/production.txt)

Greetings 
tobibe

muhammadn:
I think it has to do with the MySQL deadlock issue #750 .

@tobibe I had written some tests to check for email parsing problems specifically using the email attachment you provided but the tests passed. 

I have also looked at _process method in email_parser.rb (the filter runs  to check which is what is in the production.txt) and once it hits the `Transaction.execute` block part that updates the user record (it says 200 OK in production.txt) but in the later part the query is being rolled back and then that same query which was successful is having an error with a deadlock.

martini:
@muhammadn thanks for your PR.

For me it's always important got get the issue reproduced first. 

@tobibe Did this issue happen more times or only one time? 

If it has happen more time, how many agents do work with Zammad? And about how many emails do your receive per day?

In general it looks like an concurrency (2 processes will modify a user at some time) issue.

tobibe:
Hello @martini 

I don't know, if this happens more often. We only recognized this problem because we were awaiting this email. 
We receive 50 to 100 mails per day. 
We're still using mysql as database, so it's possibly the same concurrency issue.

I tried to search for this error and found none for email, but different other cases.

[error1.txt](https://github.com/zammad/zammad/files/1446759/error1.txt)
[error2.txt](https://github.com/zammad/zammad/files/1446760/error2.txt)

Greetings
tobibe

thorsteneckel:
Hi @tobibe - can you please verify that this still occurs in the latest version of Zammad?

MrGeneration:
I'm closing this issue, as we haven't received any feedback.
I assume that this problem has been resolved within the meantime. :-)

PR #1493, opened 2017-10-01
Email parser test and avoid parsing errors when it is not

muhammadn:
Check email parsing in test for #1481  and correct a misleading error about email parsing where somethings the error is not related to that.


thorsteneckel:
Hi @muhammadn - Sorry I don't get the issue of this. I don't see the relation between your change and the added test. The test worked for me even without the changes. I'll close this for now. Please create an issue first before creating a Pull Request.

Issue #1495, opened 2017-10-02
Increase users.note & organizations.note from 250 to 5000 chars.


martini:
Increase users.note & organizations.note from 250 to 5000 chars.

PR #1497, opened 2017-10-02
change character length limit in "note" column in both organizations …

muhammadn:
…and users #1495

Current column limit is 250 characters for "note" column field in both organizations and users table. This migration will increase that to 5000 with a way to rollback to 250 characters, if needed.

### DIFF
--- a/db/migrate/20120101000001_create_base.rb
+++ b/db/migrate/20120101000001_create_base.rb
@@ -36,7 +36,7 @@ class CreateBase < ActiveRecord::Migration[4.2]
       t.boolean :vip,                                         default: false
       t.boolean :verified,                        null: false, default: false
       t.boolean :active,                          null: false, default: true
-      t.string :note,                 limit: 250, null: true, default: ''
+      t.string :note,                 limit: 5000, null: true, default: ''
       t.timestamp :last_login,        limit: 3,   null: true
       t.string :source,               limit: 200, null: true
       t.integer :login_failed,                    null: false, default: 0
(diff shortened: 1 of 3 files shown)


martini:
@muhammadn Nice! :) 

I have 3 more things. Background: For client side validation we also have records for 4 models (User, Group, Organisation and Ticket) in ObjectManager::Attribute records (in case you create new object attributes, also new records will be created there).

1) You also need to change it for the web app (client validation) for initial setup in `https://github.com/zammad/zammad/blob/develop/db/seeds/object_manager_attributes.rb#L1080` and `https://github.com/zammad/zammad/blob/develop/db/seeds/object_manager_attributes.rb#L1332` 

2) For upgrades via migration
```
object_id = ObjectLookup.by_name('User')
attribute = ObjectManager::Attribute.find_by(object_lookup_id: object_id, name: 'note')
attribute.data_option[:maxlength] = 5000
attribute.save!

object_id = ObjectLookup.by_name('Organization')
attribute = ObjectManager::Attribute.find_by(object_lookup_id: object_id, name: 'note')
attribute.data_option[:maxlength] = 5000
attribute.save!
```

3) Some day we need to cleanup/merge migrations to keep the migrations clean. To do this easier (I know it's not rails common) we create the migrations only for update and add the following to the migration: https://github.com/zammad/zammad/blob/develop/db/migrate/20170927000001_setting_send_no_auto_response_reg_exp.rb#L5 

 -=> So if we want to cleanup the migrations we just need to remove the old migration without any other todo's/steps.

Would be nice if you could apply this changes too! 👍🏻

muhammadn:
I do it sometimes in adding code into migrations (like running a query to update the data, or update by downloading it and process) but this only happens when i work with PHP (Laravel/Symfony framework) but in rails it can still be done since migrations are basically ruby code.

So basically what you were saying is that if there is 'system_init_done' value in Settings table, the code will treat is as an already installed system and runs the migration. 

If it were a new system, it will get those configuration from the seed but the migration will still set the note columns with a length 250. 

So i have updated `20120101000001_create_base.rb` to set 'note' columns for users and organizations to a limit of 5000 characters as well, for new systems which runs the migration.

martini:
Thanks a lot!

PR #1499, opened 2017-10-03
enable list of tags in tag input without typing two character, fixes #1498

### LINKED ISSUE #1498: Tag Dropdown
### Infos:

* Used Zammad version: 2.x
* Used Zammad installation source: (source, package, ...) :package
* Operating system: Ubunutu 16.04 LTS
* Browser + version: Vivaldi

### Expected behavior:

* We want to have a dropdown list with available tags for tagging tickets, because we don't not want to add new tags by the agents. The agents should know which tags are available instead of "guess" which tags are already implemented.

### Actual behavior:

*  I have to type in the first capital of a tag before i get a list of available tags.

* Try to add an existing tag to a ticket

umarsheikh:
…ow list on focus, fixes #1498


thorsteneckel:
Thanks! Merged.

Issue #1276, opened 2017-07-20
Cannot update notification preferences when nothing is selected


hanneshal:
### Infos:

* Used Zammad version: latest
* Used Zammad installation source: source
* Operating system: OSX
* Browser + version: Chome

### Expected behavior:

* It should be possible to remove all notifications

### Actual behavior:

* If you remove all notifications, the preferences are not updated.

### Steps to reproduce the behavior:

* Remove all checkmarks from the notification matrix

![zammad_helpdesk_-_notifications](https://user-images.githubusercontent.com/804939/28417324-06e6290e-6d58-11e7-99de-4c05eb5f16f3.png)

and hit submit. 
The matrix object is empty and an error is written to the console and no update popover is shown.

PR #1514, opened 2017-10-06
Update notification preferences when nothing is selected #1276

hiendinhngoc:
The validation code, in this case, is wrong because the missing matrix is not error.


martini:
Thanks a lot!

PR #1518, opened 2017-10-08
fix #1509 where two conflicting roles are being saved without the fin…

### LINKED ISSUE #1509: {"error":"Role Customer conflicts with Admin"}
Good morning,

we have updated Zammad to version 2.0.x.

Now, when an agent wants to create a new ticket, a window with StatusCode: 500 {"error":"Role Customer conflicts with Admin"} is popping up. How must the role Agent be modified or have I also enter all Agents to Admin?

Cheers
Patrick

muhammadn:
…al .save!

The conflicting roles were not validated at the backend when associating the user with roles.

Currently the problem is that roles were updated (via db insert) before user record is saved and only after record is saved the error is being shown.

This fix would show the conflicting roles error when roles are updated as well to avoid conflicting roles to mistakenly updated.


app/models/user.rb
```diff
@@ -963,12 +963,21 @@ def ensure_uniq_email
     raise Exceptions::UnprocessableEntity, 'Email address is already used for other user.'
   end
 
-  def validate_roles
+  def validate_ooo
```
thorsteneckel: Could you please restore the method order to remove the noise from the diff? `validate_roles` should come before `validate_ooo`.

app/models/user.rb
```diff
@@ -963,12 +963,21 @@ def ensure_uniq_email
...
+    raise Exceptions::UnprocessableEntity, 'out of office start is required' if out_of_office_start_at.blank?
+    raise Exceptions::UnprocessableEntity, 'out of office end is required' if out_of_office_end_at.blank?
+    raise Exceptions::UnprocessableEntity, 'out of office end is before start' if out_of_office_start_at > out_of_office_end_at
+    raise Exceptions::UnprocessableEntity, 'out of office replacement user is required' if out_of_office_replacement_id.blank?
+    raise Exceptions::UnprocessableEntity, 'out of office no such replacement user' if !User.find_by(id: out_of_office_replacement_id)
+    true
+  end
+
+  def validate_roles(role)
     return true if !role_ids
```
thorsteneckel: Can you please add a comment why we exit if no role_ids are present? It's not instantly clear and should be therefore explained by a comment.

app/models/user.rb
```diff
@@ -963,12 +963,21 @@ def ensure_uniq_email
...
-    role_ids.each do |role_id|
-      role = Role.lookup(id: role_id)
-      raise "Unable to find role for id #{role_id}" if !role
-      next if !role.preferences[:not]
+    role = Role.lookup(id: role.id)
```
thorsteneckel: Why do we need to lookup a role that we already got via the parameter? This is an unnecessary query.

app/models/user.rb
```diff
@@ -963,12 +963,21 @@ def ensure_uniq_email
...
+    raise "Unable to find role for id #{role_id}" if !role
+    if role.preferences[:not].present?
```
thorsteneckel: Please use a guard clause over here: `return true if role.preferences[:not].blank?`

app/models/user.rb
```diff
@@ -963,12 +963,21 @@ def ensure_uniq_email
...
       role.preferences[:not].each do |local_role_name|
         local_role = Role.lookup(name: local_role_name)
         next if !local_role
```
thorsteneckel: I can't comment on the line below since it's the original state. However we should always have refactoring in mind when working on methods. Therefore please change the line below:

```ruby
raise "Role #{role.name} conflicts with #{local_role.name}" if role_ids.include?(local_role.id)
```

To a guard clause which makes the code more readable:

```ruby
next if role_ids.exclude?(local_role.id)
raise "Role #{role.name} conflicts with #{local_role.name}"
```

app/models/user.rb
```diff
@@ -963,17 +963,13 @@ def ensure_uniq_email
...
-        local_role = Role.lookup(name: local_role_name)
-        next if !local_role
-        raise "Role #{role.name} conflicts with #{local_role.name}" if role_ids.include?(local_role.id)
-      end
+  def validate_roles(role)
+    return true if !role_ids # we need role_ids for checking in role_ids below, in this method
+    return true if role.preferences[:not].blank?
+    role.preferences[:not].each do |local_role_name|
+      local_role = Role.lookup(name: local_role_name)
+      next if role_ids.exclude?(local_role.id)
```
thorsteneckel: Somehow the line `next if !local_role` got lost. This leads to a potential bug where the configured `:not` `local_role_name` is not valid and `Role.lookup(name: local_role_name)` returns `nil`. If so accessing `local_role.id` will fail hard. Please reintroduce the guard clause for this case.

muhammadn: yeah you are right. i missed that part if `local_role` is nil. Will add the guard clause.

thorsteneckel:
Merged with the squash commit. Thanks a lot @muhammadn 🚀

PR #1521, opened 2017-10-10
the old branch of forked composite key gem is no longer available, us…

muhammadn:
…e the official gem with tag v10.0.0 for rails 5.1

The composite_key gem from it's official repository has already released an official version to support Rails 5.1 (v10.0.0) and the forked repo branch that zammad uses for composite_key is no longer available (and travis build fails).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,8 +53,7 @@ gem 'mime-types'
 
 gem 'biz'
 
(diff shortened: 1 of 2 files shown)


thorsteneckel:
Hi @muhammadn - Nice spot! I did the change myself by using the dependency as any other directly like `gem 'composite_primary_keys'` and updated the `Gemfile.lock` via `bundle update`. This way we will always get the latest version which is the desired behavior.

Issue #854, opened 2017-03-17
Round Robin Ticket Assignment / Load Balancing


bchynds:
Another feature that we use in Spiceworks (although through a plugin) is a round-robin ticket assignment.  We have a small team of Helpdesk employees and tickets are load balanced with a round-robin assignment.  Agents can be added / removed from the round-robin as needed, but the round robin ensures that no tickets are sitting as open for extended periods and holds employees accountable.  Any way this can be achieved with Zammad?  I see SLA options, but I'm not seeing options for assignment.

SaschaRIO:
Hello,

We are looking for the same option since I cannot have one guy distributing the emails to the agents.

Any ideas?

vishnus:
Waiting for this feature :)

thorsteneckel:
Locking conversation. There is no new input. +1 and requesting comments create noise and distraction. We are working hard on the next upcoming/most requested features. If you want to state your interest in the feature and give valuable input please give the initial post a 👍 reaction and we are able to evaluate it. We will lock all distracting issues from now on. Thanks for your understanding.

PR #1574, opened 2017-10-23
Auto assign tickets to agents (similar to #854)

vishnus:
Hi there, 

A brief of use case as explained in the Zammad user group before:
When you have several dozen agents and thousands of tickets coming in, manual ticket claims become very impractical. There may be confusion on what tickets to pick; multiple agents may try to pick the same ticket (there's no locking mechanism); if there are economic incentives, tickets may get hoarded.

The distribution and resolution of tickets will become erratic and uneven.

With automatic assignment, the burden of assignment is not on the agent, but on the system. To start off with, a simple random assignment of tickets to agents within a group will ensure even distribution.

We have an observer for ticket model which executes auto assignment logic after_create. 

- Tickets object now have an additional flag called auto_assign which is set default to False. (Migrate file is created in this PR)
- If ticket is manually created, the auto_assign can be given as true/false in front end and if you leave the owner as blank it will take care of assigning a random agent belonging to that group.
- Logic of selecting the agent is currently random item in the list.
- If ticket is coming through a channel (email), a custom filter is written to enable ticket auto assign. Ex filter: If email to is "support@company.com" set auto_assign as True. The ticket observer is on after_create, so it will get executed once ticket is created:
image

This is just a starting point, but we'd very much love to have this feature as we have a very large support team. Please review this PR and let us know what you think.

Thanks


thorsteneckel:
Hi @vishnus - Thank you for your inspiring pull request. I really like the idea of a smart dispatching algorithm to assign tickets to agents based on their workload and skillset. This will surely become a feature in Zammad some day.
However, the proposed random assign functionality is not the sustainable solution that we are looking for. If you are interested in contributing the described assignment logic I'm more than happy to help.
Feel free to create an issue and discuss the requirements and implementation details for this one.

vishnus:
Thanks @thorsteneckel , I've opened up an issue #1581 to discuss further on this.

Regards
Vishnu

PR #1621, opened 2017-11-07
Add possibility to rescan LDAP/AD Users

ToubieObie:
It would verry nice if you add a "rescan" function that the Organisations become automatic assigned based on the E-Mail Adress.

I know that this only getting happen, if a users will be created.


monotek:
I can't see any changes here. 

Please create an issue for a feature requests not a pull request.

Issue #1674, opened 2017-11-23
OTRS Import after LDAP import capitalization problem


smalchow:
### Infos:

* Used Zammad version: zammad-2.1.0
* Used Zammad installation source: rpm package
* Operating system: CentOS 7
* Browser + version: 

### Expected behavior:

Users from an OTRS import should be merged with a previous LDAP import.

### Actual behavior:

The users seem to be partly updated, but not all, specifically it looks as if the LDAP import converts everything to lowercase, where as the OTRS import keeps the login as is.

We can see that users with userPrincipalName such as "Givenname.Familyname@mail.com" are created, whereas "givenname.familyname@mail.com" is updated. The OTRS user in the first however is still created with a lowercase login. 

Also i can see that in the first case there is a 3 digit number appended to the login, such as

OTRS: Givenname.Familyname@mail.com -> zammad: givenname.familyname@mail.com123

This behaviour is the same for all users.

### Steps to reproduce the behavior:

- Use one directory with a few users like Givenname.Familyname@mail.com and some like givenname.familyname@mail.com
- import LDAP users first (in our case with changed order of "samaccountname" and "userprincipalname" in "lib/ldap/user.rb:61"
- then import the tickets and users from OTRS as described in the documentation.
- check the logfiles, the users with Givenname.Familyname@mail.com are like this

_thread#-: add User.find_by(login: Ggggg.Fffff@mail.com)_

the users with all lowercase are like this

_thread#-: update User.find_by(login: ggggg.fffff@mail.com)_ 

when i search for the first user in zammad using

_User.find_by(email: 'Ggggg.Fffff@mail.com')_

i do not find the user, if i search with lowercase

_User.find_by(email: 'ggggg.fffff@mail.com')_

i find him, but the login is like this

login: "ggggg.fffff@mail.com507"

smalchow:
Maybe i solved the problem, (at least for me)

I changed 2 files in lib/import/otrs

user.rb
customer_user.rb

in each of the files I added this

        _mapped[:login].downcase!_

below line 70. (stricly not neccessary in the user.rb i think, but i am not a developer)

This should not be a problem with ldap backed installations, with other backends this could mean that the user has to login as "donald duck" instead of "Donald Duck" if i see this correctly.

so for now this is more of a dirty fix, but maybe someone with more development skills can make this better ?

thorsteneckel:
[...]

PR #1677, opened 2017-11-24
#1674 make sure login data is downcased. included tests to check as well

muhammadn:
A simple fix but a crucial one, kudos to @smalchow with the fix in #1674 

I have added the rspec tests to make sure this does not happen again.


muhammadn:
Question. Should i merge the code for 'Capital Email' with the Login? so to reduce code and code smell?

thorsteneckel:
Hi @muhammadn - thanks for the pull request 💪 No, it's file. Having one test for each case has a higher priority than duplicate code in tests. Surely we can reduce duplicate code and make writing tests easier by providing some helpers but these tests are fine by now. Thanks for keeping your eyes open and asking 👀 

I'll review your PR after the next stable release this week.

In general:
We follow a mixture of (our) best practices and the guidelines described in these resources:

http://www.betterspecs.org/
https://github.com/reachlocal/rspec-style-guide

PR #1717, opened 2017-12-15
Stable

infohk:
Is there any way to open new ticket form after for example a new incoming mail or a phonecall or any other remote case? I mean  automatic opening create ticket window with no user mouse or keyboard interaction.


thorsteneckel:
Use community.zammad.org for questions.

Issue #1687, opened 2017-12-03
Can't bind websocket to ipv6 ip [::]


monotek:
### Infos:

* Used Zammad version: 2.1
* Used Zammad installation source: source
* Operating system: ubuntu
* Browser + version: chrome 62

### Expected behavior:

* websocket can be started with ip "[::]" which is the ipv6 equivalent to "0.0.0.0".

### Actual behavior:

*  websocket process fails with:

```
zammad@2afbead7b431:~$ bundle exec script/websocket-server.rb -b [::] -p 6042 start
bundler: failed to load command: script/websocket-server.rb (script/websocket-server.rb)
RuntimeError: no acceptor (port is in use or requires root privileges)
  /usr/local/bundle/gems/eventmachine-1.2.5/lib/eventmachine.rb:530:in `start_tcp_server'
  /usr/local/bundle/gems/eventmachine-1.2.5/lib/eventmachine.rb:530:in `start_server'
  /usr/local/bundle/gems/em-websocket-0.5.1/lib/em-websocket/websocket.rb:46:in `run'
  /usr/local/bundle/gems/em-websocket-0.5.1/lib/em-websocket/websocket.rb:39:in `block in start'
  /usr/local/bundle/gems/eventmachine-1.2.5/lib/eventmachine.rb:171:in `run'
  /usr/local/bundle/gems/em-websocket-0.5.1/lib/em-websocket/websocket.rb:35:in `start'
  script/websocket-server.rb:123:in `block in <top (required)>'
  /usr/local/bundle/gems/eventmachine-1.2.5/lib/eventmachine.rb:194:in `run_machine'
  /usr/local/bundle/gems/eventmachine-1.2.5/lib/eventmachine.rb:194:in `run'
  script/websocket-server.rb:122:in `<top (required)>'
```

### Steps to reproduce the behavior:

* start websocket server with: 
  * bundle exec script/websocket-server.rb -b [::] -p 6042 start

Railsserver can be run flawlessly with:
* bundle exec rails server puma -b [::] -p 3000 -e production

monotek:
I've already tried. Same error.
It's in context of Kubernetes & the docker-compose container.

Imho it should work same way as in puma, where [::] is used. See:

https://github.com/zammad/zammad-docker-compose/blob/c6485593340aa9ab47a15c80be703394fdc30290/containers/zammad/docker-entrypoint.sh#L106

muhammadn:
[...]

PR #1722, opened 2017-12-19
fix IPv6 binding problem for eventmachine websockets

muhammadn:
Problem is that eventmachine does not accept IPv6 address with square brackets, (eg. [::1]). Only without it (eg ::1)

This PR checks if there is an IPv6 address with square brackets in the binding port flag and removes the square brackets.

### DIFF
--- a/script/websocket-server.rb
+++ b/script/websocket-server.rb
@@ -120,9 +120,8 @@ end
 @clients = {}
 Rails.configuration.interface = 'websocket'
 EventMachine.run do
- 
-  available_addr = Socket.ip_address_list.collect { |address| address.ip_address } # fetch all available IPs on machine
-  available_addr += ["0.0.0.0", "::"] # add IP binding for all interfaces
+  available_addr = Socket.ip_address_list.collect(&:ip_address) # fetch all available IPs on machine
+  available_addr += ['0.0.0.0', '::'] # add IP binding for all interfaces
   @options[:b] = IPAddr.new @options[:b] # check if it is valid IPv4 or IPv6 address
   raise Exceptions::UnprocessableEntity, 'IP address is not available on this machine.' if available_addr.exclude? @options[:b].to_s # raise if IP Address is not available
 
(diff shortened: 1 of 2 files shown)


monotek:
As i mentioned earlier in https://github.com/zammad/zammad/issues/1687#issuecomment-352793649 it also does not work without the square brackets.

script/websocket-server.rb
```diff
@@ -120,7 +120,11 @@ def after_fork(dir)
 @clients = {}
 Rails.configuration.interface = 'websocket'
 EventMachine.run do
-  EventMachine::WebSocket.start( host: @options[:b], port: @options[:p], secure: @options[:s], tls_options: tls_options ) do |ws|
+
+  if /(\[(?:[:0-9A-Fa-f]+)\])/.match?(@options[:b]) # check if there is an IPv6 Address
```
monotek: Checking ipv6 adresses by own regex looks wrong to me.
We could use resolv here: http://ruby-doc.org/stdlib-2.4.1/libdoc/resolv/rdoc/Resolv/IPv6.html
Example: https://stackoverflow.com/questions/3634998/how-do-i-check-whether-a-value-in-a-string-is-an-ip-address/15157862#15157862

muhammadn:
I have updated the code to use the standard built-in ruby class that parses and removes any square brackets and also shortens the code resulting in less errors than a own-rolled regex:

https://docs.ruby-lang.org/en/2.0.0/IPAddr.html

On my Mac:
Invalid IP -
```
─$ bundle exec script/websocket-server.rb -b '[::not:ipv6:addres]' -p 6389 start
bundler: failed to load command: script/websocket-server.rb (script/websocket-server.rb)
IPAddr::InvalidAddressError: invalid address
```

Valid IP:
```
╰─$ bundle exec script/websocket-server.rb -b '[::]' -p 6389 start                                                                1 ↵
2017-12-24T16:29:46Z:client(-) check unused idle connections...
```

When depending on the IPAddr class, it will also check for valid/invalid IPv4 addresses as well.

Have a merry christmas and happy new year. 🎅🎄

thorsteneckel:
@rlue - I think we can close this one now?

rlue:
Yes, the changes presented in this PR were refactored and merged separately in cd4dc31.

However, test coverage is currently excluded from the Travis CI build (see 2662c17), since it currently does not support IPv6.

Issue #1695, opened 2017-12-05
De-activating of all Administrator accounts should be prevented


thorsteneckel:
### Infos:

* Used Zammad version: 2.2 
* Used Zammad installation source: git
* Operating system: any
* Browser + version: any

### Expected behavior:

* Zammad should prevent disabling of all Administrator accounts as it does for deleting such accounts.

### Actual behavior:

* Zammad allows disabling of all Administrator accounts. See #1694. 

### Steps to reproduce the behavior:

* Use the Rails console to disable the Administrator account of your System via `User.find(*id*).update!(active: false)`

Yes I'm sure this is a bug and no feature request or a general question.

muhammadn:
@thorsteneckel I did some work but it for disallowing disabling all admin accounts. Which i think is wrong. 

Currently, zammad checks to have at least one admin user check if activerecord tries to disable all accounts. 

But this works only on record update and not create. 

So is your use case is for any admins that is deactivated or the last one?

thorsteneckel:
The use case is for the last one. So at some point there is only one active Admin user left. Then an update on that user is performed with something like `User.find(*id*).update!(active: false)` and the last admin gets deactivated too. So there is any Admin left to configure the system or re-activate any of the Admins. Hope that makes it clearer. Feel free to ask any questions.

muhammadn:
@thorsteneckel I have tried doing something similar: `User.find(3).update!(active: false)` on the only admin user left.

the message that came out is:
Exceptions::UnprocessableEntity: Minimum one user needs to have admin permissions.
	from app/models/user.rb:1007:in `last_admin_check_by_attribute'
	from (irb):45

and the last admin check seems to work.

thorsteneckel:
Ok great 👍 Then this was fixed in the meantime. Can be closed then.

PR #1726, opened 2017-12-22
#1695 prevent disable of admin users

muhammadn:
Currently, the admin user can be deactivated by rails console (or LDAP imports) which can be by accident. This fix ensures that admin user can never be deactivated by rails console or LDAP imports by checking on the record update.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -40,7 +40,7 @@ class User < ApplicationModel
 
   before_validation :check_name, :check_email, :check_login, :check_mail_delivery_failed, :ensure_uniq_email, :ensure_password, :ensure_roles, :ensure_identifier
   before_create   :check_preferences_default, :validate_ooo, :domain_based_assignment, :set_locale
-  before_update   :check_preferences_default, :validate_ooo, :reset_login_failed, :validate_agent_limit_by_attributes, :last_admin_check_by_attribute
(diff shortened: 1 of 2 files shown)


martini:
Hi @muhammadn Thanks for contributing! :)

I'm not sure about the use case. But so far as I can see, you can't disable an user with admin permission anymore, right?

In the past we checked if min. one user has admin permissions or at least one role has admin permission with valid users and do not allow to set the last user with admin permission to invalid.

Tests are here https://github.com/zammad/zammad/blob/develop/test/unit/user_test.rb#L932 

The code is here:
https://github.com/zammad/zammad/blob/develop/app/models/user.rb#L1003
https://github.com/zammad/zammad/blob/develop/app/models/user.rb#L1011

https://github.com/zammad/zammad/blob/develop/app/models/role.rb#L171
https://github.com/zammad/zammad/blob/develop/app/models/role.rb#L179

This did not work in your case?

Greetings,

 -Martin

muhammadn:
@martini yes. you cannot disable an admin user with these changes. I am not sure if that is what you wanted? I should have asked earlier.

Thanks for pointing out for me to use minitests. So is the use case right?

thorsteneckel:
Hi @muhammadn - I'll close this for now. Should be fixed with: https://github.com/zammad/zammad/commit/2c39eb47dec79441fba86e35bd844217cf98c066

Issue #1867, opened 2018-03-09
Two minor user interface/documentation issues


lpar:
- There are settings for date and time format under Translations, but I can't work out what the format specification is and it appears to be undocumented. (It's not standard Ruby/Rails.)

 - The Description help text for the Organizations management section appears to me in German.

thorsteneckel:
Hi @lpar - could you please move this over to the http://community.zammad.org/ since it's a technical question. Would be great if you could provide some more context (maybe screenshots?) because currently I don't fully get it.
Regarding the german text: This is a deeper issue and was (party) addressed in the linked PR above.

I'm closing this one here now.

Issue #812, opened 2017-02-28
Descriptions Languages


Toucan-Sam:
### Infos:
* Used Zammad version: 1.3.0
* Operating system: CentOS 7
* Browser + version: All

When clicking on "Description" at the top of a section, the language should match the base install. There are still a few sections that have German (fixed) descriptions. There is a similar issue with #556 but only covers SLAs (which looks finished). Remaining areas to translate are:

Organizations
Overviews
Calendars

em-ex:
BTW, as of 2.0.0 descriptions for Organizations, Overviews and Calendars are still shown in German, despite what language is set in browser.

thorsteneckel:
#556 is a duplicate / related issue to this.

The cause is that there is currently no performant way of translating these texts without sending all translations to the client despite only one is needed. This would (currently) slow down the initial startup time and the application in general because it will hold all translations in memory.
We have to find a proper solution for this but this might take some time 😖

Issue #556, opened 2016-12-19
SLA description only available in english


thorsteneckel:
As discussed in #551 with @abendrob we found out that the SLA description is only available in german. Maybe @rkaldung can help us out here?

Source:  https://github.com/zammad/zammad/blob/7df285d9dfc9299afc5b276db4a27b20ef395f87/app/assets/javascripts/app/models/sla.coffee#L21

pradorocchi:
@description = '''
** Service Level Agreements **, abbreviated ** SLAs **, help you to meet certain customers' time-related responses. Thus, for example, you can say customers should always get a response from you after 8 hours at the latest. In the event of an imminent violation or a breach, Zammad will alert you to such events.

It can be ** response time ** (time between the creation of a ticket and the first reaction of an agent), ** update time ** (time between a customer's request and an agent's reaction) and ** solution time ** (time between creation and closing a ticket ) To be defined.

Any threatening violations or violations are displayed in a separate view in the overviews. You can also configure ** e-mail notifications **.
'''

thorsteneckel:
Daaamn! Thats great. I'll add this asap.

thorsteneckel:
As discussed with @martini we changed the text to english. Currently there is no satisfying solution for translating long formatted strings like these. We will implement a proper solution for this.

pradorocchi:
Good morning, 
yesterday I worked on the Brazilian Portuguese translation and already completed 51% of it (I expect to complete 100% until this weekend) and while working on it I found other 3 German text-explanations around the system. I'm sending them for you here translated in English.

I did some interpretation to fit better in english because German sometimes is too "dry". And I checked the explanation by following it and navigating to the system to verify if the explanation is valid for the present system version (i checked with the developer version here). 
Feel free to read and adapt it better if needed.

The translation follows (sorry for not sending it via pull request)


______________________________
Manage - > Organization -> Description:

Mit Organisationen können Sie Kunden gruppieren. Dies hat u. a. zwei bedeutende Vorteile.
1. Als Agent haben Sie nicht nur die Übersicht über die Tickets eines Kunden sondern zusätzlich die Übersicht über die gesamte Organisation. Z. B. über die Suchen nach der Organisation, diese per einfachen klick öffnen.
2. Als Kunde können Sie die Tickets ihrer Kollegen mit einsehen und bearbeiten (sofern die Organisation eine "teilende" ist, dies können Sie je Organisation als Parameter einstellen).


[...]

PR #1885, opened 2018-03-20
Organisations: fix description

mbeijen:
It was auf Deutsch. Fixes second bullet of #1867.

schermafbeelding 2018-03-20 om 20 42 55

### DIFF
--- a/app/assets/javascripts/app/models/organization.coffee
+++ b/app/assets/javascripts/app/models/organization.coffee
@@ -17,10 +17,10 @@ class App.Organization extends App.Model
   ]
 
   @description = '''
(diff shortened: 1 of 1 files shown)


mbeijen:
- It sucks a bit that the Travis CI tests are broken
- also descriptions for other objects, such as overviews, are in German
- They're also not i18n'able!

thorsteneckel:
Hi @mbeijen so great to hear from you! Even greater to see some contributions to Zammad 🤓 

I totally agree with you on the sucking part regarding non i18n'ability for the object descriptions. This is addressed in the issues #812 and #556 . We are currently working on other tasks and lack of a better solution. I'll merge your PR since we decided to go with english over german as long as we have no proper solution.

Greetings to the Netherlands!

PR #1899, opened 2018-03-24
Fix Travis CI build failures (by switching to containerized CI environment)

rlue:
@thorsteneckel, here is the PR we discussed.

Original (virtualized) CI environment led to a test failure that I could not reproduce in development. Using the containerized CI environment instead (now the default on Travis) [fixed the error on my fork](https://travis-ci.org/rlue/zammad/jobs/357753511). ([See here for an overview of Travis Build Environments](https://docs.travis-ci.com/user/reference/overview/).)

An added benefit is that containerized Travis builds can be [reproduced on a local development machine (via Docker) for debugging/troubleshooting](https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image).

### Notes

1. Travis’ containerized environments have [fewer resources than virtualized environments](https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System), so this change leads to 34% longer build time (~25:00 to ~33:30).

   As such, it might be worth exploring ways to reduce overall build time (_e.g.,_ by running tests in DB transactions rather than dropping, migrating, and re-seeding the database multiple times per build). 

2. I moved all packages from the `apt-get install` line to the `addons.apt.packages` section. Most packages are installed in the Docker image already (with the exception of postfix), but including them here ensures that the CI build will always be working against the most recent versions. Is that important, or can some of these package install directives be removed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 dist: trusty
-sudo: required
-group: deprecated
+sudo: false
 notifications:
   email:
     - me+tv@zammad.com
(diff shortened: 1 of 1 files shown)


rlue:
Thanks @monotek!

So [the Travis build for this PR still failed](https://travis-ci.org/zammad/zammad/builds/357808386), but it looks like it was still run in the old virtualization environment rather than the containerized one that it was supposed to (see [Travis docs for how to tell the difference](https://docs.travis-ci.com/user/reference/overview/#For-a-finished-build)).

Not sure what went wrong. I'm gonna experiment on my fork in the morning to see if I can figure it out.

monotek:
i've reverted it.

Issue #1770, opened 2018-01-23
Owner reset after group selection


tobibe:
### Infos:
* Used Zammad version: 2.2.x
* Installation method (source, package, ..): package
* Operating system: ubuntu 16.4.2
* Database + version: MySQL 5.6.17
* Elasticsearch version:
* Browser + version: Chrome + Firefox

### Expected behavior:
* Change group of ticket doesn't change selected owner

### Actual behavior:
* if you change the owner and afterwards the group of ticket, the owner switches back

### Steps to reproduce the behavior:
* change owner of ticket without submitting the changes
* changes group of ticket
* now the owner of ticket is resetted to state of previous owner

![reset_owner](https://user-images.githubusercontent.com/9305993/35276309-513e8824-0043-11e8-9d13-d0fdfc3b73f6.gif)

Greetings tobibe

Yes I'm sure this is a bug and no feature request or a general question.

thorsteneckel:
Fixed with the merged PR. Thanks @mantas !

PR #1926, opened 2018-04-05
Fixes issue #1770 - Owner reset after group selection

mantas:


mantas:
Hi @martini 

I had added the browser test at https://github.com/mantas/zammad/blob/8924cc01ebacfbff0e7a362ad8f5b07851df46d2/test/browser/agent_ticket_update5_test.rb. It's changing the selected group and then checking if the user hadn't reset.

Should it test anything else?

PR #1948, opened 2018-04-12
Addition to zammad_ssl.conf

MrGeneration:
Added commented-out Proxy-Pass-settings that make the options work on openSuse and maybe other systems too. My apache 2.4 did not proxy without " "retry=1 acque=3000 timeout=600 keepalive=On" " and just threw 404. Might need testing if this is new in general. I know Ubuntu systems that work the same way.


thorsteneckel:
Thanks Marcel! @monotek will have a look at it 👍

PR #1963, opened 2018-04-20
Implement STARTTLS for IMAP

### LINKED ISSUE #769: Imap + STARTTLS
### Infos:

* Used Zammad version: 1.3.0 
* Operating system: Docker image on Ubuntu

### Expected behavior:

* Just tried to connect zammad to my mailserver but starttls seems not supported for IMAP connection.

### Actual behavior:

For IMAP connection we can only choose between SSL:
- no (port 143)
- yes (port 993)
and nothing about starttls on port 143 for example ?

adding an option to the ssl selectbox would be nice like SSL: no / yes / starttls
or just start tls automatically if supported by the server like with smtp

Thanks for this great software!

martin31821:
This PR adds support for STARTTLS when using IMAP incoming channels.

Fixes #769.

Implementation details:

Removes the additional check against port 143 in the web frontend.
Renames the fieldname to SSL/STARTTLS in the web frontent.

If the backend receives `ssl=true` and `port != 993` it will attempt to use starttls in favor of SSL.


martini:
@martin31821 you already got a job? If not, let me know. 👍🏻

martin31821:
@martini I'm self-employed in hard- and software development, focused on embedded linux and web projects.

Issue #1564, opened 2017-10-19
sorting of selects is not always working as expected


Kartoiser:
(also ß)
### Infos:

* Used Zammad version: 2.1.x
* Used Zammad installation source: package
* Operating system: Ubuntu 16.04.3 LTS
* Browser + version: IE12 (Edge)

### Expected behavior:

1. sorting should work like expected --> ascending numbers, then ascending letters (case insensitive)
2. umlauts (in addition to "ß") should be transformed into either their corresponding letter combination (ä --> ae / Ü --> UE) or transformed in replacing letter without dots (ä --> a / Ü --> U)

### Actual behavior:

1. when being in "Object" view and opening an already saved select opbject sometimes there are strange effects like 0 behind 9 or completely destroyed order:
![unbenannt](https://user-images.githubusercontent.com/26098704/31784338-9b184a94-b501-11e7-9e51-aca339b46d65.jpg)
2. when looking at the represented custom field in "Ticket" view umlauts get behind "Z":
![unbenannt](https://user-images.githubusercontent.com/26098704/31784459-04bbd2f4-b502-11e7-9f24-53c4f8ef3c00.jpg)

thorsteneckel:
@SummerStorm - please rebase with latest `develop`, close the PR (rejected), run it through the internal CI env and let me know if it's green so I can do QA 👍Thanks!

thorsteneckel:
This introduces new errors:
![screen shot 2018-08-23 at 14 36 20](https://user-images.githubusercontent.com/3873515/44527188-69a65f00-a6e6-11e8-9405-35755efebd14.png)

thorsteneckel:
This is actually a [new issue](https://github.com/zammad/zammad/issues/2209). Closing now. Sorry for the noise.

PR #1968, opened 2018-04-21
Fixed issue #1564 - sorting of selects is not always working

SummerStorm:
This fixes issue #1564.


SummerStorm:
Imported to internal branch for CI.

Issue #1259, opened 2017-07-16
Required field is not required on updating tickets


martini:
### Infos:

* Used Zammad version: 1.5
* Used Zammad installation source: 
* Operating system: any
* Browser + version: any

### Steps to reproduce
* Create a required ticket field for ticket zoom
* After creating a ticket, the required field is marked as required but actually not required

q16marvin:
any news to this?

jjw244:
Which field are you referring to @martini?

I'm running Zammad version 1.5.0 locally and have no issue with required fields as long as information is put into the field.  Can you tell me what field is "required" but is not supposed to be?

q16marvin:
hi, also in new 2.0 version this bug is there. do you know when it will be fixed?

HappyTepid:
I didn't realise my issue (#1599) was a dupe of this one, apologies. I only realised this was an issue with date fields as my other required field is a dropdown and when you make those required, the blank value at the top of the list disappears, forcing the user to select *something*.

q16marvin:
any news to this? i still cant use zammad because this "issue" :(

q16marvin:
I'm so sorry to annoy you, but is there any chance to get it fixed? this bug is now open so long and i love zammad, but i cant still use it... hope there is any way

thorsteneckel:
Dear @q16marvin ,

first of all, many thanks for your participation on the Zammad open source project.

Not usual as with proprietary software, we finance Zammad through service. So it's less money we have at our disposal to push Zammad.

If you would like to support Zammad, you can do so via Zammad Support: https://zammad.com/pricing#selfhosted

For all issues it is generally important that we tag an issue as soon as it is included in a release. This means that manual enquiry only minimizes the available resources but not the speed of implementation.

Thank you for your understanding.

thorsteneckel:
This issue is still open. Can be reproduced (with date, text etc. fields). This is caused by a missing check for required attribute over here:

https://github.com/zammad/zammad/blob/6b2e43baefdd199d8b781e9ab8c94aaa0b848194/app/assets/javascripts/app/models/_application_model.coffee#L121-L171

SummerStorm:
Fixed by 444e48e3772eb4abf8fb00170dca3d08ad65955f.

Issue #1984, opened 2018-04-28
The active flag of attributes is ignored when using the API 


SummerStorm:
### Infos:

* Used Zammad version: 2.4
* Installation method (source, package, ..): source
* Operating system: Ubuntu 16.04
* Database + version: PostgreSQL 9.5.12
* Elasticsearch version: 6.6.0
* Browser + version: Firefox ESR 52.5.2 (64-bit)

### Expected behavior:

When creating or updating a ticket using the API, parameters referring to an inactive attribute should either be silently ignored or flagged as an error.

### Actual behavior:

The `active` flag of attributes is ignored when updating or creating tickets using the API. 

### Steps to reproduce the behavior:

1. Create a new date attribute object named "required_date" and mark it as "inactive". Submit and update the database. (This bug applies to all attribute types; I'm just using the date type as an example.)

2.  `POST` the following JSON to `/api/v1/tickets` to create a new ticket:
<pre><code>{
	"title": "Test ticket",
	"group": "Users",
	"article": {
		"subject": "some subject",
		"body": "some message"
	},
	"customer": "nicole.braun@zammad.org",
	<b>"required_date":"2018-04-29"</b>
}
</pre></code>

3. The created ticket will have its `required_date` attribute set even though the attribute is inactive:
<pre><code>{
	"id": 109,
	"group_id": 1,
	"priority_id": 2,
	"state_id": 1,
	"organization_id": 1,
	"number": "25106",
	"title": "Test ticket",
	"owner_id": 1,
	"customer_id": 2,
	"note": null,
	"first_response_at": null,
	"first_response_escalation_at": null,
	"first_response_in_min": null,
	"first_response_diff_in_min": null,
	"close_at": null,
	"close_escalation_at": null,
	"close_in_min": null,
	"close_diff_in_min": null,
	"update_escalation_at": null,
	"update_in_min": null,
	"update_diff_in_min": null,
	"last_contact_at": null,
	"last_contact_agent_at": null,
	"last_contact_customer_at": null,
	"last_owner_update_at": null,
	"create_article_type_id": 10,
	"create_article_sender_id": 1,
	"article_count": 1,
	"escalation_at": null,
	"pending_time": null,
	"type": null,
	"time_unit": null,
	"preferences": {},
	"updated_by_id": 3,
	"created_by_id": 3,
	"created_at": "2018-04-28T08:43:09.514Z",
	"updated_at": "2018-04-28T08:43:52.222Z",
[...]

Issue #1598, opened 2017-10-31
"Allow past" setting is not respected


HappyTepid:
### Infos:

* Used Zammad version: 2.1
* Used Zammad installation source: Zammad.org Ubuntu instructions
* Operating system: 16.04.3 LTS
* Browser + version: Chrome latest

### Expected behavior:

* It should not be possible to open a ticket if a date field attribute with "Allow past" set to yes has a date prior to today entered.

### Actual behavior:

* Customer can choose a past date and submit a ticket successfully. 

### Steps to reproduce the behavior:

* Go to Admin -> Objects -> Ticket -> New Attribute
* Select "Date" as Format, set "Allow Past" to No
* Make the attribute required for customer
* Update DB
* Login as customer and open a ticket with a date prior to today in the new attribute field

SummerStorm:
Fixed via 444e48e.

PR #1985, opened 2018-04-28
Fixed Issue #1259, #1598, and #1984 by adding attribute validations

SummerStorm:
This fixes three separate issues:

#1259 is fixed by adding backend checks for required attributes during ticket creation/updates.

#1598 is fixed by parsing and validating any incoming date and datetime attributes against their `past` and `future` flags.

#1984 is fixed by dropping any parameters associated with any inactive attributes.

Apologies about bundling three issues in a single pull request but these issues are kind of interconnected so it's kind of hard to fix one without also fixing the others. For example, fixing #1259 requires that all supplied attributes are validated, and properly validating a date/datetime attribute is effectively fixing #1598.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -8,6 +8,7 @@ class TicketsController < ApplicationController
 
   prepend_before_action :authentication_check
   before_action :follow_up_possible_check, only: :update
+  before_action :validate_attributes, only: %i[create update]
 
   # GET /api/v1/tickets
   def index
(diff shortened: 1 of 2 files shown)


thorsteneckel:
@SummerStorm  - please migrate to an internal branch, rebase, close this PR, make sure the CI is green and let me know. Thanks!

SummerStorm:
@thorsteneckel Migrated and rebased to internal branch private-bz_1259_1598_1984_add_ticket_attribute_validations and CI is green now.

PR #1992, opened 2018-05-05
🌟 Resolve #1426: add timestamp to full quote

### LINKED ISSUE #1426: Email - full quote time specification
### Infos:

* Used Zammad version: 2.0.x
* Used Zammad installation source: source
* Operating system: CentOS 7
* Browser + version: Firefox 55.0.3

### Expected behavior:
Time when creating the article about the quote, as in OTRS 
for example: 

16.09.2017 13:06 - Firstname Lastname wrote:
>Quote
* 

### Actual behavior:
Only the last article is cited. No time is displayed above the quote to create the article.
* 

Ticket > Settings > Email - full quote > yes
Reply to an item in the ticket
*

dsukhin:
This PR resolves #1426 by using the already available info in the **article** variable/object to automatically add a timestamp and sender header to the full quotes of an email.  The next step would be localization but since the text is easily editable before sending, this is not the most critical (but should be looked at separately).

This helps bring Zammad replies even closer to looking like they came from a regular email client.

Let me know if you have any questions!

(PS, there are lots of commits here because i had to merge back to dev, was originally developing off stable. There is really only one file being changed in a very small place app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee (see diff))


thorsteneckel:
Hi @dsukhin ! Thanks a lot for your contribution. Can you please check/improve the following things:

- Something is wrong with the commit history. See the commit list above (over 250 commits 😨 ).
- The text is not translated/-able and will only be available in english. It should be possible to translate it.
- The date logic should be extracted into a separate function. This one-liner will be really hard to maintain.
- Please provide a test.

Feel free to ask any questions!

dsukhin:
Thanks. Changes are squashed to a single commit now, only changing one file. Just did some basic formatting of the date function. Should be easier to read now.

dsukhin:
Hey @thorsteneckel, Anything I can do to help here? Looks like the only two outstanding questions were a test case (for the added line) and translation options. 
* Is there already a test case for full-quoting email that I can add to? I don't see one at this commit: https://github.com/zammad/zammad/commit/bd061fb5c1d4c63f064dccfe0d92016c1c3ba1c2
* I know zammad has a generic translation framework, how does that work? Is it easy to add to this?
  * It's also important to note that the text inserted here is fully editable/removable before sending

dsukhin:
Just a heads up: added a small commit here that fixed a typo... instead of grabbing the create time of the article, I accidently was grabbing the create time of the user who created the article - in many cases this was one and the same, but obviously wanted to fix it!

SummerStorm:
Hi @dsukhin, thank you for the pull request. I used your pull request as the basis for 35a2e991a636102eaa9824eade048de9f77a71eb and 2656ee965c283745649977e9aace18abd097f731 and it's been merged into `develop`.

dsukhin:
Excellent. Glad this was helpful!

Issue #2016, opened 2018-05-14
Scheduler will stop after removing own ticket attribute but overview with condition which is including attribute git dropped


martini:
### Infos:

* Used Zammad version: 2.4
* Installation method (source, package, ..): any
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: any

### Expected behavior:

* Do not allow to drop ticket attributes via object manager where the ticket attributes is still used in trigger, overview or scheduler

### Actual behavior:

* Dropping ticket attributes leads to stopping scheduler:

```
E, [2018-05-14T10:02:52.758814 #6222] ERROR -- : PG::UndefinedColumn: ERROR:  column tickets.aaa1 does not exist
LINE 1: ...ed_at FROM "tickets" WHERE (group_id IN (1)) AND (tickets.st...
                                                             ^
: SELECT  DISTINCT "tickets"."id", "tickets"."updated_at", tickets.created_at FROM "tickets" WHERE (group_id IN (1)) AND (tickets.aaa1 IN ('Value#1')) ORDER BY tickets.created_at DESC LIMIT $1 (ActiveRecord::StatementInvalid)
```

### Steps to reproduce the behavior:

* create ticket attribute aaa1
* create overview where ticket attribute aaa1 is used in condition
* drop ticket attribute aaa1

martini:
@muhammadn did you review the pr #2033 ?

muhammadn:
@martini done with review, you can QA it now since the build is ok.

PR #2033, opened 2018-05-19
Fixed issue #2016 - Scheduler will stop after removing own ticket attribute but overview with condition which is including attribute git dropped

SummerStorm:
This fixes issue #2016 by adding frontend and backend checks to ensure that ticket attributes with outstanding object references (trigger, overview, or scheduler) cannot be deleted.

Ticket attributes with outstanding object references will have its delete button grayed out and disabled. The mouseover text of the disabled delete button will list all the objects that reference the attribute, basically hinting to the user that all dependent objects need to be deleted first in order to delete the attribute. 

asdaasdasdsd

### DIFF
--- a/app/assets/javascripts/app/controllers/object_manager.coffee
+++ b/app/assets/javascripts/app/controllers/object_manager.coffee
@@ -143,12 +143,17 @@ class Items extends App.ControllerSubContent
     e.preventDefault()
     id   = $(e.target).closest('tr').data('id')
     item = App.ObjectManagerAttribute.find(id)
+    ui = @
     @ajax(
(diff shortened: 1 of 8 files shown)


martini:
Hi @SummerStorm thanks for your pull request. We needed to do a little bit improvements/fixing small issue (e. g. it was not working if two objects as an field with same name or if a trigger/job/overview has more as one attribute for condition). But basically we merged it now into develop. Thanks for your help!

https://github.com/zammad/zammad/commit/2ec4c965de5a8d0e153ea03f4c261d281bd3fb59

Issue #2020, opened 2018-05-16
Incorrect reset after deleting single object-selection


KoJie:
### Infos:

* Used Zammad version: 2.4 (SaaS)
* Browser + version: Google Chrome Version 66.0.3359.139

### Expected behavior:

* When deleting a selection option (format: select), the selection of existing tickets should be set to "zero".

### Actual behavior:

* The selection of existing tickets categorized with the selection to be deleted is set to any other selection after deletion. 

### Steps to reproduce the behavior:

- Create an object with format: Select (e.g.: Language: English, German, Spanish)
- Now define "Spanish" for a ticket
- Delete the selection "Spanish" in the object
- In the tickets "German" is now defined as language

tkovs:
This only affects the tickets? This behaviour can be dangerous in the Zammad while a big system. I think that is important to check the scope of this behaviour. Someone can explains me the current situation?

SummerStorm:
@tkovs This is caused by the lack of a default "null" option for select attributes. If you have a select attributes with two options, English and German, the drop down menu will automatically choose English even if you don't click, and Zammad will mistakenly think you have chosen English. 

I've written a fix for it and I'll discuss with the team on what changes I need to make before merging it. 

In the meanwhile an easy workaround would be to put a "None" option at the beginning of each select attributes, e.g. None, English, German.

tkovs:
The point is that should not be possible to remove used elements from a list of objects, but the Zammad's team knows what is better.

PR #2041, opened 2018-05-24
Fixed issue #2020 - Incorrect reset after deleting single object-selection

SummerStorm:
This fixes issue #2020 by removing recently-deleted select attribute options and replacing them with defaults during ObjectManager migrations. There are two possible scenarios:

1. If the new options have a default, then all the recently-deleted options are migrated to this default.

2. If the new options does not have a default specified, then all recently-deleted options are migrated to `nil`, which is shown as the "-" empty choice in the UI. For this reason, all select attributes without a default specified will now have its `:nulloption` flag set to `true`.

### DIFF
--- a/app/models/object_manager/attribute.rb
+++ b/app/models/object_manager/attribute.rb
@@ -799,8 +799,11 @@ to send no browser reload event, pass false
       if !data_option.key?(:maxlength)
         data_option[:maxlength] = 255
       end
-      if !data_option.key?(:nulloption) && data_option[:default].empty?
+      if !data_option.key?(:nulloption)
         data_option[:nulloption] = true
+      end
+      if data_type == 'select' && data_option[:default].present?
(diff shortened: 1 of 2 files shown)


SummerStorm:
The single failed test from `rake test:units` was: `SessionEnhancedTest#test_check_clients_and_send_messages [/home/travis/build/zammad/zammad/test/unit/session_enhanced_test.rb:150]`

`rake test:units` completes fine on my dev machine so I think this is just a sporadic failure with the session tests.

thorsteneckel:
@SummerStorm - please migrate to an internal branch, rebase, close this PR, make sure the CI is green and let me know. Thanks!

SummerStorm:
This PR removes all instances of the removed option from the database. After some discussion we decided to fix the issue by a different approach: old options are left in the database intact, but can no longer be chosen.

Issue #2046, opened 2018-05-30
Umlauts in overview not working


Newan:
### Infos:

* Used Zammad version: 2.4
* Installation method (source, package, ..): Package .deb
* Operating system: Ubuntu 16.04
* Database + version: PostgreSQL
* Elasticsearch version: m/a
* Browser + version: Chrome

### Expected behavior:

* Umlauts from the organisationsname in the overview 'group by headline' must be translated

### Actual behavior:

* Umlauts from the organisationsname in  the overview 'group by headline' are showing as htmlcode
![image](https://user-images.githubusercontent.com/3693094/40720895-75cfbb98-6418-11e8-89e4-86d2b111878e.png)

### Steps to reproduce the behavior:

* Make an Organistaionname with an & charakter and group by it

Yes I'm sure this is a bug and no feature request or a general question.

PR #2049, opened 2018-06-02
Fixed issue #2046 - Special characters get HTML encoded when displayed in overviews

SummerStorm:
This fixes issue #2046. The actual fix is only two lines long: 
```
else if attributeConfig.tag is 'select'
      isHtmlEscape = true
```
Basically marking the target of `select` types as already HTML escaped so that it does not get unnecessarily HTML escaped again.


martini:
Hi @SummerStorm 

unfortunately select options can contain html content and this need to be html quoted. In this case the group by name is already quoted and also translated (because we need also to sort them bevor using it in the frontend/template). So we can take this 1:1 in the view (template file).

See here https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/controllers/_application_controller_table.coffee#L476 and https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/views/generic/table_row_group_by.jst.eco#L2

For full commit see https://github.com/zammad/zammad/commit/82acddc4200a4d7ccb8be1a8489485c71c3f2811

Thanks,

 -Martin

PR #2072, opened 2018-06-13
Update zammad_ssl.conf

Impyus:
Proposed change to the NGINX SSL Conf template to correct duplicate server block and correct syntax errors.

### DIFF
--- a/contrib/nginx/zammad_ssl.conf
+++ b/contrib/nginx/zammad_ssl.conf
@@ -1,143 +1,104 @@
 #
-# this is an example nginx config for zammad with free letsencrypt.org ssl certificates
+# this is an example nginx config for using SSL with zammad
+# this can be adjusted to be used with self-signed, trusted ca, and letsencrypt certs
 # replace all occurrences of example.com with your domain
-# create letsencrypt certificate by: certbot certonly --webroot -w /var/www/html -d www.example.com
+# If using LetsEncrypt create letsencrypt certificate by: certbot certonly --webroot -w /var/www/html -d www.example.com
 # create dhparam.pem by: openssl dhparam -out /etc/nginx/ssl/dhparam.pem 4096
-# download x3 certificate by: wget -q https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem -P /etc/nginx/ssl
+# download LetsEncrypt x3 certificate by: wget -q https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem -P /etc/nginx/ssl
 # you can test your ssl configuration @ https://www.ssllabs.com/ssltest/analyze.html
 #
 
-upstream zammad {
-  server localhost:3000;
+upstream zammad-railsserver {
+    server localhost:3000;
 }
 
 upstream zammad-websocket {
-  server localhost:6042;
+    server localhost:6042;
 }
 
 server {
-  listen 80;
-  listen [::]:80;
+    listen 80;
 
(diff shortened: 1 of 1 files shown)


thorsteneckel:
Thanks for your proposal! That looks much better now. Can you please change the indentation because it's very hard to reason what the actual changes were if each line is marked as "changed".

Impyus:
Sure, can you clarify what you mean by change the indentation?

thorsteneckel:
Sure. Let's take [this line](https://github.com/zammad/zammad/pull/2072/files#diff-33ddb85c4e705cd0b5993e64efa5e3bcR12) for example. The only difference is that you added two spaces in front (indentation). But the general functionality stays the same. The diff shows that it was changed but actually wasn't. Just added whitespaces => indentation. So please remove those so we can see what actually changed. Thanks!

Impyus:
Ah yes I understand you now, I think I've sorted that.

contrib/nginx/zammad_ssl.conf
```diff
@@ -17,35 +18,33 @@ upstream zammad-websocket {
...
-  ssl_certificate /etc/nginx/ssl/example.com-fullchain.pem;
-  ssl_certificate_key /etc/nginx/ssl/example.com-privkey.pem;
+  ssl_certificate /etc/nginx/ssl/your_cert.crt;
+  ssl_certificate_key /etc/nginx/ssl/your_private_key.key;
 
-  ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+  ssl_protocols TLSv1.2;
 
-  ssl_ciphers HIGH:!aNULL:!MD5;
+  ssl_ciphers 'EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH';
```
monotek, line 47: Did you checked that these ciphers are available on all supported platforms?

contrib/nginx/zammad_ssl.conf
```diff
@@ -57,84 +56,46 @@ server {
   ssl_stapling on;
   ssl_stapling_verify on;
 
-  ssl_trusted_certificate /etc/nginx/ssl/lets-encrypt-x3-cross-signed.pem;
+  ssl_trusted_certificate /etc/nginx/ssl/your_cert.crt;
```
monotek, line 59: This conflicts with the command in line 7.

thorsteneckel:
@monotek - what do you think?

thorsteneckel:
@monotek - ping

monotek:
My remarks above still seem to apply...

contrib/nginx/zammad_ssl.conf
Impyus: Fair point, but perhaps the guidance @ line 7 should be changed to reflect. I renamed it so it was obvious that this was referring to your certificate whether you were using LetsEncrypt or not and would need to be changed to reflect it. I'm happy to add a readme to provide clarification on editing each section of the file for the end user.

contrib/nginx/zammad_ssl.conf
Impyus: As far as I understand it, these ciphers are supported on current releases of openSSL and by Apache, Nginx and Lighttpd. The reasoning for change here is that TLS 1.0 and 1.1 are deprecated. TLS 1.2 is the standard and this set-up enables the most secure config of the webserver. Again alternatives can be offered in the readme I'll create if users wish to enable weak ciphers for some reason. As it currently stands this config achieves an A+ rating on SSL Labs.

Issue #845, opened 2017-03-16
Notification E-Mail Template


ThePatzen:
Hi!

Is there a way to change texts or to change the e-mail template for ticket-notifications?

### Infos:
* Used Zammad version: 1.3.x
* Operating system: Ubuntu 16.04
* Browser + version: Chrome 56

martini:
Hi @ThePatzen 

currently you can change/set trigger notifications via admin ui.

Notifications for agents (for a new ticket or escalations) you currently can not change it via UI. 

This templates are located unter app/views/mailer/ticket_create/$LANG.html.erb at the moment.

 -Martin

ThePatzen:
Hi @martini 

thanks for your response!

Would be great if Agent-Notifications could be manipulated the same way but for now this helps ^^

Best Regards!
David

Newan:
Is this update save. all my templates are overridden with 2.6

It is possible to create  update save templates like custom.de.html.erb, with an higher prio?

Trapulo:
I also really need to update notification template via standard GUI...

thanks

RareshLuma:
+1

MrGeneration:
Locking conversation as there hasn't been any new input apart from `+1`.
If you want to discuss this, please use our community for this and reference the issue. :-)

PR #2081, opened 2018-06-15
Added custom notification templates to make changes update save

Newan:
@martini  explain on #845, that user can change the notification templates in the files, but this are not update save. With 2.6 all templates are overridden. this small change is for custom templates for your environment

### DIFF
--- a/lib/notification_factory.rb
+++ b/lib/notification_factory.rb
@@ -40,7 +40,7 @@ returns
 
     # custom but as fallback, use 2 char locale
     if !File.exist?(location)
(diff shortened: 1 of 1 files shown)


rlue:
Thanks @Newan. We liked your changes and were eager to incorporate them into a larger refactoring of the `NotificationFactory` class in 902300e. I hope you're not discouraged by the fact that we wrote our own take here — this class was due for a refactoring, and it needed some extra test coverage anyway. Please continue to hack on Zammad and share your changes with us as you see fit!

Please note that the format of custom template files is now `<locale>.<format>.erb.custom` (rather than `custom.<locale>.<format>.erb`). Thanks again for your contribution!

PR #2243, opened 2018-09-19
Check the correct hash array if the agent is present

Jaykju:
This is a bug that I encountered in Zammad 2.6.x when an agent is also a customer. An agent does not show up if it was found as a customer beforehand. 
Changing this to the correct array fixes this very unlikely but annoying bug.

### DIFF
--- a/app/controllers/time_accountings_controller.rb
+++ b/app/controllers/time_accountings_controller.rb
@@ -46,7 +46,7 @@ class TimeAccountingsController < ApplicationController
           end
         end
       end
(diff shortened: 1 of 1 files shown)


thorsteneckel:
Hi @rlue - please create a RSpec request spec for this as a verification 👍

rlue:
Hi @Jaykju, we added your changes separately via 97b17ca so that we could incorporate a regression test around this behavior. Thanks so much for your contribution!

PR #2299, opened 2018-10-19
different roles to set up different "ticket templates"

Ziegenmann:
When you are building a new role is this role not available for define which ticket attributes the agent can see and which not. you are only able to set the access for the attributes to "ticket.agent". Is there any way you can probably define more than one "role" or is that in planning? we need to set up different ticket templates for two or more groups.  

Furthermore may there is any way to load up different "templates" when you are choosing a group? A salesman has another requirements to a ticket than a technican. And the customer dont has to see 100 different attributes in it.


thorsteneckel:
Wow there went something really wrong. Please use the [community board](http://community.zammad.org/) for asking questions.

PR #2301, opened 2018-10-20
2278 - Change real name to display name in email account add

### LINKED ISSUE #2278: Change "Real Name" to "Display Name"
### Infos:

* Used Zammad version: 2.6.x
* Installation method (source, package, ..): n/a
* Operating system: n/a
* Database + version: n/a
* Elasticsearch version: n/a
* Browser + version: n/a

### Enhancement

* I suggest renaming "Real Name" to "Display Name", as display name is what usually is meant for Names in a from of an E-Mail. Real Name is likely to irritate users - we could solve that quite easy by changing this string.

* Add a E-Mail-Account
  (Adding Aliases is optional, but has the same effect)
* Edit the E-Mail-Address, you'll find the string "Real Name"

image
image

yasirazgar:
Fixes https://github.com/zammad/zammad/issues/2278

Changed name from Real name to Display name in UI.

### DIFF
--- a/app/assets/javascripts/app/models/email_address.coffee
+++ b/app/assets/javascripts/app/models/email_address.coffee
@@ -12,7 +12,7 @@ class App.EmailAddress extends App.Model
     )
 
   @configure_attributes = [
-    { name: 'realname',   display: 'Real name',  tag: 'input', type: 'text', limit: 250, null: false },
+    { name: 'realname',   display: 'Display name',  tag: 'input', type: 'text', limit: 250, null: false },
     { name: 'email',      display: 'Email',     tag: 'input', type: 'email', limit: 250, null: false },
     { name: 'channel_id', display: 'Channel',   tag: 'select', multiple: false, null: true, relation: 'Channel', nulloption: true, filter: @filterChannel, do_not_log: true },
     { name: 'note',       display: 'Note',      tag: 'textarea', note: 'Notes are visible to agents only, never to customers.', limit: 250, null: true },


thorsteneckel:
Hi @rlue - could you please pick up this PR + issue and check the internal CI env?

martini:
@rlue and please check if the wording is correct. At least in my mail client it's named "Full Name".

rlue:
@martini it's looking fine over here:

screenshot_20181024_225630

rlue:
@thorsteneckel It looks like there's one failing browser test being caused by a missing German translation:

screenshot_2018_10_24__15_37_22_605_match_failed-2917499226053644691

I don't think I'm authorized to add translations yet (or am I?). In either case, I'm not too familiar with the procedure. Maybe you can walk me through it today and I can add a page to the internal wiki for it, then add the translation and get this test passing?

rlue:
Sorry @martini, totally misunderstood your previous comment.

"Display name" is a good choice of wording, in my opinion. A quick web search [shows that it's in pretty common use](https://www.google.com/search?q=email+"display+name") (notably, among companies like [Microsoft](https://docs.microsoft.com/en-us/office/client-developer/outlook/mapi/smtp-addresses) and some well-known universities), and IMO it's better than "real name" or "full name", because it hints at how this information is going to be used.

I don't know if this term is specified in an RFC anywhere, but I stand behind it.

rlue:
So since we're changing the term in English, should we also change the term in the german translation? Maybe, "Anzeigename" instead of "richtiger Name"?

MrGeneration:
@rlue  sorry for the delay! "Anzeigename" would be the perfect translation for that :)

martini:
JFI: In develop its Anzeigename now.

Issue #2296, opened 2018-10-18
Ticket list in overview of customer and organisation random order


MrGeneration:
### Infos:

* Used Zammad version:  2.6.x
* Installation method (source, package, ..): n/a
* Operating system: n/a
* Database + version: n/a
* Elasticsearch version: n/a
* Browser + version: n/a

### Expected behavior:

* When viewing the list of open and closed tickets of customers and/or organisations, the list should have a chronological order.

### Actual behavior:

* When viewing the list of open and closed tickets of customers and/or organisations, the list has an absolute random order.

### Steps to reproduce the behavior:

* create some tickets for a customer
* take a look on the open/closed tickets within user information.

DEV-Information:
Original ordering should be: "created_at: :desc"
Parameters are missing: https://github.com/zammad/zammad/blob/develop/app/controllers/concerns/ticket_stats.rb#L7

![image](https://user-images.githubusercontent.com/6549061/47160738-dad24f00-d2f0-11e8-8abb-fd2b57ca942c.png)

![image](https://user-images.githubusercontent.com/6549061/47161587-9fd11b00-d2f2-11e8-82fe-3a424d5bfdef.png)

Yes I'm sure this is a bug and no feature request or a general question.

This came from Ticket#1029435 

Edit: Fixed image

PR #2300, opened 2018-10-19
2296 - Customer and org ticket to be sorted created_at desc

yasirazgar:
Fixes https://github.com/zammad/zammad/issues/2296
Previously it was sorted by updated_at desc, changed it to order by created_at desc.

PR #2315, opened 2018-10-26
FIX: Ticket list in customer and org view now sorted by 'created_at'

prssanna:
Fixes https://github.com/zammad/zammad/issues/2296
Continues the work done in https://github.com/zammad/zammad/pull/2300 and also adds two browser tests for it.
@thorsteneckel 

- [x] ticket stats are now ordered by `created_at`
- [x] added a request spec for testing if the tickets are returned by `/api/v1/ticket_stats` in the correct order
- [x] added a browser test in `agent_user_profile_test.rb`
- [x] added a browser test in `agent_organization_profile_test.rb`

### DIFF
--- a/app/controllers/concerns/ticket_stats.rb
+++ b/app/controllers/concerns/ticket_stats.rb
@@ -8,6 +8,8 @@ module TicketStats
       limit: limit,
       condition: condition,
       current_user: current_user,
+      sort_by: 'created_at',
+      order_by: 'desc',
     )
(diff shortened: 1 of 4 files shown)


prssanna:
It's strange that the build fails only for mysql everytime and not postgresql, it might have something to do with the usage of `sort_by: created_at`.
Any idea why this is happening?

thorsteneckel:
As described via mail this is a duplicate of #2300 which was already merged in the meantime.

PR #2335, opened 2018-11-07
Increased popup width for text modules or placeholder list

alex-LE:
We are heavily using the text templates and the names for them got longer than the current popup could show. As the textarea is much larger, i've increased the size.
It should be possible to increase it up to 902px, so i'm free for other values.

### DIFF
--- a/app/assets/javascripts/app/lib/base/jquery.textmodule.js
+++ b/app/assets/javascripts/app/lib/base/jquery.textmodule.js
@@ -27,7 +27,7 @@
     this.active     = false
     this.buffer     = ''
 
-    this._width = 380
+    this._width = 600
 
     // check if ce exists
     if ( $.data(element, 'plugin_ce') ) {


thorsteneckel:
Hi @mrflix - what do you think?

mrflix:
Hey Alex! Thanks for the contribution! Can you provide a screenshot to illustrate your cause?

alex-LE:
@mrflix sure

grafik

mrflix:
Hey Alex! Thank you for the screenshot. This helped me a lot to understand the problem you're facing. Quite interesting to see how long your text module names are and what kind of content you put in there. Do the agents reply in all languages? Would probably be cool to filter out text blocks based on the languages the agent can speak (write).

Regarding your fix: I realized that there are hidden problems with it. To illustrated what I mean lets exaggerate and put the width to 902px: 
screen shot 2018-11-19 at 14 31 23

Now if the agents screen is not as big we would get this:
screen shot 2018-11-19 at 14 31 09

So based on your idea I built this:

screen shot 2018-11-19 at 14 26 24

screen shot 2018-11-19 at 10 55 50

So the window doesn't have a fixed width but it rather uses as much space as it needs but not more than the maximum (the textarea width).

The fix is currently in internal review but you can already try it out on your system using this patch:
[0001-Textmodule-widget-dynamically-increase-width.patch.zip](https://github.com/zammad/zammad/files/2595506/0001-Textmodule-widget-dynamically-increase-width.patch.zip)

martini:
@alex-LE as @mrflix mentioned here is his commit, https://github.com/zammad/zammad/commit/f2644cd622c37f2a2cd97890e4db47511025825c `Credits to @alex-LE ` 

It's available in upcoming Zammad 2.8.

PR #2348, opened 2018-11-14
Update de.html.erb

Charburner:
Improved wording, removed the mix of formal and informal language.

### DIFF
--- a/app/views/mailer/password_reset/de.html.erb
+++ b/app/views/mailer/password_reset/de.html.erb
@@ -4,12 +4,12 @@ Zurücksetzen Deines #{config.product_name} Passworts
 <br>
 <div>wir haben eine Anfrage zum Zurücksetzen des Passworts für #{config.product_name} Account <b>#{user.login}</b> erhalten.</div>
 <br>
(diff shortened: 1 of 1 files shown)


thorsteneckel:
Hi @Charburner - thanks a lot for your contribution! Actually the formal language is wanted in this template because it might reach customers - which in some scenarios shouldn't receive informal texts. You can use [the `.custom` file suffix](https://zammad-admin-documentation.readthedocs.io/de/latest/manage-trigger.html?highlight=custom#other-notifications) if you need to change it in your scenario.

Charburner:
@thorsteneckel 

You might want to check the content again because it contains a clumsy mix of formal and informal language. That is the actual reason I wanted to change it. 

> Zurücksetzen Deines #{config.product_name} Passworts

> <div>Hallo #{user.firstname},</div>

> <div>Dein #{config.product_name} Team</div>

thorsteneckel:
Hi @Charburner - you're right! I wasn't aware of that. We discussed this and came to the conclusion that it should be formal (because it's customer facing). Would you mind updating the Pull Request to correct into formal text only? Otherwise we will take over which might take a bit.

Issue #2365, opened 2018-11-21
Sporadic 502 Bad gateways with correct working Zammad


MrGeneration:
### Infos:

* Used Zammad version: 2.7, develop and earlier
* Installation method (source, package, ..): n/a
* Operating system: n/a
* Database + version: n/a
* Elasticsearch version: n/a
* Browser + version: n/a

### Expected behavior:

* Every request going to nginx and thus Zammad will work with 200.

### Actual behavior:

* Sometimes (quite hard to reproduce and find), nginx and even apache will proxy to Zammad by using die IPv6 localhost (::1). Puma is not listening on IPv6 so the webserver throws a 502 Bad Gateway.

### Steps to reproduce the behavior:

* To forcefully reproduce this, either ensure localhost only resolves to ::1 or change localhost to ::1 in your webserver configuration.

Yes I'm sure this is a bug and no feature request or a general question.

PR #2404, opened 2018-12-13
Fix detection of local elasticsearch service

kevin-olbrich:
If elasticsearch is local and the URL was set via rake cli to "localhost" instead of "127.0.0.1", the service is not restarted.
This results in a broken state (search does not work, background tasks fail, etc.) if zammad is updated together with elasticsearch.

### DIFF
--- a/contrib/packager.io/functions
+++ b/contrib/packager.io/functions
@@ -248,7 +248,7 @@ function setup_elasticsearch () {
 
   yes | /usr/share/elasticsearch/bin/elasticsearch-plugin -s install ingest-attachment
 
-  if [ "${ES_CONNECTION}" == "http://127.0.0.1:9200" ]; then
+  if [ "${ES_CONNECTION}" == "http://127.0.0.1:9200" ] || [ "${ES_CONNECTION}" == "http://localhost:9200" ]; then
     ${INIT_CMD} restart elasticsearch
   else
     echo -e "\n It seems you're running an external Elasticsearch server on ${ES_CONNECTION}"


monotek:
Not sure when this should happen?
The script explicitly sets 127.0.0.1.
Why should you want to change that? 

Usage of localhost could lead to problems with ipv6. 
See: https://github.com/zammad/zammad/issues/2365

thorsteneckel:
Hi @monotek - I'm with you that we should note (somewhere) that the usage of `localhost` is ambiguous and should be discouraged. Maybe the admin documentation would be the proper place - @MrGeneration ? 
@monotek - do you thing an additional check for `localhost` would be good to notify the user?
What about other local variations like e.g. `::1` etc.?

monotek:
If we want to support "::1" we should get sure that ipv6 is supported everywhere. I gueess that would need some testing...

I'd also like to know why the setting was changed to "localhost" in the first place to becom aware how likely this is for others or if it was changed by accident.

kevin-olbrich:
I changed the URL in the past to use an external cluster. This lead to many problems I did not want to further inspect and changed back to localhost. I did not know that it's using 127.0.0.1 by default.

While I understand your comment, the logic checks if the service is local. Local can be 127.0.0.1 as well as "localhost".

> Usage of localhost could lead to problems with ipv6.
> See: #2365

Actually "localhost" should resolve to both IPv4 and IPv6 with IPv6 preferred by the client. There are many reasons a connection can fail and the client retries with IPv4.
This is perfectly fine and is supposed to work exactly like that.

The script above checks if elasticsearch is local (thats that the codes says) but it lacks support for all possible values. Actually the upgrade is not done correctly in my setup, just because I entered "localhost" instead of 127.0.0.1.

Also, hardcoding IPv4 is bad practice as it's a legacy protocol with better alternatives like IPv6.
As ES5 is listening on both IPv4 and IPv6 by default, this should only cause problems on systems that are configures wrong.

kevin-olbrich:
> I'd also like to know why the setting was changed to "localhost" in the first place to becom aware how likely this is for others or if it was changed by accident.

I tried to use an external cluster which was no good idea. I used the same command to switch back to my local installation. And as there was no info that "localhost" should not be used, I did not ask.

At least for a clean install of CentOS 7, IPv6 is absolutely no problem. It just causes problems when something is not set up right like broken APs with radvd. Suchs problems should always be fixed by the net admins and not by the software. Disabling whole IPv6 via sysctl would also cause no problems, as localhost will only be looked up as 127.0.0.1.

monotek:
OK, i think we can add this check.
I just wanted to get sure that localhost was manually set.

kevin-olbrich:
Thank you very much!

PR #2461, opened 2019-02-01
Set RAILS_SERVE_STATIC_FILES for systemd units

jangrewe:
All  requests to assets result in a 404 unless i set RAILS_SERVE_STATIC_FILES in zammad.env


monotek:
This is a development option. 
Use nginx or apache to serve static files.

See: 
* https://github.com/zammad/zammad/blob/develop/contrib/nginx/zammad.conf
* https://github.com/zammad/zammad/blob/develop/contrib/apache2/zammad.conf

jangrewe:
That's the fun part, i first tried it without Nginx and got 404s, so then i used the Nginx config you provide (with the document root pointing to <path>/public), and still got the 404s. In the log it mentioned that it couldn't find a matching route...
So i re-read the docs, found https://docs.zammad.org/en/latest/install-source.html#initialize-your-database and tried that env var out - and it worked.

monotek:
Just use the DEB or RPM packages. 
They are configured the right way by default.

PR #2466, opened 2019-02-05
Password hashing: upgrade to argon2 2.x (argon2id), fix non-unique salt bug (fixes #2462)

### LINKED ISSUE #2462: Zammad reuses password hash salts
### Infos:

* Used Zammad version: 2.8.x
* Installation method (source, package, ..): docker-compose
* Operating system: Debian 9
* Database + version: `PostgreSQL 9.6.11 on x86_64-pc-linux-musl, compiled by gcc (Alpine 6.4.0) 6.4.0, 64-bit`
* Elasticsearch version: 5.6.14
* Browser + version: Chromium 71

### Background

I noticed that the users in my DB that have been created by the email parser have:

a) password hashes of an empty string (which IMO they shouldn't, they should have `NULL` instead of a hash) and
b) lots of users have the _same_ password hash, which shouldn't happen due to salting:

```
zammad_production=# select count(*) from users group by password order by count(*) desc limit 5;
 count
-------
 10014
  4805
  1912
   704
     4
(5 Zeilen)
```

With the Rails console I was able to verify that the hashes are actually of the empty string:

```
zammad ~ # docker exec -e COLUMNS="`tput cols`" -e LINES="`tput lines`" -ti zammaddockercompose_zammad-railsserver_1 bash
root@7e0d3a1dba11:/opt/zammad# rails c
I, [2019-02-02T23:35:07.457931 #4191-47145821884240]  INFO -- : Setting.set('product_logo', "1a49a80da9c6e174c9537709642498d1.png")
Loading production environment (Rails 5.1.6.1)
irb(main):001:0> PasswordHash.verified?("$argon2i$v=19$m=65536,t=2,p=1$jZG4d36BZZQ/ya0r0weBZw$FztleWSa6YTq1MSbi0AmmQxf8qBllUszVXi8UgMlQvk", "")
=> true
irb(main):002:0>
```

The reused salts are probably caused by `lib/password_hash.rb` because it reuses the same instance of the Argon2 library, which the [documentation](https://github.com/technion/ruby-argon2#usage) says you shouldn't do:

> If you follow this pattern, it is important to create a new Argon2::Password every time you generate a hash, in order to ensure a unique salt.

This wouldn't be that big of a problem if it only affected the automatically created users who have no password anyway, but unfortunately it also affects manually created users with real passwords.

### Expected behavior:

* When creating multiple users with the same password, they should have different salts and therefore different password hashes.
* Automatically created users with empty passwords IMO shouldn't have any password hashes at all, but instead simply `NULL`. Storing password hashes for empty passwords just makes it harder to identify which hashes resemble proper passwords, and which don't.

### Actual behavior:

* Creating multiple users (e.g. "User A" and "User B") with the same password leads to users with identical password hashes:

```
zammad_production=# select firstname,lastname,password from users where firstname = 'User' and lastname IN ('A', 'B');
 firstname | lastname |                                             password
-----------+----------+--------------------------------------------------------------------------------------------------
 User      | A        | $argon2i$v=19$m=65536,t=2,p=1$VJK50mnb/QiTsGEb3fbb4w$zhTUC0H/m8J142EGaB/vva96U2YUs4s+uzrmkunBW+A
 User      | B        | $argon2i$v=19$m=65536,t=2,p=1$VJK50mnb/QiTsGEb3fbb4w$zhTUC0H/m8J142EGaB/vva96U2YUs4s+uzrmkunBW+A
(2 Zeilen)
```

* Automatically created users with empty passwords have password hashes.

* Manually create multiple users in Admin -> Users with the same passwords and compare their password hashes.
* Manually create multiple users without setting their passwords and see that they have password hashes which match `PasswordHash.verified?("hash...", "")`.

martinvonwittich:
Here's my attempt at solving #2462 :)

Please review this carefully, I'm new to Ruby, Rails, RSpec and the Zammad codebase :)

Especially the removal of  `password_empty?` and the therein contained logic might be controversial; if this change is not acceptable, I can revert it.

A note about argon2 >= 2.x: this upgrade breaks the old code in `lib/password_hash.rb`! This is because argon2 >= 2.x creates `argon2id` hashes by default, but `lib/password_hash.rb` can only identify `argon2i` hashes, and will therefore misinterpret an `argon2id` hash as a plaintext password, and hash it again, thereby breaking it. If you checkout this branch and then return to develop, make sure that you downgrade `argon2` back to 1.x before authenticating against Zammad!

Another note about the way that `user.rb` currently handles the password hashing: I'm not very fond about the way this happens automagically in the `before_validation` callback. The way it currently works it has two significant disadvantages:
- If the hashing algorithm creates a new kind of hash that `lib/password_hash.rb` cannot identify (e.g. as outlined above with `argon2id`), `user.rb` will misinterpret the hash as a plaintext password and cause it to be hashed again, thereby breaking it.
- Due to the fact that Zammad supports SHA256 hashes for legacy purposes, users can exploit it to set a password that breaks the password policy. E.g. to set the password `a`, do the following:
	```
	martin@martin ~ % echo -n 'a' | sha256sum | sed 's/^/{sha2}/'
	{sha2}ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb  -
	```
	
	Then change your Zammad password to `{sha2}ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb`. As this is a valid password hash, Zammad will not attempt to hash it, and will instead put it directly into the database, where it'll get upgraded to an `argon2id` hash on the next login. As Zammad has never seen the plaintext password while setting it, it cannot validate the password against the password policy.
	
	(Only SHA256 hashes can be abused this way, as the user cannot create a useable argon2 hash without knowledge of the Zammad secret. He'll be able to get it into the database, but he won't be able to login afterwards.)
	
I feel like separate methods to set plaintext passwords (subject to password policies and therefore allowed for users) and hashed passwords (not subject to password policies and therefore only allowed for admins) might be the cleaner solution here.

From the commit message:

- Gemfile.lock: upgraded to argon2 (2.0.1)
- app/models/user.rb:
  - ensure_password will now replace an empty string with nil, to ensure that
    unset passwords are stored as NULL in the database, and not as a hash of
    the empty string.
  - removed password_empty?, and thereby also the special logic that actually
    conserved the old password hash when a caller tried to set it to '' or to
    nil. This might very well be controversial, but I found the difference
    in behavior to be very confusing.

    This did what you would expect:

      x = User.create(login: "foobar", password: "", created_by_id:1, updated_by_id:1)

    It creates a user without a password. But this IMO did not do what you
    would expect:

      x = User.create(login: "foobar", password: "barfoo", created_by_id:1, updated_by_id:1)
      x.password = ''
      x.save

    I would have expected the `x.password = ''` to clear the password, but
    instead it was just silently ignored. Until now, it was actually impossible
    to remove a user password.
    As far as I can tell, the frontend will not submit a password key in its
    JSON request when creating or editing a user without typing anything in the
    password field, so this _probably_ shouldn't break anything.
- lib/password_hash.rb:
  - crypt will now no longer reuse Argon2 instances, and will therefore now use
    unique salts for hashing passwords.
  - legacy? will now consider argon2i password hashes as legacy hashes, so that
    they will be upgraded to argon2id hashes. This has the added benefit that
    it automatically gets rid of passwords with non-unique salts.
  - hashed_argon2? now just calls Argon2::Password::valid_hash? instead of
    using a copy-pasted regex. This might be vital when a future update of
    Argon2 introduces a new hash - the old code only matched specifically
    argon2i hashes. This had the very dangerous side effect that it would have
    attempted to re-hash argon2id hashes because they would'nt match the regex,
    thereby corrupting the hashes in the database.
  - added hashed_argon2i? which now uses the copy-pasted regex to identify
    legacy argon2i hashes.
  - removed the argon2 method as storing Argon2 instances was a bad idea
    anyways :)
- spec/models/user_spec.rb:
  - when set to plaintext password -> hashes password before saving to DB:
    fixed the test not to rely on identical hashes.
  - when changed to empty string -> sets password to nil (#2462)
    added this test to ensure that app/models/user.rb overwrites the current
    password with nil when it is changed to an empty string.
  - when changed to nil -> sets password to nil
    added this test to ensure that app/models/user.rb overwrites the current
    password with nil when it is changed to nil.
  - when a user is created with an empty string as password ->
      sets password to nil
    added this test to ensure that a user created with an empty string password
    stores nil instead of a hash of the empty string.
  - when a user is created with nil as password -> sets password to nil
    added this test to ensure that a user created with a nil password
    stores nil. This was never actually broken, but I thought having it
    formally defined in a spec wouldn't hurt :)
  - when creating two users with the same password (#2462) ->
      does not generate the same password hash
    added this test to ensure that two users with the same password won't have
    the same password hash.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -89,7 +89,7 @@ GEM
     addressable (2.5.2)
       public_suffix (>= 2.0.2, < 4.0)
     arel (8.0.0)
-    argon2 (1.1.4)
+    argon2 (2.0.1)
       ffi (~> 1.9)
       ffi-compiler (~> 0.1)
     ast (2.4.0)
@@ -380,7 +380,7 @@ GEM
       thor (>= 0.18.1, < 2.0)
     rainbow (3.0.0)
     raindrops (0.19.0)
-    rake (12.3.1)
+    rake (12.3.2)
     rb-fsevent (0.10.2)
     rb-inotify (0.9.10)
       ffi (>= 0.5.0, < 2)
@@ -616,4 +616,4 @@ RUBY VERSION
    ruby 2.4.4p296
 
 BUNDLED WITH
-   1.16.2
+   1.17.3
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1227,24 +1227,20 @@ raise 'Minimum one user need to have admin permissions'
   end
 
   def ensure_password
-    return true if password_empty?
-    return true if PasswordHash.crypted?(password)
+    # don't accidentally hash the empty string
+    # see https://github.com/zammad/zammad/issues/2462
+    if password == ''
+      self.password = nil
+    end
 
-    self.password = PasswordHash.crypt(password)
-    true
-  end
-
-  def password_empty?
-    # set old password again if not given
-    return if password.present?
-
-    # skip if it's not desired to set a password (yet)
+    # don't attempt to hash nil
     return true if !password
 
-    # get current record
-    return if !id
+    # don't re-hash an already hashed passsword
+    return true if PasswordHash.crypted?(password)
 
-    self.password = password_was
+    # hash a plaintext password
+    self.password = PasswordHash.crypt(password)
     true
   end
 
--- a/lib/password_hash.rb
+++ b/lib/password_hash.rb
@@ -37,8 +37,9 @@ module PasswordHash
     # (#2462), we'll use this opportunity to replace the old argon2i hashes
     # (which might have duplicate salts) with new argon2id hashes.
     return true if hashed_argon2i?(pw_hash) &&
-        Argon2::Password.verify_password(password, pw_hash, secret)
-    return false
+                   Argon2::Password.verify_password(password, pw_hash, secret)
+
+    false
   end
 
   def hashed_sha2?(pw_hash)
@@ -46,7 +47,7 @@ module PasswordHash
   end
 
   def hashed_argon2?(pw_hash)
-    return Argon2::Password::valid_hash?(pw_hash)
+    Argon2::Password.valid_hash?(pw_hash)
   end
 
   def hashed_argon2i?(pw_hash)
--- a/lib/password_hash.rb
+++ b/lib/password_hash.rb
@@ -6,6 +6,10 @@ module PasswordHash
   extend self
 
   def crypt(password)
+    # https://github.com/zammad/zammad/issues/2462
+    # Do not reuse Argon2::Password instances, the salt is only randomized on
+    # instantiation!
+    argon2 = Argon2::Password.new(secret: secret)
     argon2.create(password)
   end
 
@@ -27,7 +31,14 @@ module PasswordHash
     return false if pw_hash.blank?
     return false if !password
 
-    sha2?(pw_hash, password)
+    return true if sha2?(pw_hash, password)
+    # Since argon2 >= 2.0.0, we can use the argon2id hash instead of the
+    # argon2i hash. As older Zammad versions accidentally reused the same salt
+    # (#2462), we'll use this opportunity to replace the old argon2i hashes
+    # (which might have duplicate salts) with new argon2id hashes.
+    return true if hashed_argon2i?(pw_hash) &&
+        Argon2::Password.verify_password(password, pw_hash, secret)
+    return false
   end
 
   def hashed_sha2?(pw_hash)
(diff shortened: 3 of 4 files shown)


MrGeneration:
Hey Billy, could you take a look on this whenever you find some spare time? :)

SummerStorm:
Ok, got it. I've also re-assigned #2462 to myself, since it's related.

martinvonwittich:
Whoops, there was a `puts(user.inspect)` left that wasn't supposed to be there. Fixed.

The CI apparently fails because I didn't run `rubocop`; I'll get to that later.

thorsteneckel:
Hey Martin! Thanks so much for your contribution. I had to change some minor things to restore a behavior (can't set empty password for existing users that previously had one). However, the most of the work is still yours. I'll close this PR now.

PR #2529, opened 2019-03-30
Show email address in user autocomplete (#2436)

### LINKED ISSUE #2436: auto completion for user selection dialogues are not consistent
### Infos:

* Used Zammad version: 2.8
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: any

### Expected behavior:

When handling customer data within Zammad, the agent/admin should always see the E-Mail-Address of the user, as you might not be able to find the correct user (without trying) if having more than one identical user.

**This affects**

* New Ticket dialogue
* Reply dialogue (via button)
* Reply dialogue (manual)
* Trigger and Scheduler

### Actual behavior:

There are two ways how the auto completion and general look of customer information looks like when selecting customers, replying and forwarding. This is not consequent and -especially on creating new tickets- might cause confusion and more work than it should be.

### Examples

This is how Customer and CC looks like when creating a new ticket:
image

This is what it looks like when pressing on "reply":
image

Manually entering E-Mail-Adresses within replying on a ticket is consequent (and how it should be):
image

Kinda optional but when we're working on consistent UI stuff, this should also apply schedulers and triggers:
image

## Wanted stuff

The e-mail address should appear behind the user name, the to field (especially when pressing on "reply" should contain the FROM-Name.

martinvonwittich:
As discussed [in the community](https://community.zammad.org/t/new-send-e-mail-ticket-show-customer-e-mail-address/1838) and in #2436 - when choosing a user/customer, Zammad currently only shows the user name, but not the email address:

image

It'll only show the user's email address after you've chosen a user from the list. This is annoying when you have multiple users with the same name, and you know which address you want to write to; you might have to repeatedly choose a user, check if it's the right one, and if not, choose another one.

This MR changes to behaviour of the user selection list to also show the email address:

image

It does _not_ completely fix #2436 though, as it only modifies the user selection list (e.g. "customer" when creating a new ticket, or "specific user" in triggers/schedulers/macros), but not the email address selection list (To and CC fields when writing emails).

### DIFF
--- a/app/assets/javascripts/app/views/generic/object_search/item_object.jst.eco
+++ b/app/assets/javascripts/app/views/generic/object_search/item_object.jst.eco
@@ -8,6 +8,9 @@
     <% else: %>
       <%= @object.displayName() %>
     <% end %>
+    <% if @object.email: %>
+      <span class="recipientList-email">- <%= @object.email %></span>
+    <% end %>
     <% if @object.organization: %>
       <span class="recipientList-detail">- <%= @object.organization.displayName() %></span>
     <% end %>
--- a/app/assets/stylesheets/zammad.scss
+++ b/app/assets/stylesheets/zammad.scss
@@ -6857,6 +6857,10 @@ footer {
     opacity: 1;
   }
 
+  .recipientList-email {
+    opacity: 0.5;
+  }
+
   .recipientList-detail {
     opacity: 0.5;
   }


MrGeneration:
Hey Billy,

could you please check this if you find some time for it? :)

Thanks!

thorsteneckel:
Hi @martinvonwittich - thanks for your PR 😍 I was talking to @martini about this and we share the concern that this change will make the UI permanently more noisy for a enhancement that you'll only need in certain cases. That's why we were thinking about introducing a tooltip so you can hover the element and see what the exact user entry is. However, we need @mrflix input on that. Any thoughts @mrflix ?

MrGeneration:
Personally I dislike the tooltip idea as I see this slowing me down in my workflow.
How about a middle way with a setting turning this on or off (globally on the system).

Reason is, that we partly already display the mail address and Martin simply made this completely straight.

mrflix:
I think that it makes sense that Zammad generally doesn't show the email when creating a new ticket but does show it when picking an email recipient. Zammad is not just about email but about all forms of communication.

But it's a very good idea to show additional data that helps to make a decision when the default data is not helping. In the screenshot example I'd show the email for "Martin von Wittich" next to the name. But not for "Cron Daemon" and "argon2 password" and "Martin von .wittich+3".

I agree with @MrGeneration that tooltips are not a good solution here. Since you move your mouse over the dropdown items you will very often trigger the tooltip all-tough you don't want to. Except if the delay is set very low. But then you'd get slowed down a lot.

martinvonwittich:
> Personally I dislike the tooltip idea as I see this slowing me down in my workflow.
> How about a middle way with a setting turning this on or off (globally on the system).

Yeah, I'd also prefer an option over the tooltip solution. A tooltip would mean that you actually have to hover with the mouse over each customer before you can see the email address, which would be rather cumbersome in situations where you actually have multiple customers that you need to differentiate by their email address. It also seems less accessible, because it won't show up when you're selecting a customer via keyboard.

> I think that it makes sense that Zammad generally doesn't show the email when creating a new ticket but does show it when picking an email recipient. Zammad is not just about email but about all forms of communication.

Hm, maybe it should be context-dependent? Creating or replying to email tickets might show the email address, creating phone tickets might show the phone number (which one though :/), ...
Not sure if there are other ticket types where other information would be helpful because I haven't used other channels yet ^^

When I'm about to create an email ticket, seeing the email address in the customer list might be useful before selecting the customer, _even_ in situations where there is only a single customer and there's no need to differentiate. E.g. a customer might call me via phone and ask for something to be sent over mail, and he gives me his email address as `john.doe@foobar.com`. I create a new email ticket to send it, and I type `john doe` into the customer list, and it suggests the following customer:

> John Doe - john.doe@acompletelydifferentdomain.com

I can see immediately that it's the wrong address (maybe even a completely different person with the same name), and so I can immediately create a new customer for `john.doe@foobar.com`. If creating email tickets would only show the email address for multiple users with the same name, I would have chosen the wrong user first, and only then I would have created a new user.

mrflix:
> Hm, maybe it should be context-dependent?

Yeah maybe thats the simplest solution to decide on what the additional info that gets displayed should be. When you pick "Send new email" it should use the email field as additional information to help pick the right one if multiple have the same name.

> John Doe - john.doe@acompletelydifferentdomain.com

Good point. Though when you search for "john doe" it will show both the incorrect user `John Doe - john.doe@acompletlydifferentdomain.com` and the correct user `John Doe - john.doe@foobar.com`. But you're right – you'd have a hard time to decide when the names don't match because of a typo or cause they are different.

So I wouldn't veto a global setting to always show the primary communication medium (email for email tickets, phone number for phone tickets, tweet handle for twitter messages etc.) as secondary information.

martinvonwittich:
Interesting sidenote: there already is a way to get Zammad to display the email addresses:

```
App.Config.set('ui_user_organization_selector_with_email', true);
```

Found that while I was becoming acquainted with the code :)

thorsteneckel:
Hi @martinvonwittich  - I'm a bit confused about the current state of this PR. Do I get it right that we should instead implement another, context aware solution instead?

martinvonwittich:
> Hi @martinvonwittich - I'm a bit confused about the current state of this PR. Do I get it right that we should instead implement another, context aware solution instead?

Sorry, the current state is "I'm too lazy to work on this". I should have closed this after I had found a way to permanently enabled, `ui_user_organization_selector_with_email`, but I forgot. I still think that always showing the address and having it context-sensitive would be a useful feature at some point (possibly as a visible option in the admin area), but it'll have to wait :)

PR #2531, opened 2019-03-31
allow agents to always reopen tickets, regardless of group.follow_up_possible

martinvonwittich:
As discussed in the community: https://community.zammad.org/t/allow-only-agent-to-re-open-the-ticket-follow-up-possible/1906

This commit changes the behavior of the option group.follow_up_possible
(admin -> manage -> groups -> <group> -> follow up possible). Before this
commit, setting this option to "do not reopen ticket but create a new ticket"
meant that _both_ agents and customers weren't allowed to reopen closed
tickets. This is problematic because it means that an agent who accidentally
closed a ticket has no way to undo this, which I believe wasn't the intention
of this option.

This commit changes the behavior so that agents are now exempt from this
limitation and can close and reopen tickets at whim.

As I said in the community thread, I see no reason to add a separate option for this new behavior, because I believe this is how it's supposed to work; preventing agents from reopening tickets just doesn't seem to be very useful to me. If this is not acceptable, I could of course still change that :)

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -659,6 +659,7 @@ class TicketsController < ApplicationController
   def follow_up_possible_check
     ticket = Ticket.find(params[:id])
 
+    return true if current_user.permissions?('ticket.agent') # agents can always reopen tickets, regardless of group configuration
     return true if ticket.group.follow_up_possible != 'new_ticket' # check if the setting for follow_up_possible is disabled
     return true if ticket.state.name != 'closed' # check if the ticket state is already closed
 
--- a/spec/requests/ticket_spec.rb
+++ b/spec/requests/ticket_spec.rb
@@ -1872,63 +1872,6 @@ RSpec.describe 'Ticket', type: :request do
 
     end
 
-    it 'does ticket with follow up possible set to new_ticket (06.01)' do
-      group = create(
-        :group,
-        follow_up_possible: 'new_ticket' # disable follow up possible
-      )
-
-      ticket = create(
-        :ticket,
-        title:       'ticket with wrong ticket id',
-        group_id:    group.id,
-        customer_id: customer_user.id,
-        state:       Ticket::State.lookup(name: 'closed'), # set the ticket to closed
-      )
-
-      state = Ticket::State.find_by(name: 'open') # try to open a ticket from a closed state
-
-      # customer
-      params = {
-        state_id: state.id, # set the state id
-      }
-
-      authenticated_as(customer_user)
-      put "/api/v1/tickets/#{ticket.id}", params: params, as: :json
-      expect(response).to have_http_status(422)
-      expect(json_response).to be_a_kind_of(Hash)
-      expect(json_response['error']).to eq('Cannot follow up on a closed ticket. Please create a new ticket.')
-
-      ticket = create(
-        :ticket,
-        title:       'ticket with wrong ticket id',
-        group_id:    group.id,
-        customer_id: customer_user.id,
-        state:       Ticket::State.lookup(name: 'closed'), # set the ticket to closed
-      )
-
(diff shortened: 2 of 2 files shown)


MrGeneration:
Hey Billy,

could you check this when you find some free time please? :)

Thanks!

thorsteneckel:
Hi @martinvonwittich - thanks a ton! Can someone (@MrGeneration / @martinvonwittich ) please create an issue for this? We need to have a reference for the CHANGELOG / Release news notes.

Other than that we need to cover the changes with tests. There is already [a request spec](https://github.com/zammad/zammad/blob/develop/spec/requests/ticket_spec.rb) for the controller but I haven't found any example for checking the follow up functionality. However, the spec is in a pretty bad shape so please take inspiration for creating a test from a proper spec like e.g. [this one](https://github.com/zammad/zammad/blob/develop/spec/requests/time_accounting_spec.rb). Let me know if you have any questions.

martinvonwittich:
Done: #2534

martinvonwittich:
> Other than that we need to cover the changes with tests. There is already a request spec for the controller but I haven't found any example for checking the follow up functionality.

Oh, there actually was a test for this, and my change broke it :D

```
  1) Ticket request handling does ticket with follow up possible set to new_ticket (06.01)
     Failure/Error: expect(response).to have_http_status(422)
       expected the response to have status code 422 but it was 200
     # ./spec/requests/ticket_spec.rb:1912:in `block (3 levels) in <top (required)>'
```

Unfortunately it had a lot of duplicated code and it also didn't actually test the `follow_up_possible = yes` behavior, so I decided to remove it completely and wrote a new one. It took me some hours to figure out a clean way to prevent code duplication; in the end I decided on a helper function so I don't have to repeat all those `expect` lines for each test. I hope that is acceptable in specs...?

I moved the new test into its own `describe '/api/v1/tickets'` section; I assume that the giant `describe 'request handling'` is one of the things that's bad style in the spec file?

Here's the new test: https://github.com/zammad/zammad/pull/2531/commits/4018d187f76edf568dea610f1598e319ea860b7a

Output:

```
zammad@martin ~ (agents-can-always-reopen-tickets) % bundle exec bin/rspec spec/requests/ticket_spec.rb -e '/api/v1/tickets'
Running via Spring preloader in process 9254
Run options: include {:full_description=>/\/api\/v1\/tickets/}

Ticket
  /api/v1/tickets
    follow_up_possible
      when follow_up_possible = yes
        allows admins to reopen tickets
        allows agents to reopen tickets
        allows customers to reopen tickets
      when follow_up_possible = new_ticket
        allows admins to reopen tickets
        allows agents to reopen tickets
        does not allow customers to reopen tickets

Finished in 4.13 seconds (files took 1.05 seconds to load)
6 examples, 0 failures
```

thorsteneckel:
Beautiful ❤️ the helper method is a step in the right direction we can easily migrate it over to a [RSpec shared example](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) like we use it in [other specs](https://github.com/zammad/zammad/blob/57b17f7b81676eb32185905edd918dfd6723fdb8/spec/models/object_manager/attribute/validation/future_past_spec.rb#L19-L41).

I'll talk to the team and let you know.

martinvonwittich:
Dammit, tried to rename the branch in my fork from `agents-can-always-reopen-tickets` to `2534-agents-can-always-reopen-tickets` now that I have a proper bug number, and that closed the merge request :/

thorsteneckel:
Hi @martinvonwittich - thanks again for creating a PR for this. Unfortunately we had to work on the spec which leads in our current workflow (GitHub PR -> review/CI on internal GitLab -> GitHub) to overwriting you as an author of the commit :/ I hope that's not too demotivating. We're still searching for a way to keep the original author of the code...

Issue #333, opened 2016-10-30
Error 500 when adding new Attributes


AlexR2106:
Hi,
I receive an Error 500 when adding new Attributes to Tickets, Customers , etc.:
{"error":"undefined method `name' for nil:NilClass"}

Database update ist not possible.
System was installed from source and is beeing serverd by Passenger / Apache2 running on Debian 8. Other functions are running fine.

Regards,
Alex

monotek:
Can't reproduce in current 1.2 develop.
Closing.

PR #2553, opened 2019-04-26
Fix user search input background-color

xf-:
Ensure the a background-color is set. In current version the font color is #333 form the body and no background. If you use an dark system theme the input can get the background-color fro your system near #333.
The user inpur is the only field without form-control (sets a color and background). Worked on my system.

### DIFF
--- a/app/assets/javascripts/app/views/generic/object_search/input.jst.eco
+++ b/app/assets/javascripts/app/views/generic/object_search/input.jst.eco
@@ -3,7 +3,7 @@
   <% if @attribute.multiple: %>
     <%- @tokens %>
   <% end %>
-  <input name="<%- @attribute.name %>_completion" class="user-select token-input js-objectSelect" autocapitalize="off" placeholder="<%- @Ti(@attribute.placeholder) %>" autocomplete="off" <%= @attribute.autofocus %> role="textbox" aria-autocomplete="list" value="<%= @name %>" aria-haspopup="true">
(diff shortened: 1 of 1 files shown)


thorsteneckel:
Hi @xf- - thanks for this PR! Sorry that you had to wait so long to get a response. Could you please provide 1-2 screenshots that show what this change does?

@mrflix - could you please have a look if this change is reasonable? Thanks!

xf-:
@thorsteneckel i will add some screenshots soon. In general - take the inspect tool and look at inputs. All have a font and a background color except the customer/user search. Depending on OS/Browser the system colors is used. If you use a dark skin and a dark font color isset the contrast is near 0

mrflix:
Hey Xaver!
The user selector is a tokenfield (can potentially hold several comma separated values instead of just one) just like the tags field. The styles of an input in a tokenfield get overwritten to look like blank space. Instead, the tokenfield element gets the style via `form-control`:
```html
<div class="tokenfield form-control u-positionOrigin">
```
In this case however the input field doesn't even get our basic input styles because of an error. It's not visible to us because the difference the white background color which is already white per default. However its visible to you because of the dark theme. `form-control` is one way to add the basic input styles, however an ugly one since we already use that class one level further up. A more elegant way is to add the attribute `type="text"` which is somehow missing from this input.

martini:
closed by accident - re-open

mgruner:
Hi @xf-. Thanks a lot for your suggestion. I tried this both in Firefox and Chrome with dark themes and cannot reproduce the issue you describe with the current develop branch. Maybe it has been fixed in the meantime?
I would be grateful if you could also try this and let me know your findings. In case it should still be an issue, please reopen this PR and give  detailed steps for reproduction. Thanks and best regards

xf-:
looks fixed to me - no issue anymore & no patch applied locally

mgruner:
Thanks! 👍

Issue #2540, opened 2019-04-08
LDAP agent group users also exist in LDAP customer groups; failure


pneumatic:
* Used Zammad version: 2.9.x (from the 'Version' tab, I can't get the zammad cmdline to produce a proper version number)
* Installation method (source, package, ..): DEB (from latest stable)
* Operating system: Ubuntu 18.04
* Database + version: psql (PostgreSQL) 10.7 (Ubuntu 10.7-0ubuntu0.18.04.1)
* Elasticsearch version: Version: 5.6.16, Build: 3a740d1/2019-03-13T15:33:36.565Z, JVM: 1.8.0_191
* Browser + version: Chrome latest

### Expected behavior:
Map bulk LDAP groups (e.g. "Staff All") to Zammad customer roles.  Then map help desk agents (who exist in the bulk staff groups) into Zammad roles.  Additionally, mutual exclusivity exists between the 'customer' and 'agent' roles.  In my particular environment where multiple departments will be leveraging Zammad for workflow, it is probable that Zammad 'agents' from one department may be 'customers' of another set of 'agents' (e.g. the IT agents may submit a request to the mail room or HR).

### Actual behavior:
If the agent mapping is done at the bottom of the LDAP config, the 'customer' role prevails and everyone is a customer.  If the agent mapping is done at the top of the LDAP config, then the users in the agent group fail upon sync.  I've tried to 'promote' individuals from the Zammad users once the bulk customer mapping has been performed.  While this is tedious with the number of agents that we have, I thought that it could be a workaround for now.  But the LDAP sync overwrites those promotions and everyone is back to customer.

### Steps to reproduce the behavior:
Simply map a bulk user group into customers and then attempt to promote some of those users to agents with either another LDAP mapping or Zammad "promotion".

Yes I'm sure this is a bug and no feature request or a general question.

pneumatic:
Additionally, once a particular LDAP group has failed, it will always fail.  In lieu of a known method for resetting the Zammad user database, I have been resorting to VM snapshots in troubleshooting this issue.  It would be nice to be able to have a purge function for the Zammad user db.

Kquinn:
I am not sure if the dev team needs to verify, but I have independently encountered, tested, and reproduced this in a lab environment.  Screenshots of the lab are attached in a zip archive.
[Zammad_LDAP.zip](https://github.com/zammad/zammad/files/3218688/Zammad_LDAP.zip)

[...]

Issue #967, opened 2017-04-19
Access to my own Tickets (where I'm customer of) in a Group im not Agent


maege:
### Infos:

* Used Zammad version: 1.4.x
* Used Zammad installation source: package
* Operating system: Debian
* Browser + version: Chrome

### Expected behavior:

* I see all my Tickets and also all Tickets from the group(s) i'm a member of.

### Actual behavior:

* If i have agent role to a single group (group A), and create a ticket (as "customer") to an other group (group B). I dont't see my own Tickets.
* I also don't have access to my old Tickets in a other Group than Group A.

### Steps to reproduce the behavior:

1. Give a user (User A) Agent role for a group (not your default incoming mail group)
2. Send a Mail to Zammad from this user. (The Ticket will be created in the default group) or open Ticket via Web.
3. The User A does not have access to his own ticket.

Thanks for your great work!

Edit by Zammad / @MrGeneration : 
Zammad should also close the tab, if creating a ticket into a group you have no access to. This refrains Zammad from showing error messages because of missing permissions.

hagen3000:
I agree. It would be nice if the agent could see his own tickets. Regardless of the group.

martini:
++

chrklo:
we need this feature too.

GregMighty:
This would be great for us as well :))))

chrklo:
any news on this topic?

I tried to get it working with the new role feature in 2.0 - but without success.

thorsteneckel:
Being a customer and agent at the same time is currently not supported. There are a lot of cases we have to cover which will take some time. We will give you an update as soon as we will work on them.

Wardrop:
Also eagerly awaiting the ability for an agent to also be a customer. For most internal help desk systems, this is requirement with no real workaround. We have power users responsible for providing support for certain products so therefore need to be agents, but these same users also need to be able to raise tickets of their own for other systems or for general IT support.

Charburner:
Hey guys,

since your last reply was about 6 months ago, would you mind to give us an update or tell us if this requested feature is anywhere near the top of your to do list? :-)

Have a nice day

NvL-WEAP:
We're eager to get this feature as well.

Kind regards

MrGeneration:
Please simpy upvote this on original post and don't comment here, as it creates a lot of noise to those who watch this. Locking conversation for now.

MrGeneration:
Just a small follow up:

[...]

PR #2592, opened 2019-05-29
Prevent assigning conflicting user roles with LDAP sync and nested groups

Kquinn:
PR to fix issue #2540

The issue:
Microsoft Active Directory LDAP implementations utilize nested groups for security and email distribution. As such, users are typically members of multiple groups. When these multiple groups are assigned to different Zammad roles which conflict, the import will simply fail for that user and continue on, resulting in unexpected outcomes in the synchronization.

What this patch does:
Add one method (resolve_role_conflicts) which can detect and resolve these error conditions by granting the user the highest role(s) that do not cause a conflict.  This means if a user is synchronized with groups assigning them both the Agent and Customer roles, the Customer role will be removed and the Agent role will remain. If a user is synchronized with groups assigning the Administrator, Agent, and Customer roles, the Customer role is removed and both Administrator and Agent remain.

### DIFF
--- a/lib/ldap/group.rb
+++ b/lib/ldap/group.rb
@@ -64,6 +64,40 @@ class Ldap
       groups
     end
 
+    # Ensures users in multiple groups do not get mapped to conflicting roles.
+    #
+    # @param user_roles_mapping [Array<Number>] The group DN to local role mapping.
+    #
+    # @example
+    #  user_roles_mapping = [1,2,3]
+    #  ldap_group.resolve_role_conflicts(user_roles_mapping)
+    #  #=> [2,1]
+    #
+    # @return [Array<Number>] The user DN to local role IDs mapping.
+    def resolve_role_conflicts(user_roles_mapping)
+
+      #Do not attempt to resolve conflicts for empty or single item arrays
+      return user_roles_mapping if user_roles_mapping.length <= 1
+
+      #Reverse sort to give preference to highest privlege roles.
+      result = user_roles_mapping.sort { |a, b| b <=> a }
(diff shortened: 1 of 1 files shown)


thorsteneckel:
Wow @Kquinn - thanks a ton for your contribution! I will have to do some tests with this but it looks really elegant and promising. There will be a release in the next week(s) and I won't be able to do so before that. However, reviewing open PRs is one of my top priorities after the the release.

Kquinn:
@thorsteneckel thank you for the feedback and expected timeline for review.

This code solves my use-case and prevents the issue mentioned in #2540, where no custom roles are created/used, but could still lead to unexpected results when custom roles are used and the values of least to most privilege do not correspond with the role IDs.

One possibility for making this universal would be to allow the user to define an "order" or "rank" that would be used to determine which role should supersede the other.

martini:
closed by accident - re-open

thorsteneckel:
Hey @Kquinn - it's been a while now. Please excuse the delay. We decided to resolve #967 so that there are no Role conflicts and Agents can be Customers at the same time (and vice versa). This means that there is now no Role conflict to resolve any longer and you can actually sync your LDAP Roles to Zammad without any change needed. This is possible since Zammad 3.5 (previous version, 3.6 is currently latest). Therefore I'll close this PR. Anyhow, thanks to your contribution.
You had to make some sacrifices regarding preference/priority but over all I really liked your code and approach. Let us know if you'd like to work full time on Zammads code base 😅 Anyhow, thanks again! Closing now.

PR #2637, opened 2019-06-30
Add Rails 5 log to STDOUT env variable support

thegcat:
The default production environment settings for Rails, I think starting with Rails 5, have an env variable with which one can send the logs to STDOUT instead of the standard Rails logs, see https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L93-L97

In our case this is useful to send the logs to systemd/journald through STDOUT, allowing us to just set `RAILS_LOG_TO_STDOUT=1` in the `/etc/zammad/zammad.env` instead of mucking in the Zammad code itself.

This PR adds the code part copied from the Rails template linked to above.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -54,6 +54,12 @@ Rails.application.configure do
   # Use a different logger for distributed setups.
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
 
+  if ENV["RAILS_LOG_TO_STDOUT"].present?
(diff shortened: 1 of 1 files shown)


monotek (approved):
Nice! :-)

I think this would help in the docker  installations too, where we use a workaround in the moment. See:
https://github.com/zammad/zammad-docker-compose/blob/master/containers/zammad/setup.sh#L36

thorsteneckel:
Merged to develop with the commit above 👍 Thanks a lot for your contribution <3

Issue #2021, opened 2018-05-16
backup script: tar - file changed warnings


lr1980:
### Infos:

* Used Zammad version: latest
* Installation method (source, package, ..): any
* Operating system: any
* Database + version: any
* Elasticsearch version: latest
* Browser + version: any

### Expected behavior:

* backup script that can used without errors or warnings (=no mails/logs from cron)

### Actual behavior:

* backup script can produce following warnings:
`tar: opt/zammad/tmp/cache_file_store_production: file changed as we read it`

It's safe to exclude the "tmp" directory?

thegcat:
We're having a similar issue (we're using borgbackup for backups, but the issue is the same). My experience with Rails apps is that `RAILS_ROOT/tmp` is safe to ignore, but I'd like a confirmation for this too.

If this is confirmed I'd be willing to contribute a patch accordingly.

thegcat:
Judging from https://github.com/zammad/zammad/blob/develop/lib/tasks/zammad/flush/cache.rake I'd say those directories (`tmp/cache*`) can be ignored when creating backups.

monotek:
Ups. Totally missed that ticket... PR is very welcome :-)

MrGeneration:
Yes, ignoring `tmp/` is safe to do. :-)

thegcat:
See PR in #2638

thorsteneckel:
Cool! Thanks @thegcat. @monotek can you please review the PR when you find the time? I'll do the GitLab -> GitHub dance then.

thegcat:
This probably can be considered fixed by #2638 and closed.

thorsteneckel:
Yes! Thanks a lot @thegcat for your contribution.

PR #2638, opened 2019-06-30
Exclude tmp, log, cache dirs from backups

thegcat:
Exclude `tmp`, `log`, `.bundle` and `vendor` dirs from backups. See also discussion in #2021

`.bundle` might be present in your zammad installation directory depending on how you installed it. For example the default from source installation documentation puts the home directory of the zammad user in the root of the zammad installation dir https://docs.zammad.org/en/latest/install-source.html#add-user

`vendor` might be present depending on how you install gems. Those can be recreated by running `bundle install` again, so it's no necessary to back them up, and they're quite large (355MB out of the 776MB of the whole zammad install in our case).

I'll be happy to make any necessary changes based on comments, thank you.

### DIFF
--- a/contrib/backup/functions
+++ b/contrib/backup/functions
@@ -40,7 +40,7 @@ function backup_dir_create () {
 
 function backup_files () {
   echo "creating file backup..."
-  tar -C / -czf ${BACKUP_DIR}/${TIMESTAMP}_zammad_files.tar.gz ${ZAMMAD_DIR#/}
+  tar -C / -czf ${BACKUP_DIR}/${TIMESTAMP}_zammad_files.tar.gz --exclude='tmp' ${ZAMMAD_DIR#/}
   ln -sfn ${BACKUP_DIR}/${TIMESTAMP}_zammad_files.tar.gz ${BACKUP_DIR}/latest_zammad_files.tar.gz
 }
 


monotek (commented):
imho we should only remove tmp for now.
log could have informations you want in a backup when you restore a crashed server.

@thorsteneckel & @martini 
i'm not sure about the bundle dir but i think vendor is needed for sure?

thorsteneckel:
I agree. We should/need to keep those. 👍

monotek (changes requested):
please remove all excludes beside tmp.

thegcat:
> vendor is needed for sure?

Ah, I had only noticed that it's large on our installation because it has the installed gems in `vendor/bundle`, so I could exclude just that instead of the whole `vendor`.

The `.bundle` is definitely not something that comes from zammad as such but is something that `bundler` throws into the user's home dir, which in some installations might be same dir zammad is installed to.

The underlying question probably is: Should the backup be a "unpack the backup and you're good to go" thing, or a "unpack the backup and re-install the gems and possibly other deps" thing.

thegcat:
> please remove all excludes beside tmp.

Done.

thorsteneckel:
I was on my way merging this when I realized that `logs` most likely will change as well while performing the backup. This will result in the very same warning #2021 complains about. So my proposal would be to exclude only `tmp` (which is save to ignore and probably even creates a more coherent backup) and add `--warning=no-file-changed`. However, this will lead to a behavior where other files that we should care about don't get logged 🤔 Any thoughts?

Regarding your question: I'd vote for `"unpack the backup and you're good to go"`-backup 👍

monotek:
hmmm.... if we would use stdout logging by default ;-)

see also: https://github.com/zammad/zammad/pull/2637/files

thegcat:
> I was on my way merging this when I realized that `logs` most likely will change as well while performing the backup. This will result in the very same warning #2021 complains about. So my proposal would be to exclude only `tmp` (which is save to ignore and probably even creates a more coherent backup) and add `--warning=no-file-changed`. However, this will lead to a behavior where other files that we should care about don't get logged 🤔 Any thoughts?

I'm rather on the "ensure consistent backup" rather than "the backup might be corrupt but at least there were no errors" side.

As for wether logs should be included in the backups: In a DR scenario I'm more interested in having the newest possible state and being able to restore it, even if it's without the logs, rather than having the logs but a potentially older backup because of inconsistent state or warnings that broke the backup off.

Furthermore, our logs grew to 1,3GB, so I'd in any case strongly recommend not keeping the logs as-is. People unfamiliar with Rails might just not be aware that the logs are not rotated/zipped and probably are not comfortable with changing the defaults. People familiar with Rails might just forget that the default is to dump the log to `/log` and have it grow indefinitely (my case 😅). While the logs discussion is another one altogether, I don't think having logs, especially one with an "infinite" file size, is harmful rather than helpful.

> Regarding your question: I'd vote for `"unpack the backup and you're good to go"`-backup 👍

OK.

thorsteneckel:
Oh wow 😱 Thanks a ton @thegcat ! Both PRs are currently running through our internal CI. LGTM! Regarding the not rotating logs: Would be cool to address this in the
[community board](http://community.zammad.org/) and see what we can make out of it 🚀

monotek:
> Regarding the not rotating logs: Would be cool to address this in the
> [community board](http://community.zammad.org/) and see what we can make out of it

i don't think zammad should find a way to rotate logs. its not an os ;-)
logging to stdout and using the os logrotate feature would be the way to go for me.

thegcat:
> i don't think zammad should find a way to rotate logs. its not an os ;-)
> logging to stdout and using the os logrotate feature would be the way to go for me.

I also think that's rather a concern for the launcher/the "OS"/whatever but not for the app itself.

As mentioned in #2637 I looked for a way to log to STDOUT as this is the preferred way for systemd launched applications. From what I read above it would also be the preferred way for docker installations. Maybe having #2637 and making it the default by adding `RAILS_LOG_TO_STDOUT=1` in the `zammad.env` would already be enough.

thorsteneckel:
Merged via the commit above. Thanks a lot!

PR #2639, opened 2019-07-01
Removed duplicate `Mailing-List` tag from postmaster match ui element

maxxer:
from postmaster filter create


martini:
@maxxer Thanks! 🤓

Issue #2741, opened 2019-09-16
SAML authentication


thorsteneckel:
### Infos:

* Used Zammad version: 3.1.x
* Installation method (source, package, ..): any
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: any

### Expected behavior:

* There should be the possibility to log users in via SAML (Keycloak, Chipotle, etc.) as it's possible for other third party providers (Office 365, GitHub etc.)

### Actual behavior:

* SAML login is currently not supported

peacepenguin:
Hey @thorsteneckel I got the latest develop version 3.2.0-1568666188.ba2afc0c.bionic installed to test the SAML integration commit: https://github.com/zammad/zammad/commit/a93250e210a1b2fcbfe61843dd2e3f6c3451ff12

I see locally installed files that include the SAML bits that were committed, but all the SAML configuration fields are missing under Security -> Third party applications.

Let me know if there's anything I can do to help!

Thanks!

MrGeneration:
If you installed via source, don't forget to precompile :>

(zammad run) rake assets:precompile

peacepenguin:
I used packager.io on the develop branch to get the bits on a new ubuntu 18.04 server vm. But I'll try compile from source to test the latest straight from GitHub.

I do notice the settings for the other auth providers are defined and listed here:   https://github.com/zammad/zammad/blob/develop/db/seeds/settings.rb

I don't see anything SAML related in that file, whereas settings for all the other third party auth providers have detailed listings that seem to also define the gui elements. 

All that said i'm new to ruby/rails in general, and very new to the architecture of Zammd, so please forgive me if i'm way off base here and I just need to compile it correctly :)

MrGeneration:
That's correct, because this will come with an migration :) ( https://github.com/zammad/zammad/blob/develop/db/migrate/20190715141227_saml_auth.rb )

Maybe also `zammad run rake db:migrate` followed by an restart might help.

If you want to be 100% sure, you could run the install-script manually:

`/opt/zammad/contrib/packager.io/postinstall.sh`

peacepenguin:
I tried the fixes above without success using the packager.io develop packages.

Now switched over to a new vm, (ubuntu 18.04 server) Got the latest sources from GitHub develop branch, got it installed without too much trouble, but still no SAML config in the GUI.

[...]

PR #2744, opened 2019-09-17
SAML missing from settings.rb

peacepenguin:
This pull request fixes the SAML issue I've been investigating since the SAML commit was added yesterday related to the SAML issue: https://github.com/zammad/zammad/issues/2741

I didn't write any of the code used in this pull request, it all came from the migration file: 20190715141227_saml_auth.rb

With this pull request applied I have successfully tested the SAML integration with the develop branch, and it works great. Without this patch the SAML gui elements are missing, and the database table: "settings" does not contain the needed rows for "auth_saml" and "auth_saml_credentials".

### DIFF
--- a/db/seeds/settings.rb
+++ b/db/seeds/settings.rb
@@ -1169,6 +1169,78 @@ Setting.create_if_not_exists(
   frontend:    false
 )
 Setting.create_if_not_exists(
+  title:   'Authentication via %s',
+  name:'auth_saml',
+  area:'Security::ThirdPartyAuthentication',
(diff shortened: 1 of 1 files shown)


thorsteneckel:
Thanks a ton @peacepenguin! Unfortunately we can't merge PR at the moment so this is why I had to take your changes and add them in a [commit](https://github.com/zammad/zammad/commit/dfb2b1bc8ae27b91eca82d9f7380ae7c045f0a18). However, great catch - and thanks making the effort digging in.

PR #2859, opened 2019-12-13
Added options for external database server

dreieinhalb:
Database config file has allowed only localhost database server. If you
want to use a database on a different host, you need the ability to set
hostname and port
- for MySQL see:
"https://stackoverflow.com/questions/5872264/correct-mysql-configuration-for-ruby-on-rails-database-yml-file"
- for PostgreSQL see:
"https://gist.github.com/jwo/4512764#file-postgres-database-yml"

### DIFF
--- a/config/database/database.yml
+++ b/config/database/database.yml
@@ -14,11 +14,19 @@ default: &default
   # username: zammad
   # password:
 
+  # If the database server is not on localhost, you can set hostname and port:
+  # host:
+  # port:
+
   #### mysql config #####
 
   # adapter: mysql2
   # username: zammad
   # password: <Password>
+
+  # If the database server is not on localhost, you can set hostname and port:
+  # host:
+  # port:
 
 production:
   <<: *default


MrGeneration:
Might be usefull for other users, if these directives work

dreieinhalb:
> Might be usefull for other users, if these directives work

I tested it for PostgresSQL and it works. For MySQL I have to confess that I relied on referenced StackOverflow question.

In bigger companies/institution (as in my case) databases are managed centrally by DBAs, so I think it can be useful for other users too.

thorsteneckel:
Merged via https://github.com/zammad/zammad/commit/5f861c566d9755a0c3f7c53b6c0a842a9f15465a - unfortunately we still lose the author because of our merge pipeline/process :( Thank you anyway for your contribution @dreieinhalb ! Closing now.

monotek:
>unfortunately we still lose the author because of our merge pipeline/process 

This is imho a no go and we should address this asap. 
Maybe 36c3 is good time to search for a solution.

thorsteneckel:
Count me in! Open to any approach to improve it.

PR #2863, opened 2019-12-16
fix nginx X-Forwarded-Proto

### LINKED ISSUE #2829: CSRF token verification failed
### Infos:

* Used Zammad version: 3.2
* Installation method (source, package, ..): YUM
* Operating system: Centos 7
* Database + version:
* Elasticsearch version:
* Browser + version:

### Expected behavior:

Upgraded to Zammad 3.2. and I get error message csrf token verification failed. Cannot log in. Newer before seen this error message.

### Actual behavior:

Cannot log in. Linux setup is stayed same. 

*

lr1980:
forward the correct X-Forwarded-Proto

fixed #2829 (and yes this is a bug)

https://community.zammad.org/t/fresh-install-3-2-x-csrf-token-verification-failed/3080/32?u=lr1980

### DIFF
--- a/contrib/nginx/zammad.conf
+++ b/contrib/nginx/zammad.conf
@@ -10,6 +10,11 @@ upstream zammad-websocket {
     server 127.0.0.1:6042;
 }
 
+map $http_x_forwarded_proto $real_scheme {
+    default $http_x_forwarded_proto;
+    '' $scheme;
+}
+
 server {
     listen 80;
 
@@ -25,11 +30,6 @@ server {
 
     location ~ ^/(assets/|robots.txt|humans.txt|favicon.ico) {
         expires max;
-    }
-    
-    map $http_x_forwarded_proto $real_scheme {
-        default $http_x_forwarded_proto;
-        '' $scheme;
     }
 
     location /ws {
--- a/contrib/nginx/zammad.conf
+++ b/contrib/nginx/zammad.conf
@@ -26,6 +26,11 @@ server {
     location ~ ^/(assets/|robots.txt|humans.txt|favicon.ico) {
         expires max;
     }
+    
+    map $http_x_forwarded_proto $real_scheme {
+        default $http_x_forwarded_proto;
+        '' $scheme;
+    }
 
     location /ws {
         proxy_http_version 1.1;
@@ -33,7 +38,7 @@ server {
         proxy_set_header Connection "Upgrade";
         proxy_set_header CLIENT_IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header X-Forwarded-Proto $real_scheme;
         proxy_read_timeout 86400;
         proxy_pass http://zammad-websocket;
     }
@@ -42,7 +47,7 @@ server {
         proxy_set_header Host $http_host;
         proxy_set_header CLIENT_IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header X-Forwarded-Proto $real_scheme;
         proxy_read_timeout 300;
         proxy_pass http://zammad-railsserver;
 


contrib/nginx/zammad.conf
```diff
@@ -26,14 +26,19 @@ server {
     location ~ ^/(assets/|robots.txt|humans.txt|favicon.ico) {
         expires max;
     }
+    
+    map $http_x_forwarded_proto $real_scheme {
+        default $http_x_forwarded_proto;
+        '' $scheme;
+    }
```
waja: The map should be used outside of the `server` context. When using the config this way I'm getting:
```
nginx: [emerg] "map" directive is not allowed here in /etc/nginx/sites-enabled/default:30
```

waja (changes requested):
@lr1980 Have you tested this config? I'm running into not starting nginx, when using `map` in the `server` context. Please put it on top of the `server` context.

thorsteneckel:
Hey @lr1980 - Thanks a lot for your contribution! This will help a lot of people out there.

I just checked the [NGINX documentation of this feature](http://nginx.org/en/docs/http/ngx_http_map_module.html). It was added in 2010 and got extended over  the years. It is required to have the module [`ngx_http_map_module`](http://nginx.org/en/docs/http/ngx_http_map_module.html) activated, which is [the default behavior](https://github.com/nginx/nginx/blob/953f53921505a884f3912f2d8db5217a71c0479a/auto/options#L80). So from the maintenance point of view I'd give green lights 👍 

However, there is still a syntax issues as @waja pointed out. Would you mind resolving it @lr1980 ?

contrib/nginx/zammad.conf
lr1980: you are absolutely right... it's only allowed in the http context...

lr1980:
there is still a security problem here ... in our environment (with loadbalancer) we can ensure that proto field comes from our loadbalancer. In an environment where the client / browser connects directly, it is possible that the proto field was "manipulated" by the client. i.e. if the field is used by zammad to ensure security, it is no longer possible.

I do not know the rails code that evaluates the field and therefore cannot judge how you can ensure that safety is guaranteed in every environment. it would be possible to additionally evaluate the X-Forwarded-For field to determine whether the server / client was local (private ip) or not local in front of the nginx. which also poses problems again => an own proxy / loadbalancer could also use a public ip...

this is how i come to the conclusion ... that my suggestion is a solution for our environment, but probably not a reliable (safe) solution for all environments. depending on what exactly happens or is evaluated on the rails side...

PS: the hack to write "https" directly into the config has the same security problems ...

MrGeneration:
Thanks for your feedback, maybe @thorsteneckel can give feedback on the ruby perspective.
I'll be waiting with commiting these changes until then as I don't want to raise security issues. :)

htuscher:
I'm also not a Ruby guy, but that security issue is usually solved by setting `TRUSTED_PROXIES` in the application.

michaldaniel:
The pull request proposes a solution for nginix only, I believe it's also necessary to reflect the changes in apache2 configuration. Something akin to:

    RequestHeader set X-Forwarded-Proto 'https'
    RequestHeader set X-Forwarded-Port 443
    RequestHeader set X-Forwarded-Ssl on

MrGeneration:
I'll close this PR as I took your changes / suggestions to our internal git.
While I can't use your commit (and thus commiter information is lost) I did reference both your commits and Github account in the reworked commit of mine.

I hope that's okay to you!

Thank you very much for your patience and time!
My commits will reference https://github.com/zammad/zammad/issues/2829 in case you're wondering. :-)

benbucksch:
> I can't use your commit (and thus commiter information is lost)

For future reference: `git commit --author 'His name <him@example.com>'`

MrGeneration:
Thanks for the hint. "Can't use your commit" because my software suite I am using does not support it.

thorsteneckel:
Hey there! So long story short is that we can't take this PR over to the core. As @lr1980 [mentioned earlier](https://github.com/zammad/zammad/pull/2863#issuecomment-567029245) it exposes a security issue for non load-balancer setups.

I'll address the further processing in issue #2829.

PS @MrGeneration : Please refer to our internal wiki documentation on how to keep the original PR commit author.

PR #2869, opened 2019-12-18
Added Swedish chat translation

pitah81:
Added long awaited Swedish translation to the chat module


thorsteneckel:
Merged with the referenced commits! Thanks a ton @pitah81 . Please note that we currently lose the authors of the GitHub contributions 😞This is because of CI and review setup. However, @monotek and I will have a look on how to improve things next week 🤞 
Packages will be ready in about 20 minutes from now.

PR #2875, opened 2019-12-22
Don't add a blank line to the signature on reply

chrisblech:
Unfortunately it's not possible to create reply-templates, so we (and maybe other users too) include the whole template in the signature, starting with the greeting - and this should start in the very first line of the reply.

If someone needs a blank line above his signature, he can add it to the signature manually, so this change doesn't break the former behaviour.

See also: https://community.zammad.org/t/empty-line-above-signature/2870

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee
@@ -303,7 +303,6 @@ class EmailReply extends App.Controller
         App.Utils.htmlStrip(signature)
         if signaturePosition is 'top'
           body.prepend(signature)
-          body.prepend('<br>')
         else
           body.append(signature)
         ui.$('[data-name=body]').replaceWith(body)


MrGeneration:
In my opinion the greeting formular shouldn't live in a signature.
Beside that (in my opinion) rather specific use case, the blank line in the signature makes sense, because if I don't have one I'd need to get my cursor in front of the first character/letter and then create one just to start to type my answer to the customer.

This affects **all** existing signatures on every Zammad installation which is quite a big impact.
At least there should be a migration of existing signatures to add a blank line by default.

Still then I'd say this is rather not intuitive thing to do.

thorsteneckel:
Hi @chrisblech - thanks for your proposal and contribution! I reviewed your change, read your community post and replied to it. [We need this line to detect the signatures in the communication properly (on a technical level) to be able to do a proper quoting of previous communication](https://github.com/zammad/zammad/commit/9c02d2e02b732bf4cb2845b389d353e5570e213d). Please see my proposal in [the community post](https://community.zammad.org/t/empty-line-above-signature/2870/3) to ease your process. Let me know if you disagree in the community post to keep GitHub technical. Closing now.

PR #2877, opened 2019-12-27
Update zammad_ssl.conf

jannheider:
Fix for: CSRF token verification failed #2829

Update for "CSRF token verification failed!" problem in nginx ssl conf.


MrGeneration:
This is a duplicate of https://github.com/zammad/zammad/pull/2863
Please note that we're currently investigating a good solve for both apache and nginx.

PR #2878, opened 2019-12-29
Add apple-touch-icon to HTML head

xf-:
This helps browser to load the existing icon.

Example screenshot Firefox [Top sites](https://support.mozilla.org/en-US/kb/customize-new-tab-page):

**Current** tile in "Top sites":
image

**Afterwards**
image

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -5,6 +5,7 @@
   <meta name="viewport" content="width=1024">
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <title><%= Setting.get('product_name') %></title>
+  <link rel="apple-touch-icon" href="apple-touch-icon.png" />
   <%= stylesheet_link_tag "application", :media => 'all' %>
   <%= stylesheet_link_tag "application-print", :media => 'print' %>
   <% if Rails.configuration.assets.debug %>


thorsteneckel:
Thanks @xf- for the PR! This has been merged into develop with the commit above.
Unfortunately we're losing the reference as you as an author. That's why we add the PR link to the commit and add your handle to the commit message. Hope this works for you.

Closing now.

xf-:
> Thanks @xf- for the PR! This has been merged into develop with the commit above.
> Unfortunately we're losing the reference as you as an author. That's why we add the PR link to the commit and add your handle to the commit message. Hope this works for you.

@thorsteneckel 
No problem in this case, but i don't understand the reason? Cherry-picking, loading a patch (https://github.com/zammad/zammad/pull/2878.patch) ... squashing online @github & rewrite the Message ... all will preserve the author.

About Icons: 
Maybe use https://realfavicongenerator.net/ and you will also get Android Icons, title, Windows tiles, Safari icon ... and other settings.

thorsteneckel:
Thanks for the link @xf- will check it out!

Unfortunately it's not that easy/straight forward in our setup. We have an internal GitLab server that runs our [comprehensive CI test suite](https://github.com/zammad/zammad/blob/develop/.gitlab-ci.yml) which we can't reflect on GitHub (Travis, CircleCI etc.) without spending 4x-5x the money.

EDIT:
Maybe some more information to avoid confusion: It's required to have a single source of truth in a synced state setup. In our case it has to be GitLab because of the CI. So we can't run the PR through the CI and then merge the PR on GitHub. This will break the sync as it already did quite a few times in the past :(

PR #2882, opened 2020-01-09
fix(optional_open_button): check if button exists

### LINKED ISSUE #2881: Chat button is no longer optional in no jQuery version.
Hi,

i am currently switching from your `chat.min.js` to `chat-o-jquery.min.js`. In our application there is no need for jQuery and we are currently importing it just for the chat.
We've now realized that there is also a no-jquery version available, so we are now migrating.

We don't have a Button with the class `.open-zammad-chat` ([default Zammad buttonClass](https://github.com/zammad/zammad/blob/3.3.0/public/assets/chat/chat-no-jquery.js#L357)) on every single page.
Unfortunately this is required in your render function.

https://github.com/zammad/zammad/blob/3.3.0/public/assets/chat/chat-no-jquery.js#L745
expects a `classList` on whatever the query selector returns. If the result is null (No Button on a page) it will therefore crash.

Could you please make it optional? (Like you're doing here: https://github.com/zammad/zammad/blob/3.3.0/public/assets/chat/chat-no-jquery.js#L1080)

In your jQuery version it's optional because jQuery allows you to call `addClass` on empty results.

I can also create a PR for this if PR's are welcome.

Thank you very much!

ma-jahn:
Closes #2881


thorsteneckel:
Hey @ma-jahn - thanks a lot for creating the issue and providing the MR to resolve it 💪Would you mind adding this to `public/assets/chat/chat.coffee` as well? Otherwise we have a (confusing) inconsistency.

ma-jahn:
Hi @thorsteneckel , thanks for the quick feedback!
I've not made any changes to chat.coffee since it's using a jquery selector in all of these cases and it's 'fail safe'.
So if jQuery doesn't return an element (or collection of elements) you can still run `addClass`, `removeClass`, etc. on the selector result.

So the optional_open_button bug did only affect the no-jquery version of your chat.

I hope the above makes sense :)

thorsteneckel:
Perfect! Thanks for your lightning fast feedback. That makes total sense 💡 One last question: Have you applied the changes to the JS files by running the gulp pipeline?

ma-jahn:
Yes :) i did only change the chat-no-jquery.coffee and then run `gulp no-jquery` to generate the js files.
My IDE also did lint the files which is why there are a couple more changes like:
`function (__obj)` --> `function(__obj)`

thorsteneckel:
Sei bellissimo! I'll have to run it through our internal CI first but will merge this today or tomorrow. Do you need it for stable or is develop sufficient? Full disclosure: Due to our current setup the original you as the author will be lost :( I'll reference you in the commit message. I hope this is fine with you.

ma-jahn:
That's great. I've implemented a workaround on our end, so it's not that urgent.
But once you publish the next stable release (maybe even incl. #2883) we'll update our production instances.
I'm watching your repo so i'll be notified as soon as the next stable release is available.

Don't worry about the author issue at all :-)

thorsteneckel:
Cool! Applied the changes with the commit above as discussed. I'll have a look at #2883 now 👍

PR #2905, opened 2020-01-23
fix(px_handling): broken px handling in no-jquery coffee version

ma-jahn:
Hi,

our Zammad Release (moving from a Google Tag Manager integration to a direct integration within our SSR Vue.js App) hit production yesterday.
Now we're experiencing a bit of noise in our Error Tracking Solution. The Reason for it are 3 little syntax errors within the chat-no-jquery.coffee.

It was using `+'px'` which is actually being compiled into a function call.
So `width +'px'` doesn't append the string 'px' to `width`. It actually results in a `width(+'px')` call.


thorsteneckel:
Fixed with the commit above thanks @ma-jahn and @mantas !

PR #2908, opened 2020-01-27
Fix storage log message typo

saz:


mantas:
Thanks @saz for the PR! This has been merged into develop with the commit above.

Unfortunately we're losing the reference as you as an author. That's why we add the PR link to the commit and add your handle to the commit message. Hope this works for you.

PR #2916, opened 2020-02-03
[TASK] Add "screenshot" to check attachement references

helsner:


helsner:
Would be nice to have within the german check too, as we use it on daily base (i would guess that i'm not the only one with this)

MrGeneration:
Personally I'm not so sure for the word "screenshot" to be honest. 
I mainly use inline screenshots and then reference "see screenshot below".

The issue with that will be that it will tell me "hey, are you sure, you may want to attach stuff..." which would slow down my workflow.

Might just be me, so it's just my opinion.

helsner:
I guess we can close this PR.
There has been no response in the community issue too.

Issue #2714, opened 2019-08-21
Roles and Users selection in Overview administration UI ambiguous


SEGGER-NV:
### Infos:

* Used Zammad version: 3.1.x
* Installation method (source, package, ..): package
* Operating system: Debian 9
* Database + version: -
* Elasticsearch version: -
* Browser + version: -

### Expected behavior:

* When setting the availability of an overview one can set it for a role and/or individuals.
When setting both a role and an individual the overview should be visible to everyone that has the role and to the individual.

### Actual behavior:

* Instead the two availability settings seem to operate as a logical "and" when both are set. So  if you set both the overview will only be visible if the individual has also the role. The role group itself will lose the overview completely. Even worse if the individual does not have the role, then neither the other role members nor the individual user will see the overview.

If the options would be evaluated as an logical "or" it would work as expected.
As a simple switch to "or" would probably break backwards compatibility the fix proposal would be to add a switch to switch between both modes.

### Steps to reproduce the behavior:

* Create new role
* Give two users new role
* Create new overview where overview availability settings are new role and user1
* Log into user2, now you don't see the overview, only user1 can see it

Yes I'm sure this is a bug and no feature request or a general question.

MrGeneration:
That's unexpected. Thanks for your report!

Documentation states:
`With these settings you can choose whether to make the overviews available to an entire group of people (by selecting the role) or to specific users. The entries in both fields apply. This means that you can also select individual users in addition to sharing the overview for all role members.`

(https://admin-docs.zammad.org/en/latest/manage-overviews.html)

So we should address this :)

thorsteneckel:
Hey there. @mantas @martini and I had a discussion about what the current implementation is and what is expected. We found out that the difference those two is the root cause of this issue: The UI is misleading and the documentation states a different behavior than what the implementation actually does. 

# Overview

## What is expected (Case 1)
An Overview can be configured to be available for a List of Roles **or** Users. The Role and User selections are optional on their own but at least a selection in one of them is required. This behavior is called **additive** as a technical term: Users that are not already covered in one of the possibly selected Roles will be added. This translates in the following variations:

- All Users of the selected Roles will see the Overview.
- All selected Users of the Users list will see the Overview.

Caveat: There is no way to limit the Users of the selected Roles. Each User has to be added manually via the Users list selection or a new Role has to be created.

## What is implemented (Case 2)
[...]

PR #2928, opened 2020-02-11
Issue 2714: fixed Ticket::Overview.all query

fleverest:
#2714 

The query restricted the `current_user` to only see the union of the following overview lists:
List 1: Overviews* where `current_user` is a member of one `overviews_roles` AND is a member of `overviews_users`.
List 2: Overviews* where `current_user` is a member of one of `overviews_roles` AND `overviews_users` is `nil`.

*(After overview filters are placed)

I can't imagine why this was the original query, but this should fix the issue so as to comply with the documentation and what the UI suggests.

### DIFF
--- a/app/models/ticket/overviews.rb
+++ b/app/models/ticket/overviews.rb
@@ -31,7 +31,7 @@ returns
       if links.present?
         overview_filter[:link] = links
       end
-      overviews = Overview.joins(:roles).left_joins(:users).where(overviews_roles: { role_id: role_ids }, overviews_users: { user_id: nil }, overviews: overview_filter).or(Overview.joins(:roles).left_joins(:users).where(overviews_roles: { role_id: role_ids }, overviews_users: { user_id: current_user.id }, overviews: overview_filter)).distinct('overview.id').order(:prio, :name)
+      overviews = Overview.joins(:roles).left_joins(:users).where(overviews_roles: { role_id: role_ids }, overviews: overview_filter).or(Overview.joins(:roles).left_joins(:users).where(overviews_users: { user_id: current_user.id }, overviews: overview_filter)).distinct('overview.id').order(:prio, :name)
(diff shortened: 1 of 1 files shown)


fleverest:
@martini Could you confirm? I checked the history and this query was implemented in your commit:
https://github.com/zammad/zammad/commit/c6cb0e553161ca3dca9f16e2eb7b09be43867b42#diff-9d620b896bc76fdff2a75aa22269fdb5

So you'd probably know better than me

thorsteneckel:
Hey @fleverest - nice to see another pull request from you :) We'll check it out. @mantas - could you please have a look?

fleverest:
Always happy to contribute, @thorsteneckel

mantas:
Hi @fleverest ,

Thank you for bringing our attention to this issue. However, we decided to solve this issue by clarifying how current code works in user interface instead.

Again, thank you for your contribution. Even if your change technically didn't get merged - it did help a lot!

Closing now.

fleverest:
Clarifying in the UI is a step in the right direction, but I still find this to be quite a counter-intuitive design choice.

Consider an overview that belongs to all members of the 'developer' role, and suppose you would also like one other agent (not a developer) to see this particular overview. Then, in order to achieve this, you will have to select 'developer' for the available roles AND a role that the other user is a member of, and also select every single user in the 'developer' role in the available users, as well as the user you'd like to add. Then, whenever you add a new developer to the team, you will need to add them to every overview you've created as a user. Can you see the problem with this?

mantas:
Hi @fleverest ,

I totally agree that current workflow is not 100% ideal. The plan is to come back to this issue later. Unfortunately the quick fix in your PR may break someone else's existing overviews.

fleverest:
Thanks @mantas,

I see there has been a deeper discussion on the original issue now

PR #2929, opened 2020-02-11
Consistently handle the 'disabled' attr in inputs

abeluck:
As discussed in https://community.zammad.org/t/automating-creation-of-custom-object-attributes/3831/5

This commit ensures that the generic input widgets handle the `disabled`
attribute consistently. Before only the checkbox, select and input widgets respected
this attribute.

I've added it to the following widgets

* tree/searchable select
* textarea
* datetime
* date

Now it is possible to define a screens policy like so:
```
screens: {
"edit"=>{
"ticket.customer"=>{"shown"=>true, "required"=>false},
"ticket.agent"=>{
  "shown"=>true,
  "required"=>false,
  "disabled"=>true # <--- add this
}}}
```

Then the fields will show up on the Agent's sidebar with the customer
submitted values, but the Agent will not be able to change the values.

This is not the same as real permissions, as presumably a manually
crafted API request could circumvent the disabled attribute.

But this is a first step towards that direction, and very useful for us
in prod right now.

### DIFF
--- a/app/assets/javascripts/app/views/generic/attachment.jst.eco
+++ b/app/assets/javascripts/app/views/generic/attachment.jst.eco
@@ -1,5 +1,6 @@
 <div class="attachments"></div>
 <div class="article-attachment">
+  <% if !@attribute.disabled: %>
   <div class="attachmentPlaceholder">
     <span class="attachmentPlaceholder-inputHolder fileUpload">
       <% uid =  _.uniqueId('fileUpload_') %>
@@ -9,6 +10,7 @@
       </label>
     </span>
   </div>
+  <% end %>
   <div class="attachmentUpload hide">
     <div class="horizontal">
       <div class="u-highlight">
--- a/app/assets/javascripts/app/views/generic/checkbox.jst.eco
+++ b/app/assets/javascripts/app/views/generic/checkbox.jst.eco
@@ -1,7 +1,7 @@
 <div class="<%= @attribute.class %> checkbox checkbox--list">
 <% for row in @attribute.options: %>
   <label class="checkbox-replacement<%= ' is-disabled' if row.disabled %>">
-    <input type="checkbox" value="<%= row.value %>" name="<%= @attribute.name %>" <%= row.checked %> <%= row.disabled %>>
+    <input type="checkbox" value="<%= row.value %>" name="<%= @attribute.name %>" <%= row.checked %> <%= row.disabled %> <% if @attribute.disabled: %> disabled<% end %>>
     <%- @Icon('checkbox', 'icon-unchecked') %>
     <%- @Icon('checkbox-checked', 'icon-checked') %>
     <span class="label-text"><%= row.name %></span>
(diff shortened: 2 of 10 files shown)


thorsteneckel:
Hey @abeluck - wow! This was like instantly fast. Would you mind sharing some screenshots on how the elements look when they are disabled?

abeluck:
Your response was instantly fast :) I was in the process of creating a screenshot while you typed that :)

zammad-consistent-disabled

"Custom attribute" and "Test_ttset" are both intentionally not disabled.

thorsteneckel:
That looks great! Would you also mind to add some tests to cover this behavior? `public/assets/tests/form.js` would be the file. We'd be happy to answer your questions.

abeluck:
Sure!

How could I run just those frontend tests? `rake test` seems to run the whole suite.

thorsteneckel:
The most convenient way is to start Zammad in `development` mode and navigate your browser to http://localhost:3000/tests_form 🚀

abeluck:
Gotcha, thanks.

I'm having trouble finding a test that will capture the whole thing from adding `disabled: true` in the screens to rendering it disabled... so that I can modify it for this usecase :)

Is there an existing test that takes a object attribute params object and renders it? The closest I see are `object manager form 1/2/3` but those don't render the final form the agent sees.. just the admin ui. 

If you could point me to it, then I can base my test on it.

abeluck:
I added basic render tests (see new changes) that use the "disabled" attribute in the form params object. This works great!

Along the way I added disabled support for `radiobox`, `richtext`, and `checkbox` just to be consistent, though ObjectManager::Attributes cannot have these types. I intentionally did not add disabled support for `password` or `active` as I don't think it makes sense for those.

If you also want a test that tests all the way from the ObjectManager::Attribute screens hash, then I'll need some help as per my previous comment.

abeluck:
Here are screenshots of the new widgets

zammad-disabled-check
zammad-disabled-radio

I didn't like how the richtext edit still looked editable, even though it wasn't, so I added some appropriate css and hid the "select attachment" link.

before css (it looks editable even though it isn't):
zammad-disabled-contenteditable

after css (now it looks properly disabled):
zammad-disabled-contenteditable2

thorsteneckel:
Damn sweet! @mantas could you please have a look and create a MR to check the impact on CI?

abeluck:
@mantas @thorsteneckel  I know a release is imminent, and chance this PR and/or  [Fix deserialization of multiselect value #2932](https://github.com/zammad/zammad/pull/2932) could make it in?

thorsteneckel:
@mantas could you please prioritize this PR?

mantas:
@thorsteneckel sure, I'll look into this today

mantas:
Hi @abeluck ,

Unfortunately this PR won't make it to the upcoming release. While PR itself is great, it brought our attention to a semi-related part of the codebase that may need some love before pulling this one in. We hope to include this in next release.

Sorry for the delay.

abeluck:
Understood. Looking forward to the next release then.

abeluck:
Now that we're passed the big 3.3 release, any chance we could get this PR looked at again?

PR #2932, opened 2020-02-14
Fix deserialization of multiselect value

abeluck:
When a multiselect Attribute is used with the Ticket object, the rails
backend persists the multiselect value as a serialized JSON array in the
database.

When this value is read back and passed down to the form widget to
render, it -- before this patch -- was simply rendering the serialized
JSON array string as an option.

This patch fixes the widget so that it will attempt to deserialize the
value as an array to render the multi selected options properly.

For the case where the passed value serializes to something other than
an array, it falls back on the default existing behavior of just
rendering the string as is.

In the event some bad data gets persisted to the database, we do not
want to break the UI. So we swallow JSON parse errors and fall back on
the existing default behavior of just rendering the string as is.

### before
multi2

### after
multi1

### DIFF
--- a/app/assets/javascripts/app/controllers/_ui_element/select.coffee
+++ b/app/assets/javascripts/app/controllers/_ui_element/select.coffee
@@ -8,6 +8,15 @@ class App.UiElement.select extends App.UiElement.ApplicationUiElement
     else
       attribute.multiple = ''
 
+    if attribute.multiple && _.isString(attribute.value) && !_.isEmpty(attribute.value)
+      try
+        value = JSON.parse(attribute.value)
+        # only accept serialized arrays, nothing else
+        if _.isArray(value)
+          attribute.value = value
+      catch e
+        console.log("Invalid json passed as attribute value to selectmulti")
+
     if form.rejectNonExistentValues
       attribute.rejectNonExistentValues = true
 
--- a/public/assets/tests/form.js
+++ b/public/assets/tests/form.js
@@ -133,6 +133,9 @@ test("form params check", function() {
     select4: undefined,
     selectmulti1: false,
     selectmulti2: [ false, true ],
+    selectmulti3: '["a","b"]',
+    selectmulti4: '{"notan": "array"}',
+    selectmulti5: '{"invalid":, "json"',
     selectmultioption1: false,
     selectmultioption2: [ false, true ],
     autocompletion2: 'id2',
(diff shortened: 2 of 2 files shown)


thorsteneckel:
Hi @abeluck - is there an existing issue for this? @mantas - could you please have a look at this? Thanks!

abeluck:
@thorsteneckel There aren't any issues open about this, no. Rather than make an issue, I just submitted a PR. But I'd be happy to open an issue if you want to solve it a different way.

mantas:
Hi @abeluck ,

What would be the steps to reproduce the original issue? How did you add the custom multiselect attribute?

abeluck:
@mantas From a package by adding an ObjectAttribute like this:

```ruby
   ObjectManager::Attribute.add(
      force: true,
      object: "Ticket",
      name: "clothes_wearing",
      display: "What clothes are you wearing right now?",
      data_type: "select",
      data_option: {
        default: "",
        "options": {
          "socks": "Socks",
          "pants": "Pants",
          "shirt": "Shirt",
          "hat": "Hat"
        },
        multiple: true,
        null: true,
      },
      active: true,
      editable: true,
      position: 1620,
      created_by_id: 1,
      updated_by_id: 1,
      to_config: false,
      to_create: true,
      to_delete: false,
      to_migrate: true,
      screens: {
        "create_middle": {
          "ticket.customer": {
            "shown": true,
            "required": false,
            "item_class": "column",
          },
          "ticket.agent": {
            "shown": true,
            "required": false,
            "item_class": "column",
          },
        },
        "edit": {
          "ticket.customer": {
            "shown": true,
            "required": false,
          },
          "ticket.agent": {
            "shown": true,
            "required": false
          },
        },
      },
    )
    ObjectManager::Attribute.migration_execute
```

mantas:
@abeluck unfortunately such use of `multiple: true` on custom object attribute is not supported. There're a bunch of places around the system that would break with `multiple: true` and JSON getting stored in database. This is why `multiple` flag is not available in Admin UI.

We're aware of this limitation and it may be solved at some point. But so far we don't have it on our roadmap.

I'm very sorry, but I'll have to reject this PR.

I appreciate your effort. Please discuss potential changes by creating an issue or on community board. This way we can make sure there're no behind-the-scenes blockers. Fingers crossed this won't discourage you from future contributions!

abeluck:
Ah, well thanks for the consideration. 

I hope that multi select makes it onto the feature list soon.

Issue #2781, opened 2019-10-17
Users can't remove their Sessions from Device list


MrGeneration:
### Infos:

* Used Zammad version: 3.1
* Installation method (source, package, ..): package
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: current Chrome and Firefox

### Expected behavior:

When going into the user preferences (Devices) you can delete one or more sessions without any problem. The list will refresh and the session is gone.

### Actual behavior:

When deleting session(s) from device list as a user, sessions are not deleted.

#### Further details:

This seems to be an WebApp Issue, because the API call seems to be malformed:

`[DELETE] https://marcels-playground.zammad.com/api/v1/user_devices/undefined`
The body contains an empty `{}`

Logfile:
```
I, [2019-10-17T12:09:54.968746 #29383-74573700]  INFO -- : Started DELETE "/api/v1/user_devices/undefined" for 195.201.98.123 at 2019-10-17 12:09:54 +0200
I, [2019-10-17T12:09:54.972642 #29383-74573700]  INFO -- : Processing by UserDevicesController#destroy as JSON
I, [2019-10-17T12:09:54.972777 #29383-74573700]  INFO -- :   Parameters: {"id"=>"undefined"}
I, [2019-10-17T12:09:54.978420 #29383-74573700]  INFO -- : Completed 200 OK in 6ms (Views: 0.1ms | ActiveRecord: 0.9ms)
I, [2019-10-17T12:09:55.025724 #29383-74572400]  INFO -- : Started GET "/api/v1/user_devices?_=1571306207755" for 195.201.98.123 at 2019-10-17 12:09:55 +0200
I, [2019-10-17T12:09:55.029574 #29383-74572400]  INFO -- : Processing by UserDevicesController#index as JSON
I, [2019-10-17T12:09:55.029633 #29383-74572400]  INFO -- :   Parameters: {"_"=>"1571306207755"}
I, [2019-10-17T12:09:55.036896 #29383-74572400]  INFO -- : Completed 200 OK in 7ms (Views: 0.3ms | ActiveRecord: 1.0ms)
```

### Steps to reproduce the behavior:

* login as any user
*  go to user preferences -> devices
* try to delete any session

Yes I'm sure this is a bug and no feature request or a general question.

PR #2935, opened 2020-02-16
Fix issue with devices session remove button

bolah2009:
# Description of issue

In the `Device` page of user's `Profile`, Devices listed on this page can't be deleted when the remove button is clicked. See screenshot
devices_bug

Checking the log, I confirmed that the API call is malformed as reported in https://github.com/zammad/zammad/issues/2781

```log
I, [2020-02-16T16:06:59.341574 #14538-70174404499800]  INFO -- : Started DELETE "/api/v1/user_devices/undefined" for ::1 at 2020-02-16 16:06:59 +0100
I, [2020-02-16T16:06:59.373638 #14538-70174404499800]  INFO -- : Processing by UserDevicesController#destroy as JSON
I, [2020-02-16T16:06:59.373844 #14538-70174404499800]  INFO -- :   Parameters: {"id"=>"undefined"}
```
This issue has to do with the undefined `id` in the API call in `app/assets/javascripts/app/controllers/_profile/devices.coffee`.
The `id` is defined by targeting an anchor tag instead of a `div` tag.

```javascript
 id = $(e.target).closest('a').data('device-id')
```

This seems to be a bug after the `app/assets/javascripts/app/views/profile/devices.jst.eco` file was changed in this commit https://github.com/zammad/zammad/commit/97d14a93b3fee85769cedf4c1e7abd770798088f#diff-f150ec767074262095f89fef46309a31

The remove button was changed from the `a` element to `div` element.

### Reference to issue #2781

# Proposed solution 

### This issue was solved locally in two ways as follows:
1. Change the remove button to an anchor tag, this way the `closest()` method will be able to get the element

2. Alternatively, the `closest()` method selector could be changed to `div`, this way the method will be able to get the element

### This PR solves this issue using the second option (change `closest()` method selector to `div`)

# Testing

I couldn't replicate the issue initially on my local browser, so I had to create user devices manually from the rails console. The table below shows the comparison of deleting the devices before and after the changes

| :-----: | :-----: |
| devices_test_bug  | devices_test_fix |

### Logfile

The log file is now getting the right `id`

```

I, [2020-02-16T17:51:55.164162 #3500-69976394582120]  INFO -- : Started DELETE "/api/v1/user_devices/13" for ::1 at 2020-02-16 17:51:55 +0100
I, [2020-02-16T17:51:55.172672 #3500-69976394582120]  INFO -- : Processing by UserDevicesController#destroy as JSON
I, [2020-02-16T17:51:55.172783 #3500-69976394582120]  INFO -- :   Parameters: {"id"=>"13"}
```

### DIFF
--- a/app/assets/javascripts/app/controllers/_profile/devices.coffee
+++ b/app/assets/javascripts/app/controllers/_profile/devices.coffee
@@ -36,7 +36,7 @@ class Index extends App.ControllerSubContent
 
   delete: (e) =>
     e.preventDefault()
-    id = $(e.target).closest('a').data('device-id')
+    id = $(e.target).closest('div').data('device-id')
 
     @ajax(
       id:          'user_devices_delete'
--- a/app/assets/javascripts/app/views/profile/devices.jst.eco
+++ b/app/assets/javascripts/app/views/profile/devices.jst.eco
@@ -22,8 +22,10 @@
         <td><%= device.location %></td>
         <td><%- @humanTime(device.updated_at) %></td>
         <td class="settings-list-controls">
-          <div class="settings-list-control" href="#" data-device-id="<%- device.id %>" data-type="delete" title="<%- @Ti('Delete') %>"<% if device.current: %>disabled<% end %>><%- @Icon('trash') %></div>
+          <div class="settings-list-control" href="#" data-device-id="<%- device.id %>" data-type="delete" title="<%- @Ti('Delete') %>"<% if device.current: %>disabled<% end %>>
+            <%- @Icon('trash') %>
           </div>
+        </td>
       </tr>
     <% end %>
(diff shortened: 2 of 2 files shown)


thorsteneckel:
Great work once again @bolah2009 ! Thanks @mantas for adding the system test <3 Please note that we currently lose the author of the PR due to our GitLab -> GitHub sync :( We'll get back to you via the @bolah2009 📫

bolah2009:
> Great work once again @bolah2009 ! Thanks @mantas for adding the system test <3 Please note that we currently lose the author of the PR due to our GitLab -> GitHub sync :( We'll get back to you via the @bolah2009

Thanks, @thorsteneckel and @mantas  🙇🏾‍♂

Issue #2776, opened 2019-10-14
incomplete and curious view of the calendar


KoJie:
### Infos:

* Used Zammad version: 3.1
* hosted
* Browser + version: Chrome Version 77.0.3865.120 and Safari Version 12.1.1
* Ticket: #1080027

### Expected behavior:

* A complete calendar display where I can see all dates without the need to recount, like:
![image](https://user-images.githubusercontent.com/22768032/66763527-46211380-eea8-11e9-9506-84dee4b4dd1b.png)

### Actual behavior:

When adding an additional holiday to the Calendar panel, the calendar view is incomplete. It seems that it only shows the dates of the prior and the next month.
If you count the fields one by one and click on the relevant date, it will be entered correctly. However, editing is still difficult.

At the moment it looks like this:

![image](https://user-images.githubusercontent.com/22768032/66751266-e9652f00-ee8e-11e9-9c9b-1de70b6b68fb.png)

### Steps to reproduce the behavior:
* go to Manage --> Calendars
* Create a new calendar or take an existing one
* Click in the "Date" field in the HOLIDAYS-Section

bolah2009:
Hi @KoJie @MrGeneration @thorsteneckel 🙋🏽‍♂

Will it be okay to open a PR for this?

I think adjusting the CSS a little can yield the following screenshot below:
It is different from the picture in the **Expected behavior** above so I am not sure if this would be the best solution. 

   ![image](https://user-images.githubusercontent.com/36057474/76362211-e75daa00-6320-11ea-91bc-8b12a2b9c23d.png)

mantas:
@bolah2009 it'd be better to adjust CSS slightly more to make calendar identical to the rest of the system. As in screenshot in comment above in `Expected behavior` section.

And yes, that'd definitely make a great PR.

bolah2009:
Thanks @mantas I will work on it later today.

bolah2009:
Hi @mantas 
I opened a PR to this but am curious about the statement
>  make calendar identical to the rest of the system.

Are there other places where this calender view exist?
Sorry am new to this app and couldn't find one when I tried checking for it.

mantas:
@bolah2009 this same calendar is used in any date input field.

For example, given that you run system locally at `http://localhost:3000`, open http://localhost:3000/tests_form in your browser. Scroll down to `datetime1` field.

PR #2970, opened 2020-03-11
Fix issue with incomplete calendar view

bolah2009:
# Description of issue

In the `Admin/Manage` > `Calendar` page (http://localhost:3000/#manage/calendars), when adding an additional holiday to the Calendar panel, the calendar view is incomplete.

This issue has to do with the CSS style having the same font `color` as the `background`.

### Reference to issue #2776

# Proposed solution 

The CSS was adjusted accordingly to fit the screenshot of the Expected Behaviour in #2776

# Testing

| :-----: | :-----: |
| image  | image  |

I am not sure if this is the best approach, also I couldn't test with other parts that use the datepicker calendar view, in case the changes have affected them.

Kidly advice if there's a better approach to this solution and also parts that could be affected with these changes.

### DIFF
--- a/app/assets/stylesheets/zammad.scss
+++ b/app/assets/stylesheets/zammad.scss
@@ -9450,6 +9450,7 @@ output {
     font-size: 12px;
     padding: 5px 5px 0;
     text-align: center;
+    font-weight: bold;
   }
 
   .day {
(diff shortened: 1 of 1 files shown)


mantas:
Fixed with https://github.com/zammad/zammad/commit/4d0707a9380f4eae226a59c16dcd6e229f0184de, closing

Issue #2977, opened 2020-03-19
Resize overview fields in RTL language


YosefAdPro:
### Infos:

* Used Zammad version: 3.3
* Installation method (source, package, ..): 
* Operating system:
* Database + version:
* Elasticsearch version:
* Browser + version: Any

### Expected behavior:

*

### Actual behavior:

When using RTL language , increasing the field width it does not go in the right direction.

### Steps to reproduce the behavior:

Use RTL language (in my case Hebrew).
Go to overview> My Tickets (or any other view)
Increase the width of the field

![2020-03-19-14-18-03](https://user-images.githubusercontent.com/11810187/77067833-45a81e00-69ee-11ea-8b05-c66869588e55.gif)

Yes I'm sure this is a bug and no feature request or a general question.

PR #2979, opened 2020-03-21
Fix issue with table column resize in RTL (right to left) languages

bolah2009:
# Description of issue

When language is set to an RTL language (for example, Hebrew), adjusting the column size of any table goes to the wrong direction, that is, when intending to reduce the size it increases and when intending to increase it reduces.

### Reference to issue #2977

# Proposed solution 

The `difference` variable in the `onColResizeMousemove` method is inverted (additive) for RTL languages, that is when the `html` have the `dir` attribute to be `rtl`, the `difference` variable is calculated by multiplying `-1` to it's calculated value.

# Testing

| :-----: | :-----: |
| rtl-zammad-bug  |rtl-zammad-fixed  |

### DIFF
--- a/app/assets/javascripts/app/controllers/_application_controller_table.coffee
+++ b/app/assets/javascripts/app/controllers/_application_controller_table.coffee
@@ -934,7 +934,7 @@ class App.ControllerTable extends App.Controller
 
   onColResizeMousemove: (event) =>
     # use pixels while moving for max precision
-    if $('html[dir="rtl"]').length
+    if App.i18n.dir() is 'rtl'
       difference = @resizeStartX - event.pageX
     else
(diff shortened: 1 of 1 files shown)


mantas:
Hi @bolah2009 ,

Thank you for you contribution.

I'd like to ask for a small change. The preferred way to detect text direction is `App.i18n.dir()` rather than checking DOM tree directly.

bolah2009:
I was thinking if there is a better way too, I will do that. Thanks!

mantas:
Thanks @ bolah2009 for the PR! This has been merged into develop with the commit above.

Unfortunately we're losing the reference as you as an author. That's why we add the PR and Github user reference.

PR #2986, opened 2020-03-24
fix(coffeescript): binary vs unary operator

ma-jahn:
As we are still using v2.9 we are facing an issue with chat-no-jquery.coffee.

This was already fixed (merged) in v3.3.0.

This Pull Request simply aims to get this fix into Zammad v2. Therefore i've selected stable-2.9 as a target. Please change the target if it's not correct.

The actual issue:
`@something +'px' `
is being interpreted as unary operator and therefore converted into
`this.something('px')`

but it's expected to be an binary operator and should result in
`this.something + 'px'`


MrGeneration:
Personally I don't think it's a good idea to backport these functions.
One of the main reasons would be the security issues that are not addressed within Zammad 2.9.

Honestly, I suggest to upgrade to 3.3 as soon as possible.
https://zammad.com/news/release-3-3

I don't know how @thorsteneckel sees this though.

mantas:
Hey @ma-jahn ,

Thank you for your contribution. Unfortunately we don't have resources to actively maintain old versions. You're welcome to maintain your legacy fork though! Closing now.

PR #2987, opened 2020-03-24
fix(optional_open_button): check if elements exist before accessing p…

### LINKED ISSUE #2881: Chat button is no longer optional in no jQuery version.

ma-jahn:
Closes #2881 for Zammad v2


MrGeneration:
Just like in PR https://github.com/zammad/zammad/pull/2986 I think sticking to Zammad 2.x is a bad idea in terms of security.

mantas:
Hey @ma-jahn ,

Thank you for your contribution. Unfortunately we don't have resources to actively maintain old versions. You're welcome to maintain your legacy fork though! Closing now.

PR #3014, opened 2020-04-07
Fullquote Header for forwarded mails

drbw:
- Add a header line containung the timestamp, sender name and From address upon forwarding an email (cloned from Reply action)

=====

Considering this requirement:

- Add tests for your changes. Feel free to ask for support. We are happy to help you.

I have no idea how to do that. I found this old commit:

https://github.com/zammad/zammad/commit/2a09863df5c15ea61d3aa173e4c3da464fa5e7b7

Is that a good starting point? The action should be very similar.

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee
@@ -201,7 +201,14 @@ class EmailReply extends App.Controller
       body = App.Utils.textCleanup(article.body)
       body = App.Utils.text2html(body)
 
-    body = "<br/><div>---Begin forwarded message:---<br/><br/></div><div><blockquote type=\"cite\">#{body}</blockquote></div><div><br></div>"
+    quote_header = ''
+    if App.Config.get('ui_ticket_zoom_article_email_full_quote_header')
+      date = @date_format(article.created_at)
+      name = article.updated_by.displayName()
+      email = article.updated_by.email
+      quote_header = App.i18n.translateInline('On %s, %s <%s> wrote:', date, name, email) + '<br><br>'
+
+    body = "<br/><div>---Begin forwarded message:---<br/><br/></div><div><blockquote type=\"cite\">#{quote_header}#{body}</blockquote></div><div><br></div>"
 
     articleNew = {}
     articleNew.body = body


thorsteneckel:
@MrGeneration and @mantas can you have a look?

MrGeneration:
Mantas and I have been talking about this topic.
The PR itself looks good to go (from wording perspective).

One thing that bugged me is the email address that's supposed to be displayed there.
In my opinion, especially if it might display the agent address, this is a very bad idea.

When talking to mantas about this, he said that this should happen with quoted replies already. During this discussion we discovered that the mail address currently is not displayed - rather that code part doesn't seem to work at all.

That's not too bad in my opinion :P
Mantas is having a look. 💪

drbw:
Yes, it's not displayed when replying. (Although the variable is set.) It's just my experience of how mail clients tend to work: When replying to someone, they don't need to be told their own mail address. But someone you forward the mail to might not know it and might want to write to that address directly.

MrGeneration:
While I agree on the e-mail thing, this is a potential issue.
You don't want, at any time, to tell the recipient what the agents mail address is.

If you tell someone the agents mail address, the agent might receive service requests directly which technically is against the sense of Zammad as helpdesk software.

I do agree in the customer part, but we'll need to ensure it's just the customer address (without `ticket.agent` permissions).

But that's just my 2 cents on this topic.

mantas:
@MrGeneration even for customers, I could see a couple potential issues

Let's say an old ticket article is forwarded. Customer may have changed his name/email since ticket article was created. Which is the "true" name/email to use?

It may be confusing if sometimes email is shown and sometimes not.

MrGeneration:
Technically Zammad should use the `FROM` of the article you're forwarding.
Maybe this should be an optional setting that defaults to  `don't show mail address` ?

mantas:
I'm worried we'd run into a myriad of edge cases when email address is no longer up-to-date, not present, shall not be included because of XYZ etc.

`Don't show mail address` may easily balloon into customer/agent and/or per-group selects. What about blacklist or whitelist of emails you don't want to forward? :)

Personally I'd vote for including name + date and calling it a day. But I'm totally open to different approaches if my concerns above are too edge-case-ish.

thorsteneckel:
> Personally I'd vote for including name + date and calling it a day. But I'm totally open to different approaches if my concerns above are too edge-case-ish.

This 💪 

As you guys correctly pointed out the different mail uses cases arise when trying to replicate the behaviour of a single person mailbox with the one of a Helpdesk aka Zammad. That's why I second mantas approach 👍

mantas:
@drbw would it work for you? Or does your use case require email?

drbw:
Well, mail would be good for us (it's more unique as an identifier than a name) but not a hard requirement. And your argument about leaking PII is certainly an important issue - making that more dynamic would probably require more substantial changes.

So if your decision is that this should be removed from the commit - no problem.

mantas:
Good to hear. I'll remove it while merging. And cleanup that old quoting-on-reply implementation while at it.

martinseener:
I though now when forwarding the original authors email address is also adding to the forward header (not the agent who is forwarding). I updated to 3.4 already but the just the name there is useless.
We usually forward mails from our internal Zammad (self-hosted) to the hosted Zammad and they also need the original email address, so they can answer them directly. Is it somehow possible to have this in the forwarded email!? I really don't see an issue here. At least it would be nice to make this configurable, so everyone can decide on their own.

MrGeneration:
> Personally I'd vote for including name + date and calling it a day. But I'm totally open to different 
> approaches if my concerns above are too edge-case-ish.

I do understand your point, however, above we've decided that it's an edge case.
I personally would also say that an email address would be better during forwarding, but at least should be optional.

I'm afraid that this issue however can't solve this.

Maybe @mantas  or @thorsteneckel  wouldn't mind a hidden setting or composer setting for this?

martinseener:
Yeah, I read through this post and understand the mentioned edge-cases but to be honest, why don't adding a simple config option - even officially in the UI - that defaults to current behaviour and enabling would also show a warning about potential edge-cases or something but at least, we have an option. Would really appreciate this since currently we have to manually add these FROM mail addresses when forwarding mails, which is annoying.

thorsteneckel:
Hi @martinseener - would you mind creating a new issue and referring to this pull request? I've the feeling that we implemented things without having the requirements specified and thought through well enough. We'll update our issue template soon to use User Stories. I hope this will improve our process of finding the right solution.

Issue #3037, opened 2020-04-25
Stay on "Verifying..." when add email using imap


Chren:
### Infos:

* Used Zammad version: 3.3.x
* Installation method (source, package, ..): docker
* Operating system: centos
* Database + version: N/A
* Elasticsearch version: N/A
* Browser + version: chrome 80.0.3987.163

### Expected behavior:

* Add email via imap

### Actual behavior:
* Stay on "Verifying..." dialog

### Steps to reproduce the behavior:

* I can't post my email account here, contact me if you want the email account and configuration.

Yes I'm sure this is a bug and no feature request or a general question.
The email account is ok and I have setup on some other email clients with the same configuration.

MrGeneration:
Sorry but this sounds rather like a technical questions than a bug.
This usually occurs if Zammad can't find it's test mail it's sending within the INBOX or the subfolder you're telling Zammad to look in.

Ensure that the mail can be found by Zammad and check your STDOUT-logging of Zammad.

Chren:
I didn't receive the test mail.
How to check STDOUT-logging?

MrGeneration:
`docker logs`
etc etc.

Please consult the Community if you require technical assistance, because technical assistance is out of scope of this Github-Repository.

Chren:
Hi, thanks for response.
Here's the logs, everything seems all right, no errors.
```zammad-railsserver_1    | I, [2020-04-25T16:30:39.670819 #1-47440357926580]  INFO -- :   Parameters: {"adapter"=>"imap", "options"=>{"host"=>"imap.exmail.qq.com", "user"=>"renchonghui@bongmi.com", "password"=>"[FILTERED]", "ssl"=>true, "port"=>"993", "folder"=>"", "keep_on_server"=>true}}
zammad-railsserver_1    | I, [2020-04-25T16:30:39.678297 #1-47440357926580]  INFO -- : fetching imap (imap.exmail.qq.com/renchonghui@bongmi.com port=993,ssl=true,starttls=false,folder=INBOX,keep_on_server=true)
zammad-scheduler_1      | I, [2020-04-25T16:30:42.434584 #1-47061827615900]  INFO -- : execute Import::OTRS.diff_worker (try_count 0)...
zammad-scheduler_1      | I, [2020-04-25T16:30:42.435116 #1-47061827615900]  INFO -- : ended Import::OTRS.diff_worker took: 0.005566146 seconds.
[...]

PR #3043, opened 2020-04-28
fix: IMAP support for unicode on verification

Chren:
This is a bugfix to issue #3037.
- Rescue doesn't catch timeout error, should use "rescue Timeout::Error" instead of rescue. In my case, `message_ids = @imap.sort(['DATE'], filter, 'US-ASCII')` never returns, and rescue not called after 6 mins. Using rescue Timeout::Error ensure it would fallback to `imap.search`.
- should decode message subject when verifying email with imap. In my case, message_meta['ENVELOPE'].subject returns something like this: 
`=?gbk?B?WmFtbWFkIEdldHRpbmcgc3RhcnRlZCBUZXN0IEVtYWlsICM2MDkyMjY1ODIzNQ==?=`
 
Reply me if you want an email configuration for reproduce this issue.

### DIFF
--- a/app/models/channel/driver/imap.rb
+++ b/app/models/channel/driver/imap.rb
@@ -1,5 +1,6 @@
 # Copyright (C) 2012-2016 Zammad Foundation, http://zammad-foundation.org/
 require 'net/imap'
+require 'mail'
 
 class Channel::Driver::Imap < Channel::EmailParser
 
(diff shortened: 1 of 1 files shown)


Chren:
Anyone have a look at this pr?

MrGeneration:
Sorry this will take some time - we'll have a look as soon as possible!

Thank you very much for your time and effort so far!

mantas:
Hi @Chren 

Apologies for this taking so long!

We've fixed several issues with IMAP driver in meantime. Some of them touching same lines as your PR.

Could you test latest `develop` if your problem is still there? I see linked issue is already closed.

thorsteneckel:
Closed due to inactivity. Develop has become stable `3.5` and `3.6` in the meantime. Please let us know if it the issue still exists in current stable @Chren .

PR #3073, opened 2020-06-04
add translation portuguese Brazil: pt-br

jhenriquejrc:


mantas:
Thank you for your contribution @jhenriquejrc!

Unfortunately we loose the reference to the original author in Git commit after all is done. Author is mentioned in a commit message instead.

Closing now.

PR #3126, opened 2020-07-22
Change Opps to Oops to be more gramatically correct

asnoeyink:
While setting up Zammad I noticed that some of the error pages say Opps instead of Oops, which is probably what was meant. So I changed it. Dictionary link for reference: https://www.merriam-webster.com/dictionary/oops?show=0 :)

image


mantas:
Thank you!

Your pull request was merged and will be included in the next release.

PR #3157, opened 2020-08-19
Update chat.js and chat.min.js to fix inactiveClass option access

TheOnlySpeedy:
Currently the inactiveClass option is not access correctly.

The call is `this.inactiveClass` and should be `this.options.inactiveClass`. The error exists in chat.js and chat.min.js files.


mantas:
Hey @TheOnlySpeedy 

Thank you for your contribution!

Technical note - `chat.js` is generated off `chat.coffee` so I moved your fix there. On top of that, turns out we had the same issue in jQuery-free version so I applied your fix on `chat-no-jquery.coffee` as well.

Anyway, the fix is merged (see commit above). Closing now.

PR #3176, opened 2020-09-02
Added List-Unsubscribe tag to identify newsletters

Meloman-zz:
Added **List-Unsubscribe** filter tag to identify newsletters more easily while this is a mandatory header attached to all trusted newsletter platform.


mantas:
@Meloman-zz thank you for your contribution! Merged in with the commit above.

Issue #3080, opened 2020-06-11
facebook scope changes with APIv7.0


MrGeneration:
### Infos:

* Used Zammad version: 3.3
* Installation method (source, package, ..): any 
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: Google Chrome latest
* Ticket-ID: 1076705, 1076758 , 1076759 

### Expected behavior:

Authentication against facebook, no matter if using the "add account" functionality for adding a new facebook channel or authentication itself for logging in to Zammad, the process will function successfully.

### Actual behavior:

If enabling API version 7, facebook complains about invalid scopes which are requested by Zammad during autorisation against facebook.

Old scopes are:
* publish_pages
* manage_pages

These scopes are, as far as I can tell, replaced by:
* pages_manage_engagement
* pages_manage_metadata
* pages_manage_posts

If provided with above mentioned permissions, Zammad will function with API version 6.

![image](https://user-images.githubusercontent.com/6549061/84383926-7d89c080-abed-11ea-9c9d-5117c94d9be0.png)

### Currently working solution:

Don't upgrade facebook Apps to API version 7 (neither for endusers nor developers).
facebook Apps up to API version 6 seems to be working without issues.

*Note: Downgrading developers API version from 7 to 6 is not possible - however, it is for endusers.*

### Steps to reproduce the behavior:

* have a app with required permissions ( `pages_manage_engagement, pages_manage_metadata, pages_manage_posts, email, default` )
* try to authenticate against facebook for either logging in to Zammad or adding a facebook page to the facebook channel
  * to double tab you can test this with API version 6 and less and API version 7
  * keep in mind that, as soon you upgrade API version to v7 for the developer, there's no way back

Yes I'm sure this is a bug and no feature request or a general question.

hasbiallah:
how to downgrade the API version?

MrGeneration:
Developer API versions can't be downgraded.
Sorry, but this is out of scope of the Zammad applicaiton.

To reduce technical questions and disctraction on this issue, I'll be limitting the conversation to contributors only.
This is nothing personal against you.

Please use the Zammad community for technical questions: https://community.zammad.org
If you require commercial grade support, you can find more information on our support contracts here: https://zammad.com/pricing

PR #3256, opened 2020-10-23
#3080 changes scopes to API versions > 6

dhufnagel:
#3080 
Change OAuth scopes for FB API version 7 and above according to the changelog: [https://developers.facebook.com/docs/graph-api/changelog/version7.0](https://developers.facebook.com/docs/graph-api/changelog/version7.0)

I skipped the "pages_manage_ads" scope, as I do not think it should be needed.

### DIFF
--- a/lib/external_credential/facebook.rb
+++ b/lib/external_credential/facebook.rb
@@ -31,7 +31,7 @@ class ExternalCredential::Facebook
     {
       request_token: state,
       #authorize_url: oauth.url_for_oauth_code(permissions: 'publish_pages, manage_pages, user_posts', state: state),
-      authorize_url: oauth.url_for_oauth_code(permissions: 'publish_pages, manage_pages', state: state),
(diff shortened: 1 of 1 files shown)


thorsteneckel:
Thanks @dhufnagel ! We will address/overhaul the implementation of the Facebook integration in one of the next versions of Zammad. This PR will help 💪 We'll let you know and update the PR/Issue accordingly.

dhufnagel:
Hi @thorsteneckel . Is there any work going on to support facebook and instagram messages in near future? Is there a branch or fork where this feature is getting worked on? I am no ruby expert, but this missing feature is preventing us to use zammad in production. So I am willing to help if there is a possible way.

thorsteneckel:
Hey @dhufnagel - sorry for not keeping you in the loop.

You can find our regularly updated work queue [here](https://github.com/zammad/zammad/projects/4). As you can see there are some tasks before it in the queue that will take some time. As mentioned we update this queue regularly based on our shifting priorities. There's always the possibility that other tasks get higher prioritised because a customer has paid for the development/implementation. We can only give ETAs for paid development.

However, would love to see/support/review a contribution from you on this topic!

Issue #2842, opened 2019-12-09
Show linked public KB Answers to Customers in Ticket view


thorsteneckel:
### Infos:

* Used Zammad version: 3.2.x
* Installation method (source, package, ..): any
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: any

### Expected behavior:

As a Customer I want to be able to see linked Knowledge Base Answers in the Ticket view.

As an Admin (or Agent) I don't want to give customer access to internal KB

Linked KB answers for a Customer shall open in the public KB

### Actual behavior:

As a Customer I don't see any linked Knowledge Base Answers in the the Ticket view.

### Steps to reproduce the behavior:

* As an Agent: Create a public Knowledge Base Answer
* As an Agent: Create a Ticket for Customer X
* As an Agent: Link the Knowledge Base Answer to the Ticket
* As a Customer X: View Ticket, don't see linked Knowledge Base Answer

4-l-Clover:
Hi. thorsteneckel.
My name is Wang Jia and I got email from Sandra Kuznic that I need to solve 3 issues for apply to your company.
I installed zammad app in my local and I checked ticket problem.
I have some questions for this ticket.
Could we discuss here or can you invite me slack?
Thanks

mantas:
Hi @4-l-Clover ,

Feel free to ask directly in GH issues if the question is about the issue itself. If it's a more generic question or you want to, ask Sandra to forward your email to Thorsten or me.

4-l-Clover:
Thanks.
This is Agent Ticket Assign Page. https://prnt.sc/uzl814
And this is customer ticker view page. https://prnt.sc/uzl8g0

Could you specify where the linked KB Answer will be on?
Thanks

mantas:
This issue is talking about "Related answers" bit in sidebar. Currently it's visible to agents only. Ideally it'd show up to customers too. But show only public answers.

4-l-Clover:
Hi. mantas.
As my checking with this issue, it doesn't show KBAnswer on the sidebar when the user sign as a customer because @editTicket doesn't process @linkKbAnswerWidget in app/assets/javascripts/app/controllers/ticket_zoom/sidebar_ticket.coffee
When I add @linkKbAnswerWidget in "else" statement, it shows KBAnswer on customer's ticket sidebar.
I attached my code with screenshot.
[...]

PR #3272, opened 2020-11-09
Fixes Issue #2842 : Show linked public KB Answers to Customers in Ticket view

romitgithub:
Fixes issue https://github.com/zammad/zammad/issues/2842

The fix includes following changes:

1) Adding knowledge base reader permission to **customer** in _db/seeds/permissions.rb_ 
`customer.permission_grant('knowledge_base.reader')`

2) Changing the agent role check for rendering **Related Answers Block** to both agent & customer role

3) Adding conditional rendering for edit actions such as link and unlink KBAnswers

Please let me know if this solution works well :)

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_ticket.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_ticket.coffee
@@ -141,18 +141,20 @@ class SidebarTicket extends App.Controller
         links:       @links
       )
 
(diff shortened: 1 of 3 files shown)


mantas:
The PR looks good from technical point of view. However, we realised that it's not exactly what we're looking for in terms of functionality. It's completely on us since the original issue was a wee ambiguous.

Anyway, thank you for the contribution. Sorry for discarding otherwise fine code. Closing this PR without merging.

Issue #563, opened 2016-12-20
Sidebar - Customer - Open tickets should be clickable and open advanced search


monotek:
### Infos:
*This issue has been edited by @MrGeneration to meet our requirements.*

* Used Zammad version: Zammad 3.5 and earlier
* Installation method (source, package, ..): any
* Operating system: Any
* Database + version: any
* Browser + version: chrome
* Note: This would be helpful to merge or close all tickets of customer as soon as bulk action is implemented in advanced search https://github.com/zammad/zammad/issues/445

### Desired behavior:

When viewing the meta information of a tickets customer, clicking on "closed" or "open" tickets will provide an advanced search tab for all customer tickets fitting said situation.

Something like
`customer_id: 4 AND state: open`
`customer_id: 4 AND state: closed`

Technically it would also be usefull if it's not hard stuck to above states but would work for any state types.
This would provide *alot* more tickets because you could have e.g. different closed states or tickets in pending state you may want to cover.

### Actual behavior:

When viewing the meta information of a tickets customer, Zammad provides a popover for closed and open tickets of said customer. However, this list is incomplete and forces the agent to touch every ticket shown there manually.

![image](https://user-images.githubusercontent.com/6549061/96731642-c92c5200-13b7-11eb-9ca7-4b327bd9068e.png)

### Steps to reproduce the behavior:

* open any ticket
* switch to the customer tab
* click on the "open" or "closed" link

MrGeneration:
Hey @monotek ,

I just touched your first article, could you please have a look at my edit to see if I catched your use case correctly?
This will help us to touch the places you really meant. :)

Thanks!

romitgithub:
Hi! I'm working on this issue as a part of the interview process. I have few questions:

1) Where is the extended search page? How to open it?
2) Is there a documentation as to how to create the url for extended search such that it lists the required items

Thanks

mantas:
Hi @romitgithub 

> Where is the extended search page? How to open it?

Search for something that returns at least one result and then "Show Search Details" will appear right bellow the search input. Or go to `http://localhost:3000/#search/Welcome` given that you run your local instance at the usual address.

> Is there a documentation as to how to create the url for extended search such that it lists the required items

I guess above answer covers this one as well :)

romitgithub:
But this only accepts a query param in url. How can I pass the customer ID and state of ticket to show those tickets only?

mantas:
What if the metadata was a part of the query?

[...]

Issue #3243, opened 2020-10-08
Relational report filters do not work (ticket.organization.*, ticket.article.*, ticket.customer.*)


rolfschmidt:
### Infos:

* Used Zammad version: all
* Installation method (source, package, ..): all
* Operating system: all
* Database + version: all
* Elasticsearch version: all
* Browser + version: all
* Ticket: #1073785

### Expected behavior:

* Following filters in report profiles should work:

![grafik](https://user-images.githubusercontent.com/5445062/95457526-519b0380-0971-11eb-8cbb-2e4cb9c86755.png)

![grafik](https://user-images.githubusercontent.com/5445062/95457550-58297b00-0971-11eb-829d-1d1df7cbfb60.png)

![grafik](https://user-images.githubusercontent.com/5445062/95457565-5c559880-0971-11eb-8dae-d1392839b368.png)

### Actual behavior:

* They do not work because of the mapping which is not able to request relational ticket attributes.

### Steps to reproduce the behavior:

* Create organization with note haha
* Create customert with organization.
* Create ticket with that customer.
* Create a report profile for a organization note haha. 

![grafik](https://user-images.githubusercontent.com/5445062/95458146-13521400-0972-11eb-9f6a-051dcd77d901.png)

* Try to find the ticket

Yes I'm sure this is a bug and no feature request or a general question.

PR #3273, opened 2020-11-09
added links for open closed tickets to view extended search

romitgithub:
Fixes issue https://github.com/zammad/zammad/issues/563

### DIFF
--- a/app/assets/javascripts/app/controllers/widget/user.coffee
+++ b/app/assets/javascripts/app/controllers/widget/user.coffee
@@ -14,6 +14,9 @@ class App.WidgetUser extends App.Controller
   release: =>
     App.User.unsubscribe(@subscribeId)
 
+  getAdvancedSearchUrl: (customer_id, state_id) =>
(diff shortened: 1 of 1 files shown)


rolfschmidt:
Hi @thorsteneckel ,

i talked to @martini and i think we should postpone this one after the merge of PR issue #3243 .

The related issue fix will build up a search index with more data of the related objects. 

For example we don't like that state ids are used here. After the merge for issue #3243 we could use `state.name:open` which would be much more readable. 

It's okay for @martini to have elastic search only features because elastic search is required for production, but not in develop. So we could build in a another check so it only shows the link if the elastic search is activated (then its more understandable in develop mode without elastic search).

After the merge of the #3243 fix i will update the PR and make it ready.

rolfschmidt:
Note by @martini: future solution will open the ticket search which may result in different ticket counts between the widget and the search result because it does check permissions differently:

https://github.com/zammad/zammad/blob/be64e3ff0c263803def7bae342a79ef09c323f40/app/jobs/ticket_user_ticket_counter_job.rb#L25

This will be changed with web 2.0.

Issue #2705, opened 2019-08-16
Backup script does not work as zammad user when DB is Postgresql


abeluck:
## background

This issue is related to https://github.com/zammad/zammad/issues/2508, however while that issue is about running as root, this one is about running as the `zammad` user.

In fact it seems people are running the Zammad backup script as root in their cron jobs specifically because of this issue ([ref](https://github.com/zammad/zammad/issues/2508#issuecomment-470526180)). Monotek also noticed this ([ref](https://github.com/zammad/zammad/issues/2508#issuecomment-470527571)), however there is no open issue for it, hence this one.

## reproduction

1. Use postgresql with zammad
2. Run the backup script as the zammad user

```console
root@host:/opt/zammad/contrib/backup# su zammad
zammad@host:/opt/zammad/contrib/backup$ /opt/zammad/contrib/backup/zammad_backup.sh

# Zammad backup started - Fri Aug 16 08:08:14 UTC 2019!

creating file backup...
creating postgresql backup...
Password:  <pressed enter>
su: Authentication failure
mv: cannot stat '/tmp/20190816080814_zammad_db.psql.gz': No such file or directory

# Zammad backuped successfully - Fri Aug 16 08:12:39 UTC 2019!
```

When no tty is available during cron mode, the script fails similarly.

## the problem
The [backup_db function in the backup functions](https://github.com/zammad/zammad/blob/develop/contrib/backup/functions#L54) uses `su` to run `pg_dump`. 

So, the backup scripts cannot complete as the `zammad` user.

Also note another minor issue, that the script reports it completed successfully when in fact it didn't.

## potential solutions

1. explicitly document that when using postgresql the backup should be run as root
2. make the scripts use the provided credentials  in `../../config/database.yml` (relative to the backup script) to connect to the database

MrGeneration:
One thing beforehand:
- Please use the issue template, especially the "Info" section is mandatory 


Note while fiddling:
As I think it's enough to change the following line in your case (`contrib/backup/functions` line 54):
`su -c "pg_dump -c ${DB_NAME} | gzip > /tmp/${TIMESTAMP}_zammad_db.psql.gz" postgres`
TO
`pg_dump -c ${DB_NAME} | gzip > /tmp/${TIMESTAMP}_zammad_db.psql.gz`

In my field tests this worked.


Now I don't see this as a technical question, as this is rather a configuration issue.
The password prompt you're receiving is **not** the database, but from `su` (for changing to the postgres user).

[...]

Issue #2508, opened 2019-03-07
Backup permission issue: Could not change directory


antondollmaier:
### Infos:

* Used Zammad version: 2.9.0-1551759158.fdc9ddac.stretch
* Installation method (source, package, ..): deb
* Operating system: Debian Stretch
* Database + version: Postgresql 9.6
* Elasticsearch version: 5.6.15
* Browser + version: not relevant

### Expected behavior:

* when running `zammad_backup.sh` as root as a cronjob, no errors should be shown.

### Actual behavior:

* Cronjob: 
  ```
  30 3 * * *  root  /opt/zammad/contrib/backup/zammad_backup.sh
  ```
* Output:
  ```
  # Zammad backup started - Thu Mar  7 03:30:01 UTC 2019!
  
  adapter=postgresql dbname=zammad dbuser=zammad dbpass=xxx
  timestamp is 20190307033001
  backup dir is /var/tmp/zammad_backup
  creating file backup...
  creating postgresql backup...
  could not change directory to "/root": Permission denied
  
  # Zammad backuped successfully - Thu Mar  7 03:33:48 UTC 2019!
  ```

### Steps to reproduce the behavior:

* create cronjob for backups
* wait for cronjob to run

Yes I'm sure this is a bug and no feature request or a general question.

monotek:
Why would you want to use the root user if the zammad user can do the job?

MrGeneration:
This only applies to the root user, as @monotek  already mentioned, the root user is normally not the user you'd want the cronjob run as.

Personally I don't think this is a real issue and I also wouldn't change the script "just because root throws an error".

However: If you want to keep using roots crontab for this backup job, you could put a `cd /where/ever/you/please/; /opt.....` in front if the main command to ensure you're in the right directory.

https://unix.stackexchange.com/questions/38951/what-is-the-working-directory-when-cron-executes-a-job
Suggests the same. Of course, you can also update your own backup script.

antondollmaier:
> Why would you want to use the root user if the zammad user can do the job?

Because it can't, at least according to this:

```
root@host:~# sudo -u zammad -H /opt/zammad/contrib/backup/zammad_backup.sh

# Zammad backup started - Thu Mar  7 13:12:37 UTC 2019!

adapter=postgresql dbname=zammad dbuser=zammad dbpass=xxx
find: Failed to restore initial working directory: /root: Permission denied
timestamp is 20190307131237
backup dir is /var/tmp/zammad_backup
[...]

PR #3352, opened 2020-12-29
backup: stop using su for pg_dump

nijel:
* It creates files owned by postgres user which cannot be later moved to correct location
* It requires password in most cases what is not suitable for execution from cron

### DIFF
--- a/contrib/backup/functions
+++ b/contrib/backup/functions
@@ -51,9 +51,11 @@ function backup_db () {
     ln -sfn ${BACKUP_DIR}/${TIMESTAMP}_zammad_db.mysql.gz ${BACKUP_DIR}/latest_zammad_db.mysql.gz
   elif [ "${DB_ADAPTER}" == "postgresql" ]; then
     echo "creating postgresql backup..."
-    su -c "pg_dump --no-privileges --no-owner -c ${DB_NAME} | \
-      gzip > /tmp/${TIMESTAMP}_zammad_db.psql.gz" postgres
-    mv /tmp/${TIMESTAMP}_zammad_db.psql.gz ${BACKUP_DIR}
+    PGPASSWORD="${DB_PASS}" pg_dump \
+        --dbname "${DB_NAME}" \
+        --username "${DB_USER}" \
+        --no-privileges --no-owner --clean \
+        --compress 6 --file "${BACKUP_DIR}/${TIMESTAMP}_zammad_db.psql.gz"
     ln -sfn ${BACKUP_DIR}/${TIMESTAMP}_zammad_db.psql.gz ${BACKUP_DIR}/latest_zammad_db.psql.gz
   else
     echo "DB ADAPTER not found. if its sqlite backup is already saved in the filebackup"


MrGeneration:
Sorry but this change could break existing setups which is why I'm not accepting this change.
By default Zammad package installations do not use password based authentication which is why your approach may work or not.

The backup script you can find in `contrib/` actually can be modified as needed for own uses cases.
We had this discussion on this repository already, if you run your cron as root the issue is not existent. 
Alternatively you could allow whatever user you use for backing up to run that specific command.

nijel:
> Sorry but this change could break existing setups which is why I'm not accepting this change.

The existing code is broken when following the documentation. It creates files owned by `postgres` user which it can not moved by the process running as `zammad` user.

>  By default Zammad package installations do not use password based authentication which is why your approach may work or not.

I did the installation from package and PostgreSQL access is using password.

> The backup script you can find in contrib/ actually can be modified as needed for own uses cases.

Yes, and they will be overwritten on package upgrade, what is really lovely.

> if you run your cron as root the issue is not existent

Why does the [documentation](https://docs.zammad.org/en/latest/appendix/backup-and-restore.html) tell to run it as `zammad` user? Also running arbitrary things as root is not a good security practice. Especially if it can be easily avoided (and the script does that for MySQL).

monotek:
A pr fixing the problem would be really welcome :)

MrGeneration:
I'll be re-thinking this topic.
Please allow me *a lot* time.

Regarding your comments
> Yes, and they will be overwritten on package upgrade, what is really lovely.

That is how a package installation works, not really something we can change. If you change something in existing files it's not update safe. If you want to use contrib files, copy them some where safe...

> I did the installation from package and PostgreSQL access is using password.

Remove your password from `database.yml` and check if Zammad can still start. It should be able to. ;)

> Why does the documentation tell to run it as zammad user? Also running arbitrary things as root is not a good security practice. Especially if it can be easily avoided (and the script does that for MySQL).

Seems like you skipped the second sentence that does link to two issues that are handling these topics.
Any way, I'll be thinking it over. Please don't push this PR.

nijel:
> Remove your password from database.yml and check if Zammad can still start. It should be able to. ;)

It indeed works. In any case, I really see no reason why su to `postgres` user should be used. Anyway my patch works fine even when PostgreSQL password is not set/empty, so it should not break anything.

monotek:
I'd like to approve the change but have no time for testing at the moment.

I'm also can't remember, why i've used the su version in the first place. 
I guess there has been some reason for it.
Git blame did'nt help so far :/

MrGeneration:
I remembered. This will break existing installations - here's the reason:

```
root@zammad:/opt/zammad/contrib/backup# ./zammad_backup.sh

# Zammad backup started - Wed Jan  6 10:26:09 CET 2021!

find: ‘/tmp/zammad_backup/*_zammad_*.gz’: No such file or directory
creating file backup...
creating postgresql backup...
pg_dump: [archiver (db)] connection to database "zammad" failed: FATAL:  role "root" does not exist

# Zammad backuped successfully - Wed Jan  6 10:26:30 CET 2021!

root@zammad:/opt/zammad/contrib/backup# su - zammad
zammad@zammad:~$ /opt/zammad/contrib/backup/zammad_backup.sh

# Zammad backup started - Wed Jan  6 10:26:49 CET 2021!

creating file backup...
creating postgresql backup...

# Zammad backuped successfully - Wed Jan  6 10:27:12 CET 2021!

zammad@zammad:~$
```

There are users using `root` for backing up their Zammad installation.
The suggested changes will break default installation with root-user backup.

Beside of that ... 
Your PR only works because the `zammad` user by default is allowed to use socket connections.

This line is clustered and partly non functional:
 https://github.com/zammad/zammad/pull/3352/files#diff-88844eaa4b6b3f848d8255dc85dbda46c59f389882049277189fb0a2b99c64afR54

nijel:
> This line is clustered and partly non functional:

Indeed, it was missing user parameter, sorry for that.

Anyway the patch certainly could be improved, the question is whether you are interested in this direction.

MrGeneration:
This will take time on my end for fiddling that approach to an end.

monotek:
Why are root users an issue?
In the documentation we advise to use the zammad user: 
https://docs.zammad.org/en/latest/appendix/backup-and-restore.html?highlight=backup

MrGeneration:
> Why are root users an issue?
Because there are users / systems that are using that approach.

> In the documentation we advise to use the zammad user:
We both know how users and documentations work. :-)

Beside of that we also had similar discussions on the following (linked) issues:
* https://github.com/zammad/zammad/issues/2508
* https://github.com/zammad/zammad/issues/2705

Again, I'll have a look at the approach here.
As it needs work it will take time from my end.

monotek:
> Because there are users / systems that are using that approach.

Why should we care for users not following the documentation?
Imho if somebody changes stuff he should also take care of the changed setup completely.

I've tested the change. 
Backup with zammad user and restore with root user works for me.

MrGeneration:
> Backup with zammad user and restore with root user works for me.

Really? Doesn't work on my vanilla installation without any administrative changes:
`pg_dump: [archiver (db)] connection to database "zammad" failed: FATAL:  Peer authentication failed for user "zammad"`

My machine in that case is a Debian Buster minimal installation.
Installation is 1 by 1 followed the Zammad documentation.

---

The reason I do care about those installations that I'm trying to keep impacts as low as possible.
Beside of that the support requests nuke my desk - I'd like to avoid that where I can.

---

Why do we use
```
   PGPASSWORD="${DB_PASS}" pg_dump \
        --dbname "${DB_NAME}" \
        --username "${DB_USER}" \
        --no-privileges --no-owner --clean \
        --compress 6 --file "${BACKUP_DIR}/${TIMESTAMP}_zammad_db.psql.gz"
```

Above approach is not recommended according to postgresql documentation:
http://www.postgresql.org/docs/current/static/libpq-envars.html

A not fully tested (yet) approach could look like the following.
This also ensures we don't receive the next PR within the next days to improve
```
    cat <<EOT >> ~/.pgpass
*:*:${DB_NAME}:${DB_USER}:${DB_PASS}
EOT
    chmod 600 ~/.pgpass
    pg_dump \
        --dbname "${DB_NAME}" \
        --username "${DB_USER}" \
        -h localhost \
        --no-privileges --no-owner --clean \
        --compress 6 --file "${BACKUP_DIR}/${TIMESTAMP}_zammad_db.psql.gz"
    rm ~/.pgpass
```

Above is the fastest and dirties approach I could find.

I did not verify
* that restoration is working
* the dump is what we expect and need

I did verify
* the user running the command does not matter

Possible improvements beside testing (also for docker...)
* also check for the database servers hostname in case it's not localhost


What do you think? ^^

MrGeneration:
Any thoughts on my above suggestion? ^^

nijel:
Changing the user configuration is something that can go easily wrong (what if the file exists, what if it contains wrong credentials for the same service, ...). You better use `PGPASSFILE` with a temporary file for that.

Anyway, passing the password in the environment is safe on Linux, you will not get it with `ps`. I have no clue if Zammad supports running on other systems which might do this.

On the other side, what certainly is not safe is passing it on a command line, and that is what you do with MySQL backups ;-).

MrGeneration:
> Changing the user configuration is something that can go easily wrong (what if the file exists, what if it contains wrong credentials for the same service, ...). You better use PGPASSFILE with a temporary file for that.

I already said: This method is not recommended by postgresql developers.
My suggestion does create a temporary `.pgpass` file to do it the supported way.

I don't see any part where we do change configurations on the instance host directly.
If you mean that we should check for that file before hand, I don't see why we shouldn't.

I may just not get what you want to tell me correctly.

> On the other side, what certainly is not safe is passing it on a command line, and that is what you do with MySQL backups ;-).

Let's please focus on the topic. ;-(

nijel:
> I already said: This method is not recommended by postgresql developers.

Yes, it is not recommended because on some systems the environment can be listed by an unprivileged user. This is not the case on Linux, so the recommendation might not be relevant for Zammad (though I have no clue on what operating systems to you support). Any other solution will be more complex, so you should evaluate whether introducing the complexity is reasonable.

>  My suggestion does create a temporary `.pgpass` file to do it the supported way.

Your code does append to `~/.pgpass` and removes it after the backup. In case the file was existing and was used for some other things it would be gone now.

Something like this would avoid that:

```
export PGPASSFILE=`mktemp`
chmod 600 "$PGPASSFILE"
cat <<EOT >> "$PGPASSFILE"
*:*:${DB_NAME}:${DB_USER}:${DB_PASS}
EOT
pg_dump ....
rm "$PGPASSFILE"
```

wucherpfennig:
Hi guys
we would very much appreciate if you could update the backup script to use the db password if it is available.

Reason: there are other systems who might access a zammad database too (like metabase) and the require passwords...

The initially proposed update from @nijel (ENV Vars) works fine.

BR wucherpfennig

mgruner:
@nijel I tried to consolidate all changes of this and another related PR in #3621. This could be the basis for a merge if there is a consensus. 
Would you be so kind to review the changes there and let me know if they work for you?

PR #3364, opened 2021-01-12
Maintenance: Dont try to modify possibly frozen string in place

e-tobi:
When trying to use Zammad with ruby 2.7.2 I ran into a FrozenError. Class names are frozen, so they can't be modified in place by gsub!(). Even if Gemfile uses ruby 2.6.6 which does not have this issue, it might be a good idea to fix this.

### DIFF
--- a/app/models/channel/email_parser.rb
+++ b/app/models/channel/email_parser.rb
@@ -740,7 +740,7 @@ process unprocessable_mails (tmp/unprocessable_mail/*.eml) again
     end
 
     # for some broken sm mail clients (X-MimeOLE: Produced By Microsoft Exchange V6.5)
-    filename ||= file.header[:content_location].to_s.force_encoding('utf-8')
+    filename ||= file.header[:content_location].to_s.dup.force_encoding('utf-8')
 
     # generate file name based on content-id
     if filename.blank? && headers_store['Content-ID'].present? && headers_store['Content-ID'] =~ /(.+?)@.+?/i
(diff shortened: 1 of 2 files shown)


MrGeneration:
Do you mind sharing which code base you used during your tests with different ruby versions?

e-tobi:
I'm using the most recent version of the develop branch of the git repository. With this change it seems to run fine with ruby 2.7.2.

self.class.name.frozen? -> true in 2.7.2, false in 2.5.5 - I haven't tried 2.6.6

thorsteneckel:
Hey @e-tobi - thanks for the heads up. We've started an internal test branch to run the entire test suite with ruby 2.7.2. However, there are some setup issues we need to resolve (as it turned out) before I can give you feedback on possible other issues. Anyhow, thanks for moving this one out of the way 🚀

e-tobi:
Ok, great! I'm looking forward to see a fully tested version for ruby 2.7.2. I've tried to run the test suite myself, but it seems to be hard to set up. The unit tests seem to be buggy. I get failures because of upper/lower case issues, asserts seem to have expected and actual values beeing swapped in some cases and so on. Are these tests still maintained? CI doesn't seem to run them as well.

The specs partially run, but I can't get the selenium specs to work with geckodriver - this is not my comfort zone :-)

e-tobi:
Found another FrozenString issue when running the specs, PR updated - see: 7c8c9f4

Right now I only have 7 failing specs with ruby 2.7.2 - I will look into these as soon as I find the time for it.

thorsteneckel:
Hi @e-tobi - thanks again for your contribution. We're happy to say that with the version after next (so 4.2) Zammad will run on latest version of Ruby 2.7. We've resolved these and other changes on the way (thanks @mgruner !). There are still some tests we need to perform but it currently looks good. I'll close this issue now.

Issue #330, opened 2016-10-29
WYSIWYG Editor 


MDXDave:
The example triggers includes bold text, but it is not possible to style your own text (except this workaround)
![zammad2](https://cloud.githubusercontent.com/assets/6099976/19830729/10704f24-9dfe-11e6-98a0-d173b0d614a6.gif)

How about include a _simple_ WYSIWYG editor including bold, italic, underlinde, font size and images.

martini:
Note: It is also requested in T#1082152


Hi @MDXDave 

did you see the ctrl+alt+h shortcuts? Would this be fine for simple WYSIWYG editor?

<img width="739" alt="screen shot 2016-10-29 at 18 27 14" src="https://cloud.githubusercontent.com/assets/110226/19831248/68389c5a-9e05-11e6-8961-8de61ea7bfe1.png">

MDXDave:
Oh i have missed this. That is quite enough.

But maybe it would be easier to use with a toolbar like here in github?

martini:
Yeah, currently it's intend to keep the UI clean as possible - but if you miss features like this... I'll discuss it with our designers.

PS: With the short cuts you can also close tabs, it's really nice to increase efficiency.

Julian0o:
The shortcut system is nice! No need for me to have a WYSIWYG editor.

MartinBorg:
New users either need some kind of buttons to change style (because they do not even know about short cuts or can't remember them).

Or there should be Zammad coffee mugs with editor shortcuts. Problem: I'm an espresso fanatic, so there is space only for one or two shortcuts.

Btw. I like the way Trac does editing with easy changing between their wiki markup editing and WYSIWYG. Includes tables etc.

AND51:
Hello,

I would like to share my opinion with you. I know the shortcut sheet cheat (ctrl+alt+h). Rarely I use a shortcut to make text **bold**, so yes, I know this feature. But you mustn't forget that one cannot keep everything in mind. I also work with Zendesk every day and even if there is a RichText editor, I only use 2 shortcuts regularly (insert code tags and creating an bullet list). Everything else like bold, italic, insert a table and so on is being done so rare that I can simply not remember the corresponding shortcut. Please, compare this screenshot I took from Zendesk:

![screenshot-2017-04-16-000310](https://cloud.githubusercontent.com/assets/26226303/25067081/faa0b85e-2238-11e7-9077-ef254e54ea15.png)

[...]

PR #3367, opened 2021-01-14
Insert preformated text

beckerr-rzht:
This commit extends the built-in richtext editor with the function to insert **preformatted text**.
This can be used for example to insert **code blocks**.
The shortcut is `Stg`+`Shift`+`,`.  

This is discussed here https://community.zammad.org/t/format-as-code-block/2630
and is related to #330

This feature has been tested with Firefox and Chrome.


thorsteneckel:
Hi @beckerr-rzht - thanks for your contribution and sorry for the late reply. Unfortunately I have to tell you that we're currently not taking any contributions or make any changes to the keyboard shortcut feature. That's because of issue #656 which is about broken keyboard shortcuts (existing ones) in different browsers. Long story short: It's really really really hard to have keyboard shortcuts that work across different browsers and operating systems without colliding with OS or browser shortcuts. So there's probably an OS and / or browser combination out there for which your change doesn't work.

We're planning to rework the implementation in a way to work around this. However, we can't tell when this will be the case. Now that the issue and your PR are linked (because of the reference above). I will keep you updated once we're working on the issue.

Therefore I'll close this PR now. Thanks for your understanding.

PR #3386, opened 2021-01-28
Settings: Add Option for `mark and quote` feature

m-ildefons:
- Add option to settings to enable/disable `mark and quote` feature
- Default: enabled (current behaviour)

This options allows to turn off the `mark and quote` feature. This is
desirable in some cases, because while convenient it may lead to agents
accidentally leaking data to customers, when not intended.

This feature was requested here:
https://community.zammad.org/t/zammad-shouldnt-automatically-quote-selected-text/4651/3


m-ildefons:
@martini, may I ask for a review on this?

mantas:
Hi @m-ildefons

Unfortunately we decided to reject this PR. It looks great from technical perspective. But this functionality would add too much complexity from usability point of view.

We'd like to improve quoting/replying workflow instead. As a first step, we'll limit quoting to selections inside of an article. Let's discuss further ideas in community thread linked above.

PR #3398, opened 2021-02-08
update license to original text

Th3Ph4nt0m:
As can be easily seen, the licence from GitHub is not summarised. This is because the original text of the licence is not used. In this PR, I have replaced the licence text with the original one so that the licence is recognised by GitHub.


thorsteneckel:
Thanks for your contribution @Th3Ph4nt0m ! Merged with the commit above.

PR #3404, opened 2021-02-11
Fixes Issue#2629

### LINKED ISSUE #2629: Full quote shows "<agent> wrote:" instead of "<customer> wrote" when agent toggled article visibility
### Infos:

* Used Zammad version: 3.0.x
* Installation method (source, package, ..): docker-compose
* Operating system: Debian 9
* Database + version: postgres
* Elasticsearch version: doesn't matter
* Browser + version: Chrome
* Ticket-ID: #1060389

### Expected behavior:

* When an agent toggles "set to internal"/"set to public" of an article of a ticket (for example accidentally), and then another agent (or the same one, doesn't matter) replies to this article, the full quote should be introduced with "On Wednesday, June 26, 2019, 8:57:55 AM, <Customer's name> wrote:".

### Actual behavior:

* When an agent toggles "set to internal"/"set to public" of an article of a ticket (for example accidentally), and then another agent (or the same one, doesn't matter) replies to this article, the full quote is introduced with "On Wednesday, June 26, 2019, 8:57:55 AM, <Agent's name> wrote:".

* As a customer, create a ticket (for example by sending a mail into Zammad). Or as an agent, create it manually as a phone ticket and choose a customer.
* Ensure that the options "Email - full quote" and "Email - quote header" are both enabled.
* As an agent, click on reply. Note that the full quote is introduced as "<Customer> wrote:", as it should be. Discard your changes.
* As an agent, toggle the visibility of the article by clicking on "set to internal"/"set to public".
* As an agent, click on reply again. Note that the full quote is now introduced as "<Your agent name> wrote:", which is wrong.

### Further analysis

The bug is caused by:

1. [app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee](https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee#L151..L157) uses `article.updated_by` to determine the name. We don't use the customer frontend, but I suspect customers may edit their own ticket articles, and therefore it could be relevant which customer edited the article last?

1. When an agent toggles the visibility of an article, _he_ becomes the user who last updated the article.

romitgithub:
Fixes https://github.com/zammad/zammad/issues/2629

Added origin_by_id attribute to ticket_article.cooffee to have origin user data in article view.
And changed quote text to (origin_by || created_by).displayName to show correct user name when replying.

### DIFF
--- a/spec/system/ticket/update/full_quote_header_spec.rb
+++ b/spec/system/ticket/update/full_quote_header_spec.rb
@@ -38,6 +38,17 @@ RSpec.describe 'Ticket > Update > Full Quote Header', type: :system, time_zone: 
       end
     end
 
+    it 'includes OP when article visibility toggled' do
+      within(:active_content) do
+        set_internal
+        highlight_and_click_reply
+
+        within(:richtext) do
+          expect(page).to contain_full_quote(ticket_article).formatted_for(:reply)
+        end
+      end
+    end
+
     context 'when customer is agent' do
       let(:customer) { create(:agent) }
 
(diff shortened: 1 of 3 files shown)


spec/system/ticket/update/full_quote_header_spec.rb
```diff
@@ -162,7 +196,7 @@ def citation
     end
 
     def name
-      expected.created_by.fullname
+      expected.origin_by.fullname || expected.created_by.fullname
```
mantas: Are you sure this won't crash of `origin_by` is `nil`? :)

romitgithub: Yep, added a null check for that.

spec/system/ticket/update/full_quote_header_spec.rb
```diff
@@ -162,7 +196,10 @@ def citation
     end
 
     def name
-      expected.created_by.fullname
+      if expected.origin_by?
+        expected.origin_by.fullname
+      else
+        expected.created_by.fullname
```
mantas: Something may be missing here :)
 
While at it, it may great to do this in minimalist single-line style. More clear business logic and simplier syntax.

romitgithub: Ah alright! Update it now.

PR #3408, opened 2021-02-13
Fix issue with disapearing treeselect that has backslash

### LINKED ISSUE #3162: Backslash "\" causes tree select to disappear during selection
### Infos:

*This bug report originally comes from community user "vak" (https://community.zammad.org/t/bug-report-version-3-4-0-4/5152)**

* Used Zammad version: 3.4
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: verified on latest Google Chrome

### Expected behavior:

The content of a tree select field doesn't matter. You can always select it's value. If you configure unsupported characters into one of your values, Zammad won't let you update the object.

### Actual behavior:

Entering "\" into any value of a tree select object causes the value selection to disappear as soon as you use such a value.

In my quick test I verified it on:
* overviews
* sla
* ticket detail view
* trigger

I guess it affects the complete UI.

* create a tree select with "\" in at least one of your values
* update your database and restart Zammad
* go to any UI part that allows you to select tree select field values and try to set your value with backslash (or a level below it).

issue_backslash-in-tree-selects

bolah2009:
# Description of issue

When a tree select has values with backtick `\`,  it gets impossible to navigate to such element children. This is because the query used to select the element to escape the backtick, so for example, when `le \ test` is the parent it selects `le  test` which would not be found.

### Reference to issue #3162

Fixes #3162

## Proposed solution 

The backtick is escaped using the existing escape utility `App.Utils.escapeRegExp` method. So when backtick or other special characters are used, the method escape this character so the query that selects the element use the character, the character keeps all it's value. Now, when `le \ test` is the value used the method escape the bactick by returning `le \\ test` and then the query uses `le \ test`.

## Screenshots

### Before changes

https://user-images.githubusercontent.com/36057474/107861871-ff02b400-6e48-11eb-8a87-7a580242ee5a.mov

### After Changes

https://user-images.githubusercontent.com/36057474/107860297-e2fa1500-6e3e-11eb-92c7-96002fc9379b.mov

### DIFF
--- a/app/assets/javascripts/app/lib/app_post/searchable_select.coffee
+++ b/app/assets/javascripts/app/lib/app_post/searchable_select.coffee
@@ -249,6 +249,7 @@ class App.SearchableSelect extends Spine.Controller
     return if @animating
     if dir > 0
       target = @currentItem.attr('data-value')
+      target = App.Utils.escapeRegExp(target)
       target_menu = @optionsSubmenu.filter("[data-parent-value=\"#{target}\"]")
     else
       target_menu = @findMenuContainingValue(@currentMenu.attr('data-parent-value'))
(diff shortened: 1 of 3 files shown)


mantas:
Merged with https://github.com/zammad/zammad/commit/4c860150744459f64e70d842f112d75e0bf1973f

PR #3423, opened 2021-02-24
Fixes #3012: Don't display "article type" selection for customers within UI.

### LINKED ISSUE #3012: Don't display "article type" selection for customers within UI
### Infos:

*This originally comes from the community (https://community.zammad.org/t/customer-note-button-next-to-text-response-doesnt-seem-to-do-anything/3821) and was suggested by "RobOwen1990"*

* Used Zammad version: 3.3
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: any Browser affected

### Desired behavior:

Don't display UI parts if a user can't use it.
This affects the article type for customers within the Ticket detail UI.

### Actual behavior:

Currently customers can only post web notes which is why it's also the only type they can select.
Zammad shouldn't provide the feeling you can select something else when you just get one option to choose from.

### Screencast:
dont-display-article-selection-for-customers

* Login as customer
* If the customer account doesn't have a ticket yet create one
* Click into the new article text form to start creating a new article and note the article type selection on the left side

dvuckovic:
Fixes #3012.

Article type selection button next to the article text input will be hidden in case there are no alternative choices.

### DIFF
--- a/app/assets/javascripts/app/views/ticket_zoom/article_new.jst.eco
+++ b/app/assets/javascripts/app/views/ticket_zoom/article_new.jst.eco
@@ -7,6 +7,7 @@
   <div class="editControls">
     <div class="js-avatar"></div>
     <div class="editControls-item is-hidden pop-select">
+      <% if @articleTypes.length > 1: %>
       <div class="editControls-iconHolder centered js-selectedArticleType">
         <% for articleType in @articleTypes: %>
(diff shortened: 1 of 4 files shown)


thorsteneckel:
Would you mind taking over here as well? You can also introduce @romitgithub or @mgruner how to handle GitHub PRs. @bolah2009 could also help out.

mantas:
Hi @dvuckovic

Your changes are now merged with the following commit: https://github.com/zammad/zammad/commit/2a61be02be6aee0f11c0414feb651d6d216c398d

Thank you for your contribution!

PR #3429, opened 2021-02-25
Fixes #3018: Times in ticket history are displayed in UTC.

### LINKED ISSUE #3018: Times in ticket history are displayed in UTC
### Infos:

*This enhancement request has been edited by @MrGeneration to meet our current requirenments. This originally comes from the discussion on the Community https://community.zammad.org/t/ticket-history-time-zone/4256 .*

* Used Zammad version: 3.3.0
* Installation method (source, package, ..): deb
* Operating system: ubuntu 18.04
* Database + version: postgre
* Elasticsearch version: 7.6
* Browser + version: any

### Desired behavior:

When opening the ticket history, the user will see all dates within his timezone.
This helps the user to get timestamps faster (and reduces the need to calculate from UTC to the actual timezone the user lifes in).

Zammad should parse all times and dates to the users timezone, because if one part in the UI doesn't, the user doesn't expect it (even though it tells you a value is in e.g. "UTC").

### Actual behavior:

Zammad shows the UTC time/date within the ticket history.
It doesn't further parse database content.

image

### Steps to reduce the behaviour

* Create a ticket
* fiddle around a bit (create articles, update the ticket)
* check the ticket history

### Scope:

This affects all time values within the history. So not just creation dates, but also e.g. pending till dates.

dvuckovic:
Fixes #3018.

Since the ticket history data does not contain type information for its values, we attempt to parse them in an oportune way and consider it a time stamp only in case it produces a valid date object. Later, these date objects can be localized in user's timezone.

Test cases have been added both for the new util method to parse ISO dates (QUnit), and ticket history modal integration (RSpec/Capybara).

### DIFF
--- a/app/assets/javascripts/app/controllers/_application_controller/generic_history.coffee
+++ b/app/assets/javascripts/app/controllers/_application_controller/generic_history.coffee
@@ -82,15 +82,17 @@ class App.GenericHistory extends App.ControllerModal
         content = "#{ @T( 'This ticket was merged into' ) } #{ @T( 'ticket' ) } #{ ticket_link }"
       else
         content = "#{ @T( item.type ) } #{ @T(item.object) } "
+
+        # convert time stamps
+        if item.value_from
+          dateObject = App.Utils.parseISO(item.value_from)
+          item.value_from = App.i18n.translateTimestamp(dateObject) if dateObject
+        if item.value_to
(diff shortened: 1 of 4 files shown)


dvuckovic:
#3018 already closed, this PR is now deprecated.

PR #3446, opened 2021-03-10
Allow single tenant only application in microsoft 365 

bmanojlovic:
In given PR you can find code that enables per tenant authorization and email integration and not just by
/common/ code path (multi-tenant azure applications) , I hope it can be used as is or what is expected to be added (not a ruby programmer so sorry for code quality)

### DIFF
--- a/app/assets/javascripts/app/views/microsoft365/app_config.jst.eco
+++ b/app/assets/javascripts/app/views/microsoft365/app_config.jst.eco
@@ -20,6 +20,14 @@
       <input id="client_secret" type="text" name="client_secret" value="<% if @external_credential && @external_credential.credentials: %><%= @external_credential.credentials.client_secret %><% end %>" class="form-control" required autocomplete="off" >
     </div>
   </div>
+  <div class="input form-group">
+    <div class="formGroup-label">
+      <label for="client_tenant">Tenant UUID/Name<span>*</span></label>
+    </div>
+    <div class="controls">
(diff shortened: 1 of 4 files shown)


thorsteneckel:
Thanks @bmanojlovic ! We will have a look and merge if everything works as expected. Would you mind adding an issue with the pre-defined templated filled out for reasoning purposes?

@mantas can you please double check and ensure a clean migration path for existing installations?

bmanojlovic:
at this moment it does not do  "defaulting" to /common/ path - that must be implemented and in that case all existing users should have no issues with, and that is better question how i should implement it, in "application" code or in database/migration code.

mantas:
@bmanojlovic I'm already on it :)

IMO the best way is to default to `common` if there's no optional tenant present. This way existing systems keep working. And new systems won't require entering `common` either.

However, I'm not familiar with Microsoft/Azure ecosystem at all. But it looks like `common` is the usual option for most installations. Thus making it optional seems to be a sensible default?

bmanojlovic:
Definitely optional, as that if someone needs functionality can enable/change from default "common" to its own tenant id/domain name.

thorsteneckel:
Closed via https://github.com/zammad/zammad/commit/85d6e198926c4b7ac8f4cf91edc1d200f26f9dc9 - thanks for your contribution! It will be available from Zammad 4.1 on.

Issue #1601, opened 2017-11-01
Make it possible to run Zammad stateless


monotek:
When running Zammad in container environments, with single container for railsserver, websocket and scheduler, we need shared storage as tmp dir needs to be written and read by all containers. This is a problem when using Zammad in Kubernetes as only one container can mount a shared directory writeable, if PV is ReadWriteOnce.

As discussed earlier with @martini it would be possible to use memcached to store all tmp files. Rails already supports writing the cache to memcached ( http://guides.rubyonrails.org/v3.2/caching_with_rails.html#activesupport-cache-memcachestore ).

At least the websockets server would need some modification to also use memcached.
Check also if scheduler needs modifications.

With this option one would not need to create shared storage solution like NFS, Ceph, GlusterFS and so on inside Kubernetes and it would be possible to just spin up an additional memcached container, which would be much easier.

### Infos:

* Used Zammad version: 2.2
* Used Zammad installation source: deb
* Operating system: ubuntu 16.04
* Browser + version: all

### Expected behavior:

* Its possible to use memcached as cache
* Processes do not create extra files, like in:
  * public/assests/images

### Actual behavior:

* Zammad uses tmp dir to store temporary files
* its not possible to change that at the moment
* some ner files are created in public/assests/images

### Steps to reproduce the behavior:

* run zammad

mgruner:
Running Zammad in a clustered environment will be much easier [with Zammad 5](https://github.com/zammad/zammad-documentation/pull/190):

- Redis is [now supported](https://github.com/zammad/zammad/commit/1eef2a0e0f5a32b7769590749aa14b512de25d00) as web socket session store
- Memcached [can now be configured](https://github.com/zammad/zammad/commit/f05f8e2d3cc1b4a91bede6c0d1452dd21f765247) as Rails cache back end

The following areas are also relevant for this topic, but don't require changes at present:

- Zammad caches the logo as an asset in public/. This can happen on every note separately and requires no shared file system.
- The store for uploads and attachments can be configured to use the database to be shared between the nodes. Adding other store options could be a future optimization, but is not required for stateless operation per se.
- Unprocessed / oversized emails are stored in a folder on the node that does this processing. At this point we believe that this is not mission-critical data that needs special treatment.
- The package manager should either not be used (only vanilla Zammad), or alternatively package installations should be performed as part of the deployment process, distributing the same codebase to all nodes.

[...]

PR #3450, opened 2021-03-16
Add redis as an (optional) session store backend

djmaze:
As [discussed on the community forum](https://community.zammad.org/t/overviews-and-notifications-not-pushed-updated-via-websocket/5854), it is currently a bit of a struggle to make a Zammad work on a multi-machine cluster. The main problem is the use of a shared directory as a kind of message queue for the sessions. Sharing that directory via a distributed filesystem turns out to be really problematic.

A much more suitable tool for this kind of work is Redis. I thus refactored the sessions module to allow for different store backends and implemented a Redis backend in addition to the file backend. 

In my local tests, this seems to work well. I still have to try this out in production though.

I am a bit unsure with the naming (is `store` the right term to use for the backends?). Also, I did not yet have a look at the tests in order to see if or how the new backend can be tested.

In order to use this, you need to set the env variable `REDIS_URL` to a running Redis server for the Zammad services.

Let me know what you think of this.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -40,6 +40,9 @@ module Zammad
     # define cache store
     config.cache_store = :file_store, Rails.root.join('tmp', "cache_file_store_#{Rails.env}")
 
+    # define websocket session store
+    config.websocket_session_store = ENV['REDIS_URL'] ? :redis : :file
+
     # default preferences by permission
     config.preferences_default_by_permission = {
       'ticket.agent' => {
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem 'delayed_job_active_record'
 # core - websocket
 gem 'em-websocket'
 gem 'eventmachine'
+gem 'redis'
 
 # core - password security
 gem 'argon2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -434,6 +434,7 @@ GEM
     rb-inotify (0.10.0)
       ffi (~> 1.0)
     rchardet (1.8.0)
+    redis (4.2.5)
     regexp_parser (1.8.2)
     rest-client (2.0.2)
       http-cookie (>= 1.0.2, < 2.0)
@@ -660,6 +661,7 @@ DEPENDENCIES
   rails-observers
   rb-fsevent
   rchardet (>= 1.8.0)
+  redis
   rspec-rails
   rszr (= 0.5.2)
   rubocop
--- a/lib/sessions.rb
+++ b/lib/sessions.rb
@@ -1,13 +1,9 @@
 module Sessions
 
-  # get application root directory
-  @root = Dir.pwd.to_s
-  if @root.blank? || @root == '/'
-    @root = Rails.root
+  @store = case Rails.application.config.websocket_session_store
+    when :redis then Sessions::Store::Redis.new
+    else Sessions::Store::File.new 
   end
-
-  # get working directories
-  @path = "#{@root}/tmp/websocket_#{Rails.env}"
 
   # create global vars for threads
   @@client_threads = {} # rubocop:disable Style/ClassVars
@@ -25,10 +21,6 @@ returns
 =end
 
   def self.create(client_id, session, meta)
-    path         = "#{@path}/#{client_id}"
-    path_tmp     = "#{@path}/tmp/#{client_id}"
-    session_file = "#{path_tmp}/session"
-
     # collect session data
     meta[:last_ping] = Time.now.utc.to_i
     data = {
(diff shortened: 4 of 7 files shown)


djmaze:
As Github does not want to finish running the tests, I did it all locally. Both with and without a `REDIS_URL`. In both cases, the following tests failed on my machine:

```
rspec ./spec/requests/integration/github_spec.rb:91 # GitHub request handling does save ticket issues
rspec ./spec/requests/integration/gitlab_spec.rb:96 # GitLab request handling does save ticket issues
rspec ./spec/requests/proxy_spec.rb:41 # Manage > Settings > System > Network request handling when proxy settings uses invalid config with unknown proxy
rspec ./spec/requests/proxy_spec.rb:51 # Manage > Settings > System > Network request handling when proxy settings uses invalid config with invalid proxy username
rspec ./spec/requests/proxy_spec.rb:61 # Manage > Settings > System > Network request handling when proxy settings uses invalid config with invalid proxy password
```

The detailed error messages look like they are not related to my changes but maybe to my development setup. (As I was not able to find any documentation on how to setup a development environment, there might be something missing.)

MrGeneration:
Hey,

maybe this can help you a bit further:
https://docs.zammad.org/en/latest/contributing/ci.html#local

Documentation on this regard currently lacks a bit. Sorry! :-(

djmaze:
Well, that seems to be outdated. I took the commands out of *.github/ci.sh* since that is what the CI is currently doing.

MrGeneration:
> Well, that seems to be outdated. I took the commands out of .github/ci.sh since that is what the CI is currently doing.

Not exactly, it's quite fresh actually. ;-x
https://github.com/zammad/zammad-documentation/pull/163

thorsteneckel:
Hi @djmaze - I'm stunned! Fantastic to receive a PR for this. We're currently right before the 4.0 release so I don't find much time for other things. I just wanted to let you know that these were all tests we added / changed over the last week(s). We haven't had the chance to see the impact on PRs. Looks like there's something broken on our side. Please ignore them for now.
I'll get back to you with more detailed feedback, solutions and a review as fast as I can. This might take 1 or 2 weeks though.

monotek:
Linking for reference: https://github.com/zammad/zammad/issues/1601

config/application.rb
```diff
@@ -40,6 +40,9 @@ class Application < Rails::Application
     # define cache store
     config.cache_store = :file_store, Rails.root.join('tmp', "cache_file_store_#{Rails.env}")
```
monotek, line 41: In container context i've used memcached for cache_store.
See: https://github.com/zammad/zammad-docker-compose/blob/master/containers/zammad/docker-entrypoint.sh#L58

Could we use redis here too to have a single solution for all file based caches and sessions stores?

monotek, line 41: In container context i've used memcached for cache_store.
See: https://github.com/zammad/zammad-docker-compose/blob/master/containers/zammad/docker-entrypoint.sh#L58

Could we use redis here too to have a single solution for all file based caches and sessions stores?

thorsteneckel: Redis is actually our preferred long term solution 👍

djmaze: From my point of view, yes, that would make sense.

djmaze:
If we wanted to take this all the way to Redis, we could probably also improve responsiveness by using redis' blocking list commands instead of polling for updates. But that would be a larger refactoring I'd leave for later.

config/application.rb
monotek: Would you mind to extend the pr for that?

djmaze: I am not sure that belongs here because this is not about the cache store.

Also, you could probably just replace the `:file_store` with `:redis_store` in your docker image similar to how it is currently done with memcached. (Which is of course ugly and should be handled differently.)

Also, switching both cache and session store to Redis [might bring a new class of problems](https://stackoverflow.com/a/20216054/3515146), so we should probably also think about how to prevent those.

thorsteneckel: I agree. We should limit the scope of this change to a minimum and instead iterate faster with smaller changes.

mgruner:
Hi @djmaze. The failing tests were caused by missing credentials, that should be fixed now in `develop`.
Would you be so kind to update your branch to the current state?

mgruner:
@djmaze FYI, your changes are now in an internal branch and are playing well with the CI. We can take it from here, I believe. More updates soon.

djmaze:
@mgruner Sorry for not answering quicker. Thanks for taking care! Just tell me if you need help / testing.

djmaze:
@mgruner Sorry for not answering quicker. Thanks for taking care! Just tell me if you need help / testing.

thorsteneckel:
Fantastic work @djmaze - thanks a ton for your contribution and patience 💪 Will be part of Zammad 5

mgruner:
@djmaze thanks again for this excellent contribution. I merged it more or less unchanged, with some changes to our CI to properly cover this.
This will be a very nice new feature for Zammad 5.0. Thanks again!

mgruner:
Oh and for the record - I agree that using Memcached for the Rails cache is probably the better tool for the purpose. I will make sure this will be configurable and documented in a nice way for Zammad 5.0 too.

Issue #3371, opened 2021-01-15
nginx zammad.conf file can not handle IPv6


Supersonic112:
Dear Zammad team,

The default zammad.conf file for Nginx does not make the server listen on IPv6, which leads to very strange behavior when running Zammad on a Machine with IPv6.

**Proposed solution**: This issue can be fixed by adding the line `listen [::]:80;` below `listen 80;` in the files contrib/nginx/zammad.conf and contrib/nginx/zammad_ssl.conf and possibly also `listen [::]:9200` below `listen 9200;` in contrib/nginx/elasticsearch.conf .

### Infos:

* Used Zammad version: 3.6.0-1610545422.444c3a5d.buster
* Installation method (source, package, ..): package
* Operating system: Debian 10 (Buster)
* Database + version: postgresql, 
* Elasticsearch version: 
* Browser + version: Firefox 84, Chrome 87

### Expected behavior:

* when visiting zammad.domain.tld, the Zammad login page should be loaded, including all resources and assets (e.g. https://domain.tld/assets/application-9f83dd....css, https://zammad.domain.tld/assets/images/logo.svg, ...)

### Actual behavior:

* most of the time only some of the assets load, leaving a half-loaded web page and some requests that return a 404
* sometimes only the "Welcome to Nginx" start page shows up, giving no error at all

### Steps to reproduce the behavior:

* install Zammad on a machine that has (and uses) both an IPv4 and IPv6 address behind an Nginx reverse proxy

Yes I'm sure this is a bug and no feature request or a general question.

PR #3511, opened 2021-04-16
Listen to both IPv4 and IPv6 #3371

maxxer:

### DIFF
--- a/contrib/nginx/zammad.conf
+++ b/contrib/nginx/zammad.conf
@@ -12,6 +12,7 @@ upstream zammad-websocket {
 
 server {
     listen 80;
+    listen [::]:80;
 
     # replace 'localhost' with your fqdn if you want to use zammad from remote
     server_name localhost;


thorsteneckel:
@mantas would you mind having a look?

mantas:
Hi @maxxer 

Thank you for your contribution! Your changes are now merged with the following commit https://github.com/zammad/zammad/commit/062f19df3c227441e72ea548fcce294ebbf2211a

monotek:
This breaks our compose container :(

```
zammad-nginx_1          | starting nginx...
zammad-nginx_1          | nginx: [emerg] a duplicate listen 0.0.0.0:8080 in /etc/nginx/sites-enabled/default:15
```

@maxxer 

Maybe this works for ipv4 and ipv6.

```
listen [::]:80 ipv6only=off;
```

I'm a bit bussy at the moment.
Could you try that out?

MrGeneration:
The reason for this issue lays within the docker-entry point of nginx:
https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/10-listen-on-ipv6-by-default.sh#L24

I personally believe that vHost files shouldn't contain listen directives at all for default ports. 
A workaround would to provide the IPv6 option commented out to help admins but ensure functionality on existing systems.

As it's already merged I believe this requires a regression issue.

maxxer:
> The reason for this issue lays within the docker-entry point of nginx:
> https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/10-listen-on-ipv6-by-default.sh#L24

The `grep` contain two spaces, maybe that's why it's failing?

> I personally believe that vHost files shouldn't contain listen directives at all for default ports.
> A workaround would to provide the IPv6 option commented out to help admins but ensure functionality on existing systems.

Indeed, one has to always edit the file for `server_name` at least.

> As it's already merged I believe this requires a regression issue.

Apologies for the trouble

monotek:
We don't use the nginx docker container.
We're installing nginx in a debian container.

Imho `listen [::]:80 ipv6only=off;` should work for ipv4 and ipv6 in a single line but the sed in the docker entrypoint would still fail.
Added a comment to @MrGeneration pr, which should fix it for any version: https://github.com/zammad/zammad-docker-compose/pull/207#discussion_r619641298

Closing this pr again.

Issue #3512, opened 2021-04-16
local gitlab without https is only partially supported


jan-wi:
### Infos:

* Used Zammad version: 4.0.0
* Installation method (source, package, ..): package
* Operating system: Ubuntu Linux 20.4
* Database + version: postgresql 13
* Elasticsearch version: elasticsearch 7.12
* Browser + version: Firefox Developer 88

### Expected behavior:

* Adding a gitlab Issue to a ticket while using a local gitlab with an HTTP Address similar to the API endpoint

### Actual behavior:

* Adding a gitlab Issue to a ticket while using a local gitlab without HTTPS is not allowed since the regex requires an HTTPS Address while the API Configuration allows for an HTTP Address.
https://github.com/zammad/zammad/blob/260afe1160bb0c9709b83afa31265bf9a9efd0fb/lib/gitlab/linked_issue.rb#L100 

### Steps to reproduce the behavior:

* On entering an http:// ... Address the regex rejects the URL with the error "Invalid GitLab issue link format" 
* The regex could be changes to accomodate http urls in local enviroments

Yes I'm sure this is a bug and no feature request or a general question.

PR #3513, opened 2021-04-16
Fixes Issue #3512

jan-wi:
allow http address for local gitlab instance

Fixes Issue #3512

### DIFF
--- a/lib/github/linked_issue.rb
+++ b/lib/github/linked_issue.rb
@@ -97,7 +97,7 @@ class GitHub
     end
 
     def variables!(url)
-      if url !~ %r{^https://([^/]+)/([^/]+)/([^/]+)/issues/(\d+)$}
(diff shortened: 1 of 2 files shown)


thorsteneckel:
Thanks @jan-wi for your contribution! We should also do this change for GitHub because people might use their GitHub Enterprise installation with HTTP only as well? See: https://github.com/zammad/zammad/blob/3baec9c6060da269b0ede3126eef5fe987b1363c/lib/github/linked_issue.rb#L100

@mantas would  you mind taking this one over?

jan-wi:
> We should also do this change for GitHub because people might use their GitHub Enterprise installation with HTTP only as well?

@thorsteneckel Change was adapted to the github lib as well ( 05ab1dae6e1fbef339c3e4c979ed447996dd9cee )

thorsteneckel:
Fixed with https://github.com/zammad/zammad/commit/a5de91064a2cf05855b0158ebece5e7819e3b840 - backport to 4.0.1 in the making 🚀

PR #3531, opened 2021-04-27
prevent "CSRF token verification failed!" error

amigthea:
the ssl configuration was incomplete, I implemented the necessary parameter

### DIFF
--- a/contrib/nginx/zammad_ssl.conf
+++ b/contrib/nginx/zammad_ssl.conf
@@ -94,6 +94,7 @@ server {
     proxy_set_header Connection "Upgrade";
     proxy_set_header CLIENT_IP $remote_addr;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto	$scheme;
     proxy_read_timeout 86400;
     proxy_pass http://zammad-websocket;
   }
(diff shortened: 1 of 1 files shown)


MrGeneration:
Sorry, but we decided that we will not provide such workarounds within the default configuration. 
You can read more on the reasons why here:
https://github.com/zammad/zammad/issues/2829#issuecomment-757861101

The docker-compose in fact already supports this:
https://github.com/zammad/zammad-docker-compose/blob/master/containers/zammad/docker-entrypoint.sh

See also:
* https://docs.zammad.org/en/latest/getting-started/configure-webserver.html
* https://docs.zammad.org/en/latest/install/docker-compose/environment.html

amigthea:
but that's a loop!

1) the default installation is on http
2) to switch to https on docker compose we need to provide an additional .conf file because the default one (/etc/nginx/sites-enabled/default) is overwritten at every container restart, making your entrypoint workaround useless because it will not apply to the additional ssl .conf
3) the default template of the ssl .conf file don't provide that mandatory parameter
4) ???

MrGeneration:
Nevermind me, sorry. You're right. The directive is absolutely missing while it's available (and required) for running the instance without issues. No idea why this was never added. 

I was going to reference another reverted PR but saw that variables are different.

@mantas could you please take over?

mantas:
@monotek do you mind to take a quick look?

monotek (approved):

monotek:
LGTM

mantas:
Hi @Azertooth,

Your PR is now merged with the commit above. Thank you for your contribution!

PR #3555, opened 2021-05-11
ogg preview with html5 audio control

hugodeco:
untitled


mgruner:
Hi @hugodeco. Thanks a lot for your proposal and your improvement that makes the usability of Zammad better! We believe this is a good solution for your use-case and could be distributed e. g. via an addon.
The handling of attachments is something that needs to be improved in general, and with an appropriate architecture in place that supports many use-cases. Since the PR does not provide such a new structure but focuses on a particular use-case, we cannot merge it into the Zammad core at this point. Therefore I'll close it now; it will stay available here for reference.
Thanks and best regards, Martin

PR #3561, opened 2021-05-15
Improve PostgreSQL backup/restore

alxndr42:
- More robust config parsing
- Support for non-local database
- Various other bugs fixed

### DIFF
--- a/contrib/backup/functions
+++ b/contrib/backup/functions
@@ -20,10 +20,12 @@ function delete_old_backups () {
 }
 
 function get_db_credentials () {
-  DB_ADAPTER="$(grep 'adapter:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*adapter:\(\| \)//g')"
-  DB_NAME="$(grep 'database:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*database:\(\| \) //g')"
-  DB_USER="$(grep 'username:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*username:\(\| \)//g')"
-  DB_PASS="$(grep 'password:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*password:\(\| \)//g')"
+  DB_ADAPTER="$(grep -m 1 '^[[:space:]]*adapter:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*adapter:\(\| \)//g')"
+  DB_HOST="$(grep -m 1 '^[[:space:]]*host:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*host:\(\| \) //g')"
+  DB_PORT="$(grep -m 1 '^[[:space:]]*port:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*port:\(\| \) //g')"
+  DB_NAME="$(grep -m 1 '^[[:space:]]*database:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*database:\(\| \) //g')"
+  DB_USER="$(grep -m 1 '^[[:space:]]*username:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*username:\(\| \)//g')"
+  DB_PASS="$(grep -m 1 '^[[:space:]]*password:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*password:\(\| \)//g')"
 
   if [ "${DEBUG}" == "yes" ]; then
     echo "adapter=${DB_ADAPTER} dbname=${DB_NAME} dbuser=${DB_USER} dbpass=${DB_PASS}"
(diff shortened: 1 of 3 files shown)


contrib/backup/functions
```diff
@@ -20,10 +20,12 @@ function delete_old_backups () {
...
-  DB_ADAPTER="$(grep 'adapter:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*adapter:\(\| \)//g')"
-  DB_NAME="$(grep 'database:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*database:\(\| \) //g')"
-  DB_USER="$(grep 'username:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*username:\(\| \)//g')"
-  DB_PASS="$(grep 'password:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*password:\(\| \)//g')"
+  DB_ADAPTER="$(grep -m 1 '^[[:space:]]*adapter:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*adapter:\(\| \)//g')"
+  DB_HOST="$(grep -m 1 '^[[:space:]]*host:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*host:\(\| \) //g')"
+  DB_PORT="$(grep -m 1 '^[[:space:]]*port:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*port:\(\| \) //g')"
+  DB_NAME="$(grep -m 1 '^[[:space:]]*database:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*database:\(\| \) //g')"
+  DB_USER="$(grep -m 1 '^[[:space:]]*username:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*username:\(\| \)//g')"
+  DB_PASS="$(grep -m 1 '^[[:space:]]*password:' < ${ZAMMAD_DIR}/config/database.yml | sed -e 's/.*password:\(\| \)//g')"
```
7adietri, line 28: If you do only minimal edits to `config/database.yml` the original `grep` commands will return all occurrences of each setting, including commented out DB adapters. These versions should return only the first occurrence of each active setting, which should not change anything for configuration files that had no problems before.

Ideally the YAML should be parsed properly instead of using `grep`, but I'll leave that for another PR. :grimacing:

contrib/backup/functions
```diff
@@ -51,9 +53,7 @@ function backup_db () {
     ln -sfn ${BACKUP_DIR}/${TIMESTAMP}_zammad_db.mysql.gz ${BACKUP_DIR}/latest_zammad_db.mysql.gz
   elif [ "${DB_ADAPTER}" == "postgresql" ]; then
     echo "creating postgresql backup..."
-    su -c "pg_dump --no-privileges --no-owner -c ${DB_NAME} | \
-      gzip > /tmp/${TIMESTAMP}_zammad_db.psql.gz" postgres
-    mv /tmp/${TIMESTAMP}_zammad_db.psql.gz ${BACKUP_DIR}
+    su -c "${DB_PASS:+PGPASSWORD=$DB_PASS} pg_dump --no-privileges --no-owner ${DB_HOST:+-h $DB_HOST} ${DB_PORT:+-p $DB_PORT} ${DB_USER:+-U $DB_USER} -d ${DB_NAME}" zammad | gzip > ${BACKUP_DIR}/${TIMESTAMP}_zammad_db.psql.gz
```
7adietri, line 56: I guess the original command works if Zammad and PostgreSQL are running on the same host **and** you're running the backup script as `root`. I think it should really be the `zammad` user running `psql`, like the restore command below.

contrib/backup/functions
```diff
@@ -177,7 +177,7 @@ function restore_zammad () {
       zammad run rake db:drop
       zammad config:set DISABLE_DATABASE_ENVIRONMENT_CHECK=0
     else
-      ${ZAMMAD_DIR}/bin/rake db:drop
+      DISABLE_DATABASE_ENVIRONMENT_CHECK=1 ${ZAMMAD_DIR}/bin/rake db:drop
```
alxndr42, line 180: Without the variable, `rake` would refuse to drop the production DB on my test setup.

contrib/backup/zammad_backup.sh
```diff
@@ -3,7 +3,6 @@
 # zammad backup script
 #
 
-PATH=/sbin:/bin:/usr/sbin:/usr/bin:
```
alxndr42, line 6: This prevented `ruby` from being found on my test setup.

contrib/backup/functions
7adietri: The rest is adding values for DB host, port, user and password, if present, to allow for a non-local DB.

contrib/backup/functions
```diff
@@ -51,9 +53,7 @@ function backup_db () {
     ln -sfn ${BACKUP_DIR}/${TIMESTAMP}_zammad_db.mysql.gz ${BACKUP_DIR}/latest_zammad_db.mysql.gz
   elif [ "${DB_ADAPTER}" == "postgresql" ]; then
     echo "creating postgresql backup..."
-    su -c "pg_dump --no-privileges --no-owner -c ${DB_NAME} | \
-      gzip > /tmp/${TIMESTAMP}_zammad_db.psql.gz" postgres
-    mv /tmp/${TIMESTAMP}_zammad_db.psql.gz ${BACKUP_DIR}
+    su -c "pg_dump --no-privileges --no-owner ${DB_HOST:+-h $DB_HOST} ${DB_PORT:+-p $DB_PORT} ${DB_USER:+-U $DB_USER} -d ${DB_NAME}" zammad | gzip > ${BACKUP_DIR}/${TIMESTAMP}_zammad_db.psql.gz
```
alxndr42, line 56: I guess the original command works if Zammad and PostgreSQL are running on the same host **and** you're running the backup script as `root`. I think it should really be the `zammad` user running `pg_dump`, like the restore command below.

The rest is adding values for DB host, port, and user, if present, to allow for a non-local DB.

contrib/backup/functions
7adietri, line 28: If you do only minimal edits to `config/database.yml`, the original `grep` commands will return all occurrences of each setting, including commented out DB adapters. These versions should return only the first occurrence of each active setting, which should not change anything for configuration files that had no problems before.

Ideally the YAML should be parsed properly instead of using `grep`, but I'll leave that for another PR. :grimacing:

alxndr42, line 28: If you do only minimal edits to `config/database.yml`, the original `grep` commands will return all occurrences of each setting, including commented out DB adapters. These versions should return only the first occurrence of each active setting, which should not change anything for configuration files that had no problems before.

Ideally the YAML should be parsed properly instead of using `grep`, but I'll leave that for another PR. :grimacing:

mgruner:
@7adietri  I tried to consolidate all changes of this and another related PR in #3621. This could be the basis for a merge if there is a consensus. 
Would you be so kind to review the changes there and let me know if they work for you?

alxndr42:
@mgruner Thanks! I'll try to run some tests over the weekend.

PR #3570, opened 2021-05-25
replacing find -mtime with find -mmin

dvnkln:
Changing the find parameter for deleting old backups would make things more intuitive.

mtime +1 returns files accessed two days ago.
mmin +$(((60*24)*1)) returns files accessed 1 day (1440mins) ago.

Therefore you're configuring your HOLD_DAYS let's say to "3" expecting to keep backups for three days but it's keeping backups for four days right now.
mmin behaves more accurate to what the user would expect.

see find man (https://linux.die.net/man/1/find) for details

### DIFF
--- a/contrib/backup/functions
+++ b/contrib/backup/functions
@@ -16,7 +16,7 @@ function get_backup_date () {
 }
 
 function delete_old_backups () {
-  test -d ${BACKUP_DIR} && find ${BACKUP_DIR}/*_zammad_*.gz -type f -mtime +${HOLD_DAYS} -delete
+  test -d ${BACKUP_DIR} && find ${BACKUP_DIR}/*_zammad_*.gz -type f -mmin +$(((60*24)*${HOLD_DAYS})) -delete
 }
 
 function get_db_credentials () {


dvnkln:
Figured i could provide a small example instead of only linking some man page. Here you go:

Let's act as if `HOLD_DAYS` is configured to `1`.

Current time (`date "+%Y-%m-%d %H:%M:%S"`):
`2021-05-25 16:24:50`. So everything before `2021-05-24 16:24:50` should be considered, right?

```
find -type f -mtime +1 -exec stat -c "name: %n -- mtime: %y" {} \; | sort
name: ./20210521220002_zammad_db.psql.gz -- mtime: 2021-05-21 22:29:35.000000000 +0200
name: ./20210521220002_zammad_files.tar.gz -- mtime: 2021-05-21 22:00:40.000000000 +0200
name: ./20210522220001_zammad_db.psql.gz -- mtime: 2021-05-22 22:28:03.000000000 +0200
name: ./20210522220001_zammad_files.tar.gz -- mtime: 2021-05-22 22:00:33.000000000 +0200
```

```
find -type f -mmin +1440 -exec stat -c "name: %n -- mtime: %y" {} \; | sort
name: ./20210521220002_zammad_db.psql.gz -- mtime: 2021-05-21 22:29:35.000000000 +0200
name: ./20210521220002_zammad_files.tar.gz -- mtime: 2021-05-21 22:00:40.000000000 +0200
name: ./20210522220001_zammad_db.psql.gz -- mtime: 2021-05-22 22:28:03.000000000 +0200
name: ./20210522220001_zammad_files.tar.gz -- mtime: 2021-05-22 22:00:33.000000000 +0200
name: ./20210523220001_zammad_db.psql.gz -- mtime: 2021-05-23 22:29:32.000000000 +0200
name: ./20210523220001_zammad_files.tar.gz -- mtime: 2021-05-23 22:00:39.000000000 +0200
```

The backup created on `2021-05-23` is missing when using `mtime` instead of `mmin`.
Since `mmin` could be "too" precise considering the function isn't called on the exact same second you could add "some" minutes after the submitted calculation to make sure the backup created f. e. 1441 minutes ago instead of 1440 isn't deleted.

monotek:
If we implement this, docker-compose repo should be adjusted likewise.

See: 
https://github.com/zammad/zammad-docker-compose/blob/master/docker-compose.yml#L12
https://github.com/zammad/zammad-docker-compose/blob/master/containers/zammad-postgresql/backup.sh#L27

MrGeneration:
Sorry up to now I thought that we're using the zammad/zammad sources for this regard. 
Seems like I was wrong. Thanks for the heads up.

Current ETA on this PR from my end is unknown because of missing time ressources.

mgruner:
@svnr-dvnkln thanks a lot for this contribution. I added a +60 as a safety net for HOLD_DAYS=1 as you also suggested, to make sure that backups created with +1441 minutes (the day before, with some script run time) are not already deleted and thus no backup present. I hope this works for you.

@MrGeneration @monotek ping.  @monotek will you update docker-compose?

PR #3673, opened 2021-07-23
English Grammar Correction

nichogenius:
"Zammad need a restart" is poor grammar

Either of these are correct:
"Zammad requires a restart" - Preferred technical form
"Zammad needs a restart"


MrGeneration:
Not sure which modal this test aims at but this commit may break the test:
https://github.com/zammad/zammad/blob/2e8415d5d0bec0278402d742940f72f07ea86c55/app/assets/javascripts/app/controllers/_plugin/maintenance.coffee#L46-L73

nichogenius:
Sorry - I closed this pull request and opened a new one thinking it would be cleaner after a rebase.  I honestly just searched for the string 'Zammad need a restart' because I had seen it in a modal.  I found one hit and committed a change not realizing I only fixed the test, but not the modal.

PR #3674, opened 2021-07-24
English grammar correction

nichogenius:
"Zammad need a restart" is poor grammar

Either of these are correct:
"Zammad requires a restart" - Preferred technical form
"Zammad needs a restart"


mgruner:
Thanks @nichogenius, this was recently fixed via 64a87b1c67c2dad7e9311491115f73606b14acc6.

PR #3676, opened 2021-07-26
Fixes #3137 - New email account expert view cannot be opened without …

### LINKED ISSUE #3137: New email account expert view cannot be opened without filling in all fields
### Infos:

* Used Zammad version: 3.4, 4.1/stable
* Operating system: any
* Database + version: any
* Elasticsearch version: any
* Browser + version: any

### Expected behavior:

* When adding a new email account, expert settings can be brought up right away

### Actual behavior:

* When adding a new email account, all form fields has to be filled in to be allowed into expert view

Screenshot 2020-07-29 at 13 50 50

* Go to Manage > Email and click "Add new"

* Click "Experts"

goost:
…filling in all fields

This fix tries to balance:

1. Improving the user experience when a user wants to input
   expert/advanced settings directly.
2. Do as little code changes as possible.
3. Leave the internal state as is.

A such, the fix assumes that the 'Experts' form is an extension of the
base one instead of being additional settings and shows all the relevant
settings from the base form in the expert view too. This allows the user
to directly open the experts view when adding a new email. Already entered
values are not lost and related fields are filled in.

Pros of this approach:

- In the basic and expert case, the user has only one form to fill in
  with all the relevant settings.
- The validation of the form does not need special handling.
- The user gets immediate feedback about validation errors.
- Housekeeping on the code side is minimal.

Cons:

- An additional array, which is semantically but not syntactically linked
  to the `configureAttributesBase` array is created (email.coffee:416-418)
- `@account.meta` needs manual update to be in sync (email.coffee:553-556)

The last con could be not needed but because of (3) and my unfamiliarity
with the implications of having `@account.meta` out-of-sync, I opt-in
to update it.

Another approach considered was moving the validation of both forms to
occur when the 'submit' button of the expert form was pressed. This would
get rid of the mentioned cons but introduce jarring behaviour for the user
when a base form validation failed on the expert view and the modal would
change back.

Potential issues:

- Going back from the expert view does not carry over related settings
  as it does going from the base view to the expert view. This is in-line
  with the behaviour before this fix.
- Going to the expert view does not use its own callback but reuses
  `probeBasedOnIntro` (email.coffee:484)
- Changing the 'SSL/STARTTLS' setting to 'yes' in the expert view does
  not set the port if the port is 143 (the default for no SSL).
  Changing the setting to 'no' does set the port.

The last point is technically not in scope of the issue and the other
points would need more code changes than strictly necessary for
functionality, violating (2). However, those small further
improvements could result in a net win and could be added with further
commits.

### DIFF
--- a/app/assets/javascripts/app/controllers/_channel/email.coffee
+++ b/app/assets/javascripts/app/controllers/_channel/email.coffee
@@ -410,6 +410,15 @@ class ChannelEmailAccountWizard extends App.ControllerWizardModal
       { name: 'options::keep_on_server',  display: 'Keep messages on server', tag: 'boolean', null: true, options: { true: 'yes', false: 'no' }, translate: true, default: false, item_class: 'formGroup--halfSize' },
     ]
 
+    if !@channel
+      #Email Inbound form opened from new email wizard, show full settings
+      configureAttributesInbound = [
+        { name: 'options::realname', display: 'Organization & Department Name', tag: 'input',  type: 'text', limit: 160, null: false, placeholder: 'Organization Support', autocomplete: 'off' },
+        { name: 'options::email',    display: 'Email',    tag: 'input',  type: 'email', limit: 120, null: false, placeholder: 'support@example.com', autocapitalize: false, autocomplete: 'off' },
+        { name: 'options::group_id', display: 'Destination Group', tag: 'select', null: false, relation: 'Group', nulloption: true },
+      ].concat(configureAttributesInbound)
+
+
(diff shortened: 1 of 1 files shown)


mgruner:
Merged via 0c80232. Thanks!

MrGeneration:
👋 This approach has a clear down side if you choose to not provide the email address up front. 
Personally I'm not sure if that's a good idea.

This also requires documentation updates because you chose to change behaviors.

MrGeneration:
Follow up: Documentation updates ar enot required as I've been clever enough in the screencast and the change does not fall into account. :)

PR #3677, opened 2021-07-26
Fixes #3460 - "Drop files here" drag area not always hiding

### LINKED ISSUE #3460: "Drop files here" drag area not always hiding
### Infos:

* Used Zammad version: 3.6.0-1615986441.da478686.bionic, 4.1/stable
* Installation method (source, package, ..): package
* Operating system: Ubuntu 18.04.5 LTS 
* Browser + version: Google Chrome Version 89.0.4389.90 (Official Build) (64-bit)

### Expected behavior:

* "Drop files here" drop area always hiding itself when releasing a drag

### Actual behavior:

* When releasing dragged text in an close proximity below the actual drop area results in it staying visible

* Select text
* Start dragging selected text
* Release it slightly below the box

Also see this gif:
Peek 2021-03-19 19-01

goost:
Selecting, dragging & dropping a text into the text-field under
`/#ticket/zoom/:id` calls `onDrop(e)` from html5Upload.js:104.
This sets `inCounter` to 0 but call calls `onDragEnter()` afterwards,
followed by the method to process files. (As only text was dragged, no
files will be processed.)
In the end, this sequence of events result in `inCounter` being 1 instead
of 0 as `onDragEnter()` increments the counter by one. I imagine the
invariant is that each `onDragEnter()` is followed by a `onDragLeave()`,
which is broken in this case.

As `inCounter` starts at 1 instead of 0, the next time a text is dragged
around and dropped somewhere, the 'Drop Files Here' area will not be
hidden. The first commit of the pull-request fixes that by moving
`inCounter = 0` in the `onDrop(e)` method after `onDragEnter(e)`,
instead of before.

When dragging & dropping e.g. selected text on the text-field under
`#ticket/create/id/:id` the 'drop' event is usually captured by the
'dropArea' and bubbled up until the to the form attached event handler
captures it and calls `onDrop(e)` of html5Upload.js.
However, the area of the div with the id 'TicketArticle_:id_body' is
slightly larger as the area of the 'dropArea' div (see screenshots).
Those two divs are siblings to one another and the
'TicketArticle_:id_body' has a 'drop' event listener to
`Plugin.onDrop(e)` from jquery.contenteditable.js:414, which captures
the event and stops propagation, if the user let go of a drag just
directly below the 'dropArea' (a few pixels there belong to
'TicketArticle_:id_body').
This (a) throws a null-exception as there is no file to upload if text
was dragged and (b) stops propagation and in consequence, the removal
of the 'dropArea', blocking input.

The second commit of this pull-request fixes it by explicitly
checking for null in `Plugin.onDrop(e)`, however, changing the sizes
of the divs could also be a viable fix.

It may be needed to re-evaluate the whole drag'n'drop mechanism but my
unfamiliarity with the code base leaves me hesitant to do any big changes.

Screenshot from 2021-07-26 15-59-30
Screenshot from 2021-07-26 15-59-36

### DIFF
--- a/app/assets/javascripts/app/lib/base/html5Upload.js
+++ b/app/assets/javascripts/app/lib/base/html5Upload.js
@@ -102,9 +102,9 @@
                     }
                 };
                 onDrop = function (e) {
-                    inCounter = 0
                     onDragEnter(e);
                     hideDropZone(dropContainer)
+                    inCounter = 0
                     manager.processFiles(e.dataTransfer.files)
                 };
                 showDropZone = function(dropContainer) {
--- a/app/assets/javascripts/app/lib/base/jquery.contenteditable.js
+++ b/app/assets/javascripts/app/lib/base/jquery.contenteditable.js
@@ -412,8 +412,6 @@
   }
 
   Plugin.prototype.onDrop = function (e) {
-    e.stopPropagation();
-    e.preventDefault();
     this.log('drop')
 
     var dataTransfer
@@ -431,6 +429,10 @@
     x = e.clientX
     y = e.clientY
     var file = dataTransfer.files[0]
+    // Do not stop propagation if the there is no file to handle
+    if (!file) return;
+    e.stopPropagation();
+    e.preventDefault();
 
     // look for images
     if (file.type.match('image.*')) {


MrGeneration:
**Attention**
This PR collides with assigned developer work and may thus be dropped.

image

mantas:
Hi @goost,

Thank you for your contribution! Unfortunately, as Marcel said, another developer was already working on this. I'll have to close this PR without merging.

PR #3682, opened 2021-07-28
LDAP: apply user filter during auth on bind

sumpfralle:
Previously the user filter was not used for the login attempt.
Thus in case of multiple LDAP entries matching the login name (e.g. `cn=foo`), the wrong entry could be used (by default the first result is attempted).

In our case we have an LDAP tree with an "identity" and an "account" for each user (both residing in different branches of the LDAP tree).

The `user_filter` needs to be applied during login attempts in order to avoid this ambiguity.

Please note, that I am can barely parse ruby. Thus you may feel tempted to use a more conventional style for this change. You are very welcome to do so.

### DIFF
--- a/lib/ldap/user.rb
+++ b/lib/ldap/user.rb
@@ -95,7 +95,7 @@ class Ldap
     def valid?(username, password)
       bind_success = @ldap.connection.bind_as(
         base:     @ldap.base_dn,
-        filter:   "(#{login_attribute}=#{username})",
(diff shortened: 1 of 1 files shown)


dominikklein:
Thanks for the contribution, merged in: d4dfd750b722071c9b8f1ad4a2b7b59f82656966.

(because of a mistake with the wrong reference)

PR #3744, opened 2021-09-19
Stop the execution of the backup script if a command or pipeline has an error

helmo:
Hi,

I was surprised by the log output below:
```
# Zammad backup started - Sat Sep 18 03:35:01 CEST 2021!

creating file backup...
creating postgresql backup...

gzip: stdout: No space left on device

# Zammad backuped successfully - Sat Sep 18 04:00:59 CEST 2021!
```

The 'No space left on device' is obviously mine to fix, but the 'Zammad backuped successfully' line below it annoyed me ;)

Adding `set -e` should prevent that from happening.

### DIFF
--- a/contrib/backup/zammad_backup.sh
+++ b/contrib/backup/zammad_backup.sh
@@ -3,6 +3,9 @@
 # zammad backup script
 #
 
+# Stop the execution of the script if a command or pipeline has an error
+set -e
+
 # shellcheck disable=SC2046
 BACKUP_SCRIPT_PATH="$(dirname $(realpath $0))"
 


thorsteneckel:
Hi @helmo - Thanks for your contribution. @MrGeneration / @monotek are you OK with it?

monotek (approved):

MrGeneration (approved):

mgruner:
Hi @helmo, thanks for your suggestion. When testing this, it showed to cause unwanted side effects about exit codes of commands in other places. For example, the backup script failed if `/var/tmp/zammad_backup` has no files or if the directory is entirely missing. This would then stop the backup process, which is probably not desirable.
I'll close this PR for now, please feel free to improve and reopen it at your convenience.

helmo:
@mgruner Those sound like scenario's where you'd most definitely want things to blow up. Could you be more specific where it failed?

Or is it just test code that fails?  you could add `set +e` and `set -e` around that specific command to ignore the error/

mgruner:
@helmo errors like mentioned could happen when you run the backup script for the first time, for example. The `zammad_backup` directory is not present at that point yet. And that is not a case where things should blow up.

helmo:
I don't agree. failing silently is worse.

The installation process should create this directory, or at the very least should the user be warned that the dir is not present causing backups to fail.

mgruner:
@helmo there might be a misunderstanding here. Right now the script works well in this case, because it doesn't stop on a single line with exit code != 0, for example a `find` call. It will continue and create the directory later on.
Your PR causes it to fail by changing the behaviour without adapting the other code.

helmo:
Do you mean the function delete_old_backups()?

It's the shell expansion that fails in the line below:
`test -d ${BACKUP_DIR} && find ${BACKUP_DIR}/*_zammad_*.gz -type f -mmin +$(((60*24)*${HOLD_DAYS}+60)) -delete`

A safer way would be to use `-name`:
`test -d ${BACKUP_DIR} && find ${BACKUP_DIR}  -name '*_zammad_*.gz' -type f -mmin +$(((60*24)*${HOLD_DAYS}+60)) -delete`

Then it will not fail on shell expansion if the backup dir is empty.

PR #3775, opened 2021-09-28
Fix postinstall.sh getting es_url

### LINKED ISSUE #2674: Zammad preflight check warning output causes Syntax-Error in postinstall.sh and failing installation
### Infos:

* Used Zammad version: 3.1.x
* Installation method (source, package, ..): package installation
* Operating system: tested on Debian and SuSe, should affect all
* Database + version: any
* Elasticsearch version: ES 5.6.16
* Browser + version: any

### Expected behavior:

postinstall.sh runs clean and throws no errors.

### Actual behavior:

postinstall.sh throws this error:
```
After this you might need to rebuild the searchindex by:
zammad run rake searchindex:rebuild
sh: 1: Syntax error: ")" unexpected
Setting default Logging to file, set via "zammad config:set RAILS_LOG_TO_STDOUT=true" if you want to log to STDOUT!
# Starting Zammad
```
(Offending line: https://github.com/zammad/zammad/blob/develop/contrib/packager.io/functions#L265 )

* upgrade Zammad from 2.9 to 3.1 as example

benrubson:
Hi,

This fixes #2674.

During `postinstall.sh`, the following command is executed :
`zammad run rails r "puts Setting.get('es_url')"`

Which can lead to the following output :
```
Warning: Database config value 'max_allowed_packet' less than Zammad setting 'Maximum Email Size'
Zammad will fail to process emails (both incoming and outgoing)
larger than the value of 'max_allowed_packet' (16MB).
Please increase this value in your MariaDB configuration accordingly.
http://127.0.0.1:9200
```
However, as a warning, the first 4 lines are yellow-coloured, which leads to some invisible characters at the beginning of the last line :
`^[[0mhttp://127.0.0.1:9200`

This PR simply gets ride of them adding a newline before `es_url` output.

Post install script will now run flawlessly.

Thx 👍

### DIFF
--- a/contrib/packager.io/functions
+++ b/contrib/packager.io/functions
@@ -228,7 +228,7 @@ function create_webserver_config () {
 function setup_elasticsearch () {
   echo "# Configuring Elasticsearch..."
 
-  ES_CONNECTION="$(zammad run rails r "puts Setting.get('es_url')"| tail -n 1 2>> /dev/null)"
+  ES_CONNECTION="$(zammad run rails r "puts '',Setting.get('es_url')"| tail -n 1 2>> /dev/null)"
 
   if [ -z "${ES_CONNECTION}" ]; then
     echo "-- Nevermind, no es_url is set, leaving Elasticsearch untouched ...!"


benrubson:
Could this be perhaps included in the next package build ?
So that subsequent updates will proceed flawlessly.
Many thx 👍

thorsteneckel:
Yes! It's on my list 👍

PR #3885, opened 2021-12-09
Wrong translation on Turkish

mandev0:
This translation was false. Someone added a hospital name to the footer. The part was "IKCU Atatürk Eğitim Ve Araştırma Hastanesi". This is the correct translation.

### DIFF
--- a/i18n/zammad.tr.po
+++ b/i18n/zammad.tr.po
@@ -11302,9 +11302,7 @@ msgstr ""
 msgid ""
 "You're already registered with your email address if you've been in touch "
 "with our support team."
-msgstr ""
-"IKCU Atatürk Eğitim Ve Araştırma Hastanesi Bilgi İşlem Arıza Bildirim "
-"Uygulaması"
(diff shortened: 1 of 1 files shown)


MrGeneration:
@mgruner how to we proceed with direct translation file PRs that do not come via weblate?
In this scenario, at least per translation list, the translator credit gets lost with the current state of this PR.

mgruner:
@MrGeneration there are two options. Either we do the transfer work, or we reject all PRs about this and ask people to get an account in Weblate.
I would recommend the latter. We can also place a README file that exaplains this in the i18n/zammad folder.

MrGeneration:
I can't decide that.
If we'd transfer it via File import you can also provide a Name and email address, that automatically creates an account and would require consent.

mgruner:
Hi @mandev0. Thank you so much for spotting this and suggesting a fix. May I encourage you to get an account on [translations.zammad.com](https://translations.zammad.com) and making the required changes there?
To clarify this new process (we just introduced gettext / Weblate a few days ago), I also added a `README.md` file to the `i18n` folder. Sorry for the trouble.

PR #3886, opened 2021-12-09
Make form style for `customer`-input consistent

xf-:
Set parameters like color and background with form-control class.

This template uses the `form-control` it as well without a problem.
https://github.com/zammad/zammad/blob/develop/app/assets/javascripts/app/views/layout_ref/chat_to_ticket.jst.eco

Before
image

After
image

I entered the field, so you can see the position from complete is fine.

(The line end was done by GitHub editor)

### DIFF
--- a/app/assets/javascripts/app/views/generic/object_search/input.jst.eco
+++ b/app/assets/javascripts/app/views/generic/object_search/input.jst.eco
@@ -3,7 +3,7 @@
   <% if @attribute.multiple: %>
     <%- @tokens %>
   <% end %>
-  <input name="<%- @attribute.name %>_completion" class="user-select token-input js-objectSelect" autocapitalize="off" placeholder="<%- @Ti(@attribute.placeholder) %>" autocomplete="off" <%= @attribute.autofocus %> role="textbox" aria-autocomplete="list" value="<%= @name %>" aria-haspopup="true">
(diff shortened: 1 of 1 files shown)


MrGeneration:
Can you please let us know which browser you verified this on?
I can't seem to reproduce this on a current Google Chrome.

xf-:
I use Firefox Linux Dark Theme system setting & browser, but the issue is only on dark themes.

Without a style it fallsback to system style. This is only field without a defined color and background-color.

There are multiple addons to help with it, but because most are set and one has no value i get a mix of fields.
https://addons.mozilla.org/en-US/firefox/addon/text-contrast-for-dark-themes/

OR
https://addons.mozilla.org/en-US/firefox/addon/fix-dark-theme-input-boxes

MrGeneration:
Without being able to further qualify your PR (from my end), please keep in mind:
We have to control over browser and addon behavior. If you use manipulation tools to change UI appearance, this is out of the core's scope.

A developer will have a look any way as soon as there's time.

xf-:
No, the Add-ons trys to fix it. The issue appears in a vanilla installed browser. I wanted to show, it is a more or less common issue with dark theme and not perfect websites.

Input field normally in zammad
image


This special field - Customer select
image


Organization field - Edit Customer has form-control - same autocompelte stuff
image

**Every Field in zammad has this class to get the same input behavior except this one.**

If you have a dark theme the page without any style is dark with a light font and in this case everything is set white background and dark font **EXCEPT** this one field.

MrGeneration:
Thank you for the clarification!

xf-:
Chrome uses a base stylesheet that has a light theme in a dark setup. Firefox provides a dark theme.

Good example - no style page
https://wiki.selfhtml.org/extensions/Selfhtml/example.php/Beispiel:Html5_grundger%C3%BCst.html

Firefox/Chrome
image

rolfschmidt:
thx for your support @xf- ... backport is following

PR #3903, opened 2022-01-03
Fix capitalization for similar string

maxxer:
The same string appears twice in translations, so I made them the same


maxxer:
I'll open a new PR without this mess...

PR #3918, opened 2022-01-14
Various translation fixes

maxxer:
Misc fixes on translation strings


mgruner:
Hi @maxxer, thank you so much for this PR! There was one change about the scheduler job that I couldn't apply at this point, because it requires writing a migration.
There's quite a bunch of source string changes coming in the next days, stay tuned!

Issue #1580, opened 2017-10-24
Time accounting by day


jonasboettcher:
### Infos:

* Used Zammad version: 2.0.x
* Used Zammad installation source: Ubuntu package
* Operating system: Ubuntu 16.04.3 LTS
* Browser + version: Firefox 56.0 (64-Bit)

### Expected behavior:

* Alike #1007 it was nice to add a daily selector to the time accounting overviews in: http://zammad/#manage/time_accounting
With this feature I'd like to get an overview for any selected date, on which tickets somebody has been worked.

### Actual behavior:

* The time accounting overview only shows selectors for each year an month.

### Steps to reproduce the behavior:

* Go to "Admin" -> "Time accounting"

rolfschmidt:
### Proposal

Maybe related to https://github.com/zammad/zammad/issues/4123#issuecomment-1549267400. If the export was moved to the Profile → Reports and extended similar, then a daily export would be possible.

martini:
For me it's currently not clear if this issue is about exporting accounted time by day (not only my month) or only improve the view in the admin interface of the logged accounted time per month?

KoJie:
this function is party covered by the reporting-tool. This is why we block this enhancement for now.

PR #3976, opened 2022-02-24
implemented time accounting export by activity

amirburi:
We use this to transfer accounted time based on activity to our billing system.

It is realted to #1580 but in our opinion it makes more sens to have it activity based per month (not per day).


martini:
PM feedback 2022-03-29: Looks good for PM. Next: QA from a Zammad developer.

rolfschmidt:
Hi @amirburi , 

thanks for your support 👍  I merged your changes.

PR #4043, opened 2022-03-31
Add copying ticket number via click in ticket zoom

Stubenhocker1399:
Hello! :wave: 

I've added a small functionality to copy Ticket# as described this [community post](https://community.zammad.org/t/easy-way-to-copy-ticket-with-a-single-click/8986).

image
image
Peek 2022-03-31 16-40

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom.coffee
@@ -7,12 +7,12 @@ class App.TicketZoom extends App.Controller
     '.scrollPageHeader':   'scrollPageHeader'
 
   events:
-    'click .js-submit':                  'submit'
-    'click .js-bookmark':                'bookmark'
-    'click .js-reset':                   'reset'
-    'click .js-draft':                   'draft'
-    'click .main':                       'muteTask'
-    'click .ticket-number-copy-header':  'copyTicketNumber'
+    'click .js-submit':                 'submit'
+    'click .js-bookmark':               'bookmark'
+    'click .js-reset':                  'reset'
+    'click .js-draft':                  'draft'
+    'click .main':                      'muteTask'
+    'click .ticket-number-copy-header': 'copyTicketNumber'
 
   tooltipCopied = undefined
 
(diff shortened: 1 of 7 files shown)


MrGeneration:
Love it tbh.

app/assets/javascripts/app/controllers/ticket_zoom.coffee
```diff
@@ -1221,6 +1225,26 @@ class App.TicketZoom extends App.Controller
       overview_id: @overview_id
     )
 
+  copyTicketNumber: =>
+    text = $('.active.content .js-objectNumber').first().data('number') || ''
```
martini: IMO this should be `text = @el.find('.js-objectNumber').first().data('number') || '' ` to only query the local @el and not whole browser dom.

app/assets/javascripts/app/controllers/ticket_zoom/meta.coffee
```diff
@@ -14,3 +16,23 @@ class App.TicketZoomMeta extends App.ControllerObserver
       isCustomer: ticket.currentView() is 'customer'
     )
     @renderPopovers()
+
+  copyTicketNumber: =>
```
martini, line 20: I the the same code 2 times. It would be great if copyTicketNumber function would be a generic function in app/assets/javascripts/app/controllers/_application_controller/_base.coffee  where you just need to pass a selector, the texts and maybe a callback if needed.

martini (commented):
Awesome cool pull request! :) Just added 2 notes.

app/assets/javascripts/app/controllers/ticket_zoom/meta.coffee
Stubenhocker1399: Yeah, I had a feeling a comment like this would pop up, just was not sure where to best put this function. :smile:
Speaking of `_base.coffee`, is the existing copyToClipboard function there even used anywhere still? :thinking:

fliebe92:
Hey @Stubenhocker1399, thanks for your pull request. 😄

app/assets/javascripts/app/controllers/ticket_zoom/meta.coffee
Stubenhocker1399: Guess it's used potentially in addon packages.

PR #4104, opened 2022-05-28
Add notification volume slider

Stubenhocker1399:
Hello! :wave: 

I added a volume slider for notifications as described in this [community post](https://community.zammad.org/t/notification-volume-slider/9061).

image

It uses the personal localStorge as different agents might have different audio volume preferences depending on which device they are using.

(Just a note: The PR creation template asked me to run coffeelint, however you've removed this gem in https://github.com/zammad/zammad/commit/6fbdb3ca2abca5e937a3b19701218259254665a8, is this still expected to be run?
Running `coffeelint.rb coffeelint.json -r src/` manually at least seemed to have caused no complaints for this change. :smile: )

### DIFF
--- a/app/assets/javascripts/app/controllers/_profile/notification.coffee
+++ b/app/assets/javascripts/app/controllers/_profile/notification.coffee
@@ -94,6 +94,7 @@ class ProfileNotification extends App.ControllerSubContent
       groups: groups
       config: config
       sounds: @sounds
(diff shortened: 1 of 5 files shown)


dominikklein:
Hello @Stubenhocker1399 

thanks for the idea and the support, but this time we decided in our product team that we do not want to add this feature to the standard of Zammad.

Best regards,
Dominik

Stubenhocker1399:
Alright, no problem!

PR #4146, opened 2022-06-27
No need to echo the token in monitoring/health_check

helmo:
I noticed that the token we use for the monitoring Health Check is echoed back in every request.

This causes it to be exposed in the logs on our monitoring servers.

It's not highly critical, but I see no reason why it should be returned in the first place.

### DIFF
--- a/app/controllers/monitoring_controller.rb
+++ b/app/controllers/monitoring_controller.rb
@@ -40,7 +40,6 @@ curl http://localhost/api/v1/monitoring/health_check?token=XXX
       message: health_status.message,
       issues:  health_status.response.issues,
       actions: health_status.response.actions,
-      token:   token,
     }
     render json: result
   end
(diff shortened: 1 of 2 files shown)


MrGeneration:
**Warning** This is a behavior change and thus might be a potential breaking change.

dominikklein:
@mgruner @mantas maybe you can check this?

mantas:
I agree with this change in general. But at the same time I agree with @MrGeneration that this is a breaking change. Might have to wait till major release...

dominikklein:
@helmo Thanks for your contribution, the plan is to add this with the next Major-Release, because it's a breaking change for this API.

mgruner:
@helmo thanks again for your contribution. We just merged it with some changes. Unfortunately, your author information was lost. That was not intended and I'm sorry for that.

PR #4176, opened 2022-07-17
Update last contact on merge

Stubenhocker1399:
Hi! :wave: 

In this PR I've adjusted ticket merging to update `last_contact_customer_at` and `last_contact_at` as described in [this community post](https://community.zammad.org/t/update-last-contact-on-ticket-merge-if-using-latest-customer-article-time/9666).

I've also added updating `last_contact_agent_at` as it seemed sensible.

image

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -307,6 +307,23 @@ returns
       # which might be a long time ago
       target_ticket.updated_at = Time.zone.now
 
+      # update last contact on target ticket depending on setting or if agent already had responded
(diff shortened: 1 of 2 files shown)


dominikklein:
@Stubenhocker1399 Looks to me like a bug, because of its inconsistent state after the ticket merge action. Maybe you can create an issue with this behaviour?

Stubenhocker1399:
Sure can do! However can only do it in a couple of days though, as currently I'm on the road.

dominikklein:
@Stubenhocker1399 Thanks for your effort :-) Like described in the related issue is this a more complex topic and at the moment we want not only to merge this small piece for the solution in the core product.

Issue #2104, opened 2018-06-28
Shown ticket changed unexpected when viewing an attached image


rkaldung:
### Infos:

* Used Zammad version: 2.5
* Installation method (source, package, ..): source
* Operating system: CentOS 7
* Database + version: Postgresql 
* Elasticsearch version: ?
* Browser + version: Chrome latest

### Expected behavior:

* Modal dialog closed when opening another ticket by clicking on a browser notification.

### Actual behavior:

* When a ticket is opened via browser (desktop) notification a modal dialog stays open

### Steps to reproduce the behavior:

* Enable preview feature `rails r 'Setting.set("ui_ticket_zoom_attachments_preview", true)'`
* Open a ticket with an attached image
* Wait for a new ticket browser (desktop) notification
* Click on the browser (desktop) notification and see how the ticket under the modal dialog changes while the image stays.

Yes I'm sure this is a bug.

SummerStorm:
Fixed via bc489aecef6a1026fc2f4bbeeeb1dbdd89c386c7.

PR #4177, opened 2022-07-17
Fix scrolling image view when using arrow keys

Stubenhocker1399:
Hi! :wave: 

This fixes two bugs when switching between multiple image views using arrow keys; one being sometimes the modal not showing a scroll bar (and not being scroll-able) even though it is too tall to fit, and the other not being able to scroll using arrows up and down.

As a note; changing that the next modal only opens on `onClosed` does not seem like the best fix for the scroll bar bug, as it seems to be an underlying bug in the general modal code and potentially is already an issue in other places as well.

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom.coffee
@@ -337,7 +337,7 @@ class App.TicketZoom extends App.Controller
 
   hide: =>
     @activeState = false
-    $('body > .modal').modal('hide')
+    $('body > .modal').modal('hide') if @shown
     @positionPageHeaderStop()
     @autosaveStop()
     @shortcutNavigationstop()
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_image_view.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_image_view.coffee
@@ -21,12 +21,12 @@ class App.TicketZoomArticleImageView extends App.ControllerModal
     )
 
   nextRight: =>
-    this.nextElement = @parentElement.closest('.attachment').next('.attachment.attachment--preview')
+    @nextElement = @parentElement.closest('.attachment').next('.attachment.attachment--preview')
     return if @nextElement.length is 0
     @close()
 
   nextLeft: =>
-    this.nextElement = @parentElement.closest('.attachment').prev('.attachment.attachment--preview')
+    @nextElement = @parentElement.closest('.attachment').prev('.attachment.attachment--preview')
     return if @nextElement.length is 0
     @close()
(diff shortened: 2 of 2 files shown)


dominikklein:
@Stubenhocker1399 Thanks for the input. When this is related to a bug, can you please create a related issue with some information to reproduce the problem?

Stubenhocker1399:
Okay, sure! :smiley:

fliebe92:
Hey @Stubenhocker1399, thanks for your pull request 😄 

We checked it and right now we're facing two issues:

1. You are using `this.nextElement`, but we encourage to simply use `@nextElement` in our CoffeeScript files.
2. It seems this pull request is introducing a new bug, by opening the image preview for one of your provided test images (special thanks for this), the dialog is automatically closed once. This only happens after a page reload, every attempts to open the image preview are fine afterwards.

We would really appreciate if you could check and improve your pull request.

Hope to read you soon! 😎

Stubenhocker1399:
Glad you found the test images useful :slightly_smiling_face: 

Regarding 2.; are you sure that is a new bug? At least, if I wait for the page to fully load (including any sidebar tickets), I have no automatic modal closing triggered even with the new changes. It is only an issue when there are tickets on the side bar still loading, which once they load trigger all modals to be closed. This however is already the same behaviour without my changes.

(It is caused by this line hiding all modals https://github.com/zammad/zammad/blob/8ea22f28ebb439e6b4b5ac330032199497153594/app/assets/javascripts/app/controllers/ticket_zoom.coffee#L340; triggered by the finished loading ticket zoom controller in the background being hidden and calling the above https://github.com/zammad/zammad/blob/8ea22f28ebb439e6b4b5ac330032199497153594/app/assets/javascripts/app/lib/app_post/task_manager.coffee#L325-L327)

If it is not this particular bug, then please let me know how to reproduce it, as I don't have any automatic modal closing happening for me otherwise.

Of course the above is an issue by itself and should preferably be fixed; even if it is already an existing bug. :smile:

Stubenhocker1399:
I've pushed a change that seems to fix that particular issue while still closing any modals correctly as described in #2104.

If there are any further issues let me know. :smiley:

fliebe92:
Good morning @Stubenhocker1399 😎 

Thanks for your time to check my last comment and do some further fixing. I was not able to reproduce the described behavior without your changes, but maybe it was a timing issue or something else. I am going to check your changes soon to get this merged. Stay tuned!

Thanks for your patience and for improving Zammad! ❤️

fliebe92:
Hey @Stubenhocker1399, your changes are fine and merged to `develop` by now, `stable` will follow soon!

Thanks again for your pull request, we really appreciate it! ❤️

PR #4182, opened 2022-07-20
Feature: make action mailer configurable via environment variable

moritz-makandra:
When I run Zammad in containers [zammad/zamamd-docker-compose](https://github.com/zammad/zammad-docker-compose) the password reset mails can not be sent, because there is no MTA in the container. To use a different MTA it would be great to have an environment variable `SMTP_URL` which configures the MTA for `ActionMailer`.

I added some lines which set up the `action_mailer.smtp_settings` if there is an environment variable `SMTP_URL`.


moritz-makandra:
The settings can be configured in Settings > Channels > Email > Email notification

PR #4191, opened 2022-07-27
Fixes #849 - Allow additional usage of Ticket Number in (Zoom) URL

### LINKED ISSUE #849: Allow additional usage of Ticket Number in (Zoom) URL
### Infos:

* Used Zammad version: 1.3.x
* Operating system: Debian
* Browser + version: Chrome

### Expected behavior:

* Ticket Number in URL is correct

### Actual behavior:

* Ticket Number (ID) in URL is different to the ticketnumber

* The Ticket ID 279 is missing:
> SELECT id, number, title FROM tickets ORDER BY id;
 277 | 35277  | Probleme mit Outlook
 278 | 35278  | Probleme mit Outlook TEST
 280 | 35279  | Defekte Festplatte im Server
 281 | 35280  | Fristverlängerung online
 282 | 35281  | NB191 Tunnel zu iSource offline

image

We haven't deletet any ticket manually or somthing else.
Any idea how to fix this?

Stubenhocker1399:
Hi! :smile: 

I've implemented a ticket linking functionality to Zammad as requested in this [community forum post](https://community.zammad.org/t/open-detail-page-for-ticket-with-ticket-number-via-url/9197).

Visiting #ticket/number/xxxxx automatically redirects to the correct zoom view.

I wasn't able to test my tests as I don't have Selenium set up, hope they work as expected. :sweat_smile:

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_number.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_number.coffee
@@ -8,6 +8,25 @@ class TicketZoomHookRouter extends App.ControllerPermanent
     if (target_ticket)
       window.location.replace(target_ticket.uiUrl())
     else
-      window.location.replace("#ticket/zoom/0")
+      App.Ajax.request(
+        type:  'GET'
+        url:   "#{@apiPath}/search"
+        data:
+          query: ''
+          condition: [{ 'ticket.number': { operator: 'is', value: params.ticket_number } }]
+          limit: 1
+        processData: true
+        success: (data, status, xhr) =>
+          for item in data.result
+            continue if item.type isnt 'Ticket'
+            App.Collection.loadAssets(data.assets)
+            target_ticket = App.Ticket.findNative(item.id)
+            if (target_ticket)
+              window.location.replace(target_ticket.uiUrl())
+            return
+          window.location.replace('#ticket/zoom/0')
+        error: =>
+          window.location.replace('#ticket/zoom/0')
+      )
 
 App.Config.set('ticket/number/:ticket_number', TicketZoomHookRouter, 'Routes')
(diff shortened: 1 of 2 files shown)


rolfschmidt:
rel https://github.com/zammad/zammad/issues/849

rolfschmidt:
Hi @Stubenhocker1399 ,

I think this code will only work for tickets which are already present in the frontend UI. To solve this you would need to do an ajax request (`App.Ajax.request`) to the `/v1/tickets/search` endpoint.

here an example of an ajax request,
https://github.com/zammad/zammad/blob/1aa8c9e43a2f403bc0688a49894f097195b8bd69/app/assets/javascripts/app/lib/app_post/global_search.coffee#L33

here the endpoint 
https://github.com/zammad/zammad/blob/1aa8c9e43a2f403bc0688a49894f097195b8bd69/app/controllers/tickets_controller.rb#L455

You could try to use a post request with a condition hash (then it also works without elastic search):

``` 
condition: { 'ticket.number' => { operator: 'is', value: '71001' } }
```

grafik


this should return the ticket id and the assets which can be loaded via (`App.Collection.loadAssets(data.assets)`) and then you can find your ticket and redirect.

Stubenhocker1399:
@rolfschmidt Regarding the
> I think this code will only work for tickets which are already present in the frontend UI.

is there some easy way to achieve that state? Because even when disabling all overviews the current code worked fine for me.

I'm not too familiar with what decides a ticket being "present in the frontend UI". Seems to me that it'd be preferable to use this data source first than an extra request if possible.

SamuelPlentz:
Thanks to everyone involved to make that happen in Zammad 5.2.2! (#849)

FYI: To use this really nice feature you have to use:
```
/#ticket/zoom/number/10123
```

There were several proposals of the syntax that don't work:
```
/#ticket/number/10123
/#ticket/external/10123
```

PR #4197, opened 2022-08-02
Maintenance: Optimize GitHub Actions configuration.

mgruner:


.github/workflows/ci.yaml
```diff
@@ -1,45 +1,50 @@
...
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgres
-        ports:
-        - 5432:5432
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
-    strategy:
-      matrix:
-        dbs:
-          - mysql
```
monotek, line 28: so we're not testing mysql anmyore?
why?

mgruner: @monotek the definitive source of CI truth is our GitLab instance. There's much more tested than here on GitHub already with the present state, including MySQL and MariaDB. We cannot try to reproduce that here as it requires too much computation power.

So for GitHub, my goal would be to ensure basic standards and quality checks for incoming Pull Requests on GitHub. For that, MySQL coverage would not be mandatory IMHO, but I'm open about that part. All PRs will also go through the full GitLab CI pipeline before they can get merged.

monotek: Imho thats a bit intransparent for people creating PRs here.
Can we maybe link the Gitlab CI runs or at least add the output to the checks list?
I've seen something like this for terraform enterprise runs, showning the ouput in the "Github checks list" even though a the link to Gitlab might need a login, one might at least see what check fails.

mgruner: That is true, this is not transparent. Has been like that for quite a long time now, though.
Do you have any recommendations for how to create such an integration?

monotek: Would need to check out if gitlab provides something like this.
But i'm a little short in time at the moment.

mgruner: Also, just realized that the PRs on GitHub don't exist in GitLab, only after a developer creates a branch from them and pushes it to GitLab. So we don't have test results to show initially... Not optimal, but I don't have an alternative right now.

mgruner:
Merged in 5289db43.

PR #4200, opened 2022-08-05
add dependabot

monotek:
* add dependabot for:
  * github actions
  * ruby

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


MrGeneration:
Zammad already uses dependabot internally and does check for bundler and npm.
https://github.com/zammad/zammad/blob/develop/.gitlab/dependabot.yml#L3-L13

monotek:
Removed bundler, so only Github actions are checked.

mgruner:
Hi @monotek, thanks a lot for your suggestion. We recently switched from dependabot to renovatebot, and I will add updating of GitHub actions there as well.

PR #4211, opened 2022-08-10
Feature/get roles from saml

ClFeSc:
Hi there!

We've implemented role mappings for SAML authentication as described in [this community post](https://community.zammad.org/t/feature-request-saml-group-mappings/9204/2). This is our first time contributing to this project, so feel free to point out any aspects we've not handled as intended.

### How to use the SAML role synchronization?

1. Set up SAML authentication as you would normally (see [the documentation](https://admin-docs.zammad.org/en/latest/settings/security/third-party/saml.html)).
1. Add a mapper to the client that you use with your Zammad instance (for example, a Keycloak instance). This mapper must have the `Mapper Type` `Role list`, the `Role attribute name` must be `Role`, `SAML Attribute NameFormat` must be set to `Basic`, and `Single Role Attribute` must be disabled.
1. Make sure that `role_list` is added to the assigned default client scopes.
1. To synchronize roles, the same role names must exist in your client and in Zammad. Note that additional roles that may be assigned to users are not harmful, there just must be at least one matching role.
1. Enable role synchronization in Zammad's SAML settings.

### DIFF
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -37,6 +37,12 @@ class Authorization < ApplicationModel
         end
       end
 
+      if Setting.get('auth_saml_credentials')['role_sync']
+        # set roles according to the SAML response
+        saml_roles = Role.get_role_names_from_saml(hash)
+        user.role_ids = Role.get_matching_role_ids(saml_roles) if saml_roles
+      end
+
       # update image if needed
       if hash['info']['image'].present?
         avatar = Avatar.add(
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -84,6 +84,46 @@ returns
 
 =begin
 
+get array of role names provided by the SAML response hash, returning nil when roles are not provided in hash
+
+  Role.get_role_names_from_saml(saml_hash)
+
+returns
+
+  [roleName1, roleName2, ...]
+
+=end
+
+  def self.get_role_names_from_saml(hash)
+    all_raw_info = hash.dig(:extra, :raw_info)&.all
+    all_raw_info.nil? ? nil : all_raw_info['Role']
+  end
+
+=begin
+
+intersect the input role names with all role names in Zammad and return the ids of matching roles
+raises an UnprocessableEntity Exception when no matching roles exist
+
+  Role.get_matching_role_ids(input_roles)
+
+returns
+
+  [roleId1, roleId2, ...]
+
+=end
+
+  def self.get_matching_role_ids(input_roles)
+    matching_role_ids = []
+    input_roles.each do |input_role_name|
+      zammad_role = Role.find_by(name: input_role_name)
+      matching_role_ids << zammad_role.id if zammad_role
+    end
+
+    matching_role_ids
+  end
+
+=begin
+
 get signup role ids
 
   Role.signup_role_ids
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -368,6 +368,12 @@ returns
 =end
 
   def self.create_from_hash!(hash)
+    matching_role_ids = nil
+    if Setting.get('auth_saml_credentials')['role_sync']
+      # set roles according to the SAML response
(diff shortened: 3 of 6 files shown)


MrGeneration:
Please remove all custom changes to files in `i18n/* ` and ensure to follow the readme of said directory:
https://github.com/zammad/zammad/tree/develop/i18n#updating-the-source-string-catalog

Translations are handled via weblate, adding them this way is not supported. Thank you

ClFeSc:
> Please remove all custom changes to files in `i18n/* ` and ensure to follow the readme of said directory: https://github.com/zammad/zammad/tree/develop/i18n#updating-the-source-string-catalog
> 
> Translations are handled via weblate, adding them this way is not supported. Thank you

Thanks for the tip, changed it.

ClFeSc:
By the way: Is it intended behavior that when a user without any roles logs in, the user gets redirected a lot, and ultimately is on a page that looks like this:
grafik
This was the reason why we disallowed logging in without roles (although this assigns a user to the customer role, which may not be the intended behavior when you unassign all roles from a user).

MrGeneration:
> By the way: Is it intended behavior that when a user without any roles logs in, the user gets redirected a lot, and ultimately is on a page that looks like this:

Looks like a bug you're introducing there.
Technically a user can be created without role assignment. The role assignment stays empty without issues, Zammad would fallback to the default signup role (which is customer permissions).
That's as far as I can provide help here.

ClFeSc:
> > By the way: Is it intended behavior that when a user without any roles logs in, the user gets redirected a lot, and ultimately is on a page that looks like this:
> 
> Looks like a bug you're introducing there. Technically a user can be created without role assignment. The role assignment stays empty without issues, Zammad would fallback to the default signup role (which is customer permissions). That's as far as I can provide help here.

On second thought, you're completely right. We've changed it back so that it uses the default role (`Customer`) when no matching roles are provided.

dominikklein:
@ClFeSc Thanks for your contribution. We will clarify this topic beginning of next year and check the next steps.

dominikklein:
@ClFeSc We talked internally about this topic and in general it's an interesting feature, but we didn't see this request very often. Nevertheless in the long run it makes sense to have such a possibility, but the current solution in this PR is very technical and without any UI for a defined mapping (similar to the LDAP configuration).
So in the current state, we can not merge this PR in the codebase, but feel free to come up with an idea that will match more with the needs (UI for mapping).

thegcat:
Just to chime in that we'd be interested in importing roles from SAML or OIDC. We currently still have LDAP as a data backend for our SAML/OIDC authenticator, but would prefer switching away from LDAP entirely. Group sync in certain software, notably Zammad, is holding us back with this project.

ClFeSc:
> @ClFeSc We talked internally about this topic and in general it's an interesting feature, but we didn't see this request very often. Nevertheless in the long run it makes sense to have such a possibility, but the current solution in this PR is very technical and without any UI for a defined mapping (similar to the LDAP configuration). So in the current state, we can not merge this PR in the codebase, but feel free to come up with an idea that will match more with the needs (UI for mapping).

@dominikklein Thanks for looking into this. We don't think Zammad is the right place for such a UI. Mapping roles to users is the job of the IDP (e.g. Keycloak), where user-friendly UIs already exist. Doing this again here would only increase the amount of maintenance required and could confuse users as to where to configure the mapping. The only place we could imagine having a UI in Zammad itself is for configuring claim names. Do you mean this aspect?

dominikklein:
@ClFeSc Thanks for your input. I will talk again internally about the topic because maybe the last decision was not done from the best perspective.

ClFeSc:
What is the current situation here? We're currently maintaining our own fork with these changes which is a far-from-perfect situation.

dominikklein:
@ClFeSc Thanks for the input, we will handle this in the future with the following issue: https://github.com/zammad/zammad/issues/4943

PR #4268, opened 2022-09-24
Bump ruby 312

ghost:
Bump 'Ruby-Lang' version from 3.0.4 to 3.1.2


mgruner:
Hi @chetanmittaldev, thanks for your suggestion. Our version policy is that we use the second most recent version of Ruby and Rails that is available, so we will update to Ruby 3.1 later.

PR #4310, opened 2022-10-20
Feature/assets precompile upon upgrade

burnsy1882:
I would like to add a function to precompile the assets after a package upgrade for those that have some customizations to CSS and/or JS. I have added a new function that calls the rake command before restarting the server after an upgrade to the DEB/RPM package. Upon first install, the command will also probably be called, but should not matter as the assets would be the same.

### DIFF
--- a/contrib/packager.io/functions
+++ b/contrib/packager.io/functions
@@ -309,6 +309,11 @@ function update_or_install () {
   chmod 600 ${ZAMMAD_DIR}/config/database.yml
 }
 
+function assets_precompile () {
+  echo "# Precompiling assets"
+  zammad run rake assets:precompile
+}
+
 function set_env_vars () {
   zammad config:set RUBY_MALLOC_ARENA_MAX=${ZAMMAD_RUBY_MALLOC_ARENA_MAX:=2}
   zammad config:set RUBY_GC_MALLOC_LIMIT=${ZAMMAD_RUBY_GC_MALLOC_LIMIT:=1077216}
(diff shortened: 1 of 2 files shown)


mgruner:
@dominikklein @rolfschmidt we should clarify this with @MrGeneration. Right now the design is that the generated assets are part of the bundled package, and there is no dependency to NodeJS on a production system. On the other hand, the Zammad package does also include a NodeJS binary. On first invocation, it would have to download all JS packages though. This will definitely fail on systems that have no internet connection available.

MrGeneration:
This is rather a PM topic sorry. I can give opinion but not decide. I don't care too much (except for update delay for compiling), I'll just add it to the docs if needed.

mgruner:
@dominikklein I also think this should go to PM for clarification. Do we support installations with local customizations like this?
I'd say if we do this, we may want to somehow detect that packages are installed, and only then run `assets:precompile`?

dominikklein:
@mgruner You mean with local customizations really code changes in existing files or additional JS files and/or css files inside of `app/assets/stylesheets/custom`?

MrGeneration:
Instead of detection one could also use environments to allow admins to decide if they want specific things or not.
Just as an idea. Should default to no because precompiling takes a lot of time.

dominikklein:
@burnsy1882 Thanks for your idea :-)
We had an internal discussion with our product team about this topic and currently the usage of this customization is very low, so we plan currently not to spend time on this topic here. The current idea of improvement in the PR will trigger always a precompile, but this should really only happen when needed (e.g. customization exists).
For the future, the plan is to improve the handling of packages in this context, this will then maybe also lead to improvements in general customizations.

PR #4326, opened 2022-10-27
Feature: Add explanation above ticket note

fflorent:
This pull request implements a suggestion someone else also had in mind:
https://community.zammad.org/t/usability-confusion-about-whether-a-note-is-locked-or-unlocked/1345/5

Also I hope it will solve confusion complaints about internal / public notes: https://community.zammad.org/t/internal-public-confusion/6239

I assumed this is a minor change and no tests were needed. Don't hesitate to ask me to add some if you disagree.

Feedback of course welcome! Thanks in advance.

internal note
public note

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_action/internal.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_action/internal.coffee
@@ -11,7 +11,7 @@ class Internal
       }
     else
       actions.push {
-        name: ('set to internal')
+        name: __('set to internal')
         type: 'internal'
         icon: 'lock'
       }
--- a/app/assets/javascripts/app/views/ticket_zoom/article_new.jst.eco
+++ b/app/assets/javascripts/app/views/ticket_zoom/article_new.jst.eco
@@ -39,8 +39,14 @@
     <% end %>
   </div>
   <div class="article-content bubble-gap">
+    <% if !_.isEmpty(@C('ui_ticket_add_note')): %>
+      <div class="article-visibility-text-wrapper" style="height: 0">
+        <% for type, note of @C('ui_ticket_add_note'): %>
+            <div class="article-visibility-text is-hidden" role="alert" data-type="<%= type %>"><%- @T(note) %></div>
+        <% end %>
+      </div>
+    <% end %>
     <div class="internal-border">
-
       <div class="input form-group">
         <div class="formGroup-label">
           <label for=""><%- @T('To') %></label>
--- a/app/assets/stylesheets/zammad.scss
+++ b/app/assets/stylesheets/zammad.scss
@@ -6894,6 +6894,20 @@ a.list-group-item.active > .badge,
   display: none;
 }
 
+.article-add {
+  .article-visibility-text {
+    display: none;
+  }
+
+  &.is-open.is-internal .article-visibility-text[data-type='internal-note'] {
+    display: block;
+  }
+
+  &.is-open.is-public .article-visibility-text[data-type='public-note'] {
+    display: block;
+  }
+}
+
 .textBubble {
   padding: 10px 20px;
   background: var(--background-secondary);
--- a/app/assets/stylesheets/zammad.scss
+++ b/app/assets/stylesheets/zammad.scss
@@ -6894,18 +6894,8 @@ a.list-group-item.active > .badge,
   display: none;
 }
 
-.article-add {
-  .article-visibility-text {
-    display: none;
-  }
-
-  &.is-open.is-internal .article-visibility-text[data-type='internal-note'] {
-    display: block;
-  }
-
-  &.is-open.is-public .article-visibility-text[data-type='public-note'] {
-    display: block;
-  }
+.article-add .article-visibility-text.is-hidden {
+  display: none;
 }
 
 .textBubble {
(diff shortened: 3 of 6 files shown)


app/assets/javascripts/app/controllers/ticket_zoom/article_action/internal.coffee
```diff
@@ -10,7 +10,7 @@ class Internal
       }
     else
       actions.push {
-        name: ('set to internal')
+        name: __('set to internal')
```
fflorent, line 14: This is unrelated but I found what seemed to be a mistake here.

app/assets/javascripts/app/views/ticket_zoom/article_new.jst.eco
```diff
@@ -39,8 +39,8 @@
     <% end %>
   </div>
   <div class="article-content bubble-gap">
+    <div class="article-visibility-text is-hidden"></div>
```
fflorent: I am not very used to Zammad, could you tell me if there is other places this text should also add the explanation text?

fflorent (commented):
(added side notes / questions to my own code)

MrGeneration:
Note that this PR requires PM feedback.

fflorent:
> Note that this PR requires PM feedback.

What PM means please?

MrGeneration:
> What PM means please?

Product Management. Does not directly affect you but Zammad as a product.

fflorent:
OK, sure, that's legit.

fflorent:
@MrGeneration Also I just noticed you added a `specification required` tag. Is there anything more I should provide to this PR? Or is this should be done by someone else?

dominikklein:
@fflorent thanks for your contribution. We talked inside our product team about the topic and in general we like the idea of adding this information for notes. The problem is that the topic is not such easy, because it could be that a trigger is configured in the background so that the note is not only visible on the web interface, instead it will also be sent by mail. 

The idea is to have it implemented like the current `ui_ticket_create_notes` setting: https://docs.zammad.org/en/latest/admin/console/hidden-settings.html#adding-a-warning-to-the-ticket-creation-process
So the user can add a custom message on their own for a public/internal note.
When you are able to change the PR in this direction, we would like to merge this.

fflorent:
@dominikklein @MrGeneration I made changes with the hope that's what you got in mind.

Thanks in advance for your review

fflorent:
Is there anything I can do to push forward this PR?

Thanks in advance and Merry Christmas!

mantas:
@dominikklein we allow `private` for phone call writeups and emails as well. However, with this implementation, it shows same message for any type. Showing `private note / public note` for phone and email is a bit ambiguous.

I'd suggest to:

a) add private/public pair for each article type.

b) have more generic text in sample, avoiding word `note`

Did you cover this in previous PM discussion?

app/assets/stylesheets/zammad.scss
```diff
@@ -6894,6 +6894,20 @@ a.list-group-item.active > .badge,
...
+
+  &.is-open.is-internal .article-visibility-text[data-type='internal-note'] {
+    display: block;
+  }
+
+  &.is-open.is-public .article-visibility-text[data-type='public-note'] {
+    display: block;
+  }
+}
+
```
mantas: @fflorent this looks like a very fragile way to handle this.

I'd rather extend [App.TicketZoomArticleNew#setArticleInternal](https://github.com/zammad/zammad/blob/5c1cbb966ea6e4a19414d6b6cfb95f7fa0add254/app/assets/javascripts/app/controllers/ticket_zoom/article_new.coffee#L367) along the spirit of [App.TicketZoomArticleNew#setArticleTypePre](https://github.com/zammad/zammad/blob/5c1cbb966ea6e4a19414d6b6cfb95f7fa0add254/app/assets/javascripts/app/controllers/ticket_zoom/article_new.coffee#L376)

fflorent: Why would that be fragile?

mantas: Please take this with a grain of salt since it's project-specific :) It may be totally fine (and even preferred) in another codebase where it's CSS-first.

We usually do this sort of switching in Coffeescript event callbacks. Finding elements based on `data-*` or classes and then manipulating them. This allows to put all UI manipulations to related callbacks.

When a part of this goes directly into CSS, somebody editing this area must remember to also look into CSS. Aside from just checking out related callbacks.

If this was relying on new article data type and visibility, we'd have seen early on that there're 3 types all of which can be public or private. Meanwhile this approach allows to easily use `-note` in context of phone/email. Which is a bit ambiguous. `note` is only 1 of 3 available `article` types.

On top of that, this creates a lot of repetitive CSS. We'd have to have 6 nearly identical CSS selectors to cover all types/visibility pairs.

fflorent: Happy new year!

Thanks :). I didn't take personally your remark (it could have been truly a bad piece of code), but I could not understand what were the reasons that led you state the code was fragile.

Your point is fair, I take into account this remark

fflorent: I think it's done

fflorent:
The CI fails again, but it seems that's not due to my patch.

YetAnotherGerrit:
> I'd suggest to:
> a) add private/public pair for each article type.
> b) have more generic text in sample, avoiding word `note`

I'd personally prefer to go with option a) and no default values at all. Means empty Setting-variables by default. Like it is with:
https://docs.zammad.org/en/latest/admin/console/hidden-settings.html#adding-a-warning-to-the-ticket-creation-process

dominikklein:
@mantas we want to go with option a)

dvuckovic:
@fflorent Merged, thanks for your contribution!

Note that the new setting name was changed slightly into `ui_ticket_add_article_hint`.

We also adjusted the style of the hint a bit so it aligns well with the rest of the UI:

Screenshot 2023-02-08 at 16 36 16
Screenshot 2023-02-08 at 16 36 08

PR #4338, opened 2022-11-04
Fix spacing between asterisks and text. Fix #4337

### LINKED ISSUE #4337: Remove spaces before ** in SLA description text
### Infos:

* Used Zammad version: 5.2.3-1666860729.1f5facea.focal
* Installation method (source, package, ..): packages
* Operating system: Ubuntu 20.04
* Database + version: Postgresql 14+232.pgdg20.04+1
* Elasticsearch version: elasticsearch-oss 7.10.2
* Browser + version: any

### Expected behavior:

* Every setting item having its own title

### Actual behavior:

* Some settings share the same title

https://github.com/zammad/zammad/blob/63fecf55ca3b8291396699b3762c748cb969aabb/app/assets/javascripts/app/models/sla.coffee#L21-L25

Is it possible to remove spaces before and after the `**`s? As it's now, introduces spaces between words and punctuations. Other strings have asterisks attached, so it shouldn't be, but I don't know if the description is using a different rendering method...

immagine

maxxer:

### DIFF
--- a/app/assets/javascripts/app/models/sla.coffee
+++ b/app/assets/javascripts/app/models/sla.coffee
@@ -22,9 +22,9 @@ class App.Sla extends App.Model
   ]
 
   @description = __('''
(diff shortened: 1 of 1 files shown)


mgruner:
Nice catch! I found another related issue in the data privacy page. Fix landing soon.

PR #4343, opened 2022-11-07
fix rails env in docker container

monotek:
Signed-off-by: André Bauer <andre.bauer@staffbase.com>

* fix rails env in docker container
* add test of contianer via docker-compose


monotek:
close in favour of gitlab pr

PR #4382, opened 2022-11-22
Remove inline image styling before showing modal

Stubenhocker1399:
Hi there :wave: 

I propose adding this change to strip the inline styling of images in articles so they can be shown in full size
when viewing them in the image modal.

Without this change:

With this change:

See [this related thread](https://community.zammad.org/t/remove-inline-image-styling-before-showing-preview/10229) for more info.

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_image_view.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_image_view.coffee
@@ -32,7 +32,6 @@ class App.TicketZoomArticleImageView extends App.ControllerModal
 
   content: ->
     @image = @image.replace(/view=preview/, 'view=inline')
-    @image = @image.replace(/style="[^"]*"/, '')
     "<div class=\"centered imagePreview\">#{@image}</div>"
 
   onSubmit: =>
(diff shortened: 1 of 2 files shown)


MrGeneration:
You may want to have look at Zammad 5.3 first, there was changes to exactly this modal.

Stubenhocker1399:
The example gifs are from an older Zammad 5.2.x, but I did test it on the newest version.

Current develop branch 56bc08be449f87c5ebcd2e4dc44dc2d903accdeb without this change produces this modal:
image


With the change:
image

Stubenhocker1399:
To recreate it, paste a large image ([this one for example](https://raw.githubusercontent.com/zammad/zammad/60cb04ee05e4cb45622aaf4b3df2638b0e41ab7d/spec/fixtures/files/image/large.png)) into the text field and scale it down by clicking on it and using the handles and sending it off.

dominikklein:
@dvuckovic Maybe you can check the idea behind this PR and when we have some insides we can schedule a call with the PO.

dvuckovic:
@Stubenhocker1399 Thanks for your proposal!

While your code does work as expected, I'm not really sure if it's the best approach.

It might be better to have this style forced by global CSS, for example [around here](https://github.com/zammad/zammad/blob/develop/app/assets/stylesheets/zammad.scss#L10125).

Because in CSS inline styles win the specificity battle, something like this might do the trick:

```diff
   img {
-    max-width: 100%;
-    max-height: calc(100vh - 218px);
+    width: auto !important;
+    max-width: 100% !important;
+    max-height: calc(100vh - 218px) !important;
   }
```

Let us know what you think!

Stubenhocker1399:
Oh I absolutely agree, that is much cleaner! (Sorry for the slow response, happy new year! :slightly_smiling_face:)

dvuckovic:
@Stubenhocker1399 Merged, thanks for your contribution!

PR #4388, opened 2022-11-24
Fixes #4387 - Colors open ticket indicator when count is 3

### LINKED ISSUE #4387: Open ticket indicator does not get colored when ticket count is 3 with enabled colored indicators
### Infos:

* Used Zammad version: 5.3.0-1669114633.0266d173.focal and develop 806111fc2772b4d4dd958d2dc8cff453efda94a3
* Installation method (source, package, ..): package and source
* Operating system: Ubuntu 20.04.5 and Arch Linux
* Database + version: irrelevant
* Elasticsearch version: irrelevant
* Browser + version: Google Chrome Version 107.0.5304.110 (Official Build) (64-bit)

### Expected behavior:

* Setting `ui_sidebar_open_ticket_indicator_colored`  colors side bar open ticket count for any number above 2

### Actual behavior:

* Setting `ui_sidebar_open_ticket_indicator_colored`  does not work for an open ticket count of exactly 3, continues to work at 4

* Set `Setting.set('ui_sidebar_open_ticket_indicator_colored', true)` in rails console
* Have customer with exactly three open tickets
* Check customer icon on sidebar to see counter not being colored

image
image
image
image

On a sidenote: the warning color could probably be improved for readability here in the light theme.

Stubenhocker1399:


dominikklein:
We should also check if the functionality is correctly implement in: `app/assets/javascripts/app/controllers/agent_ticket_create/sidebar_customer.coffee`.

Stubenhocker1399:
From my understanding it is correct, the only difference being it already coloring the indicator when the ticket count is one, rather than two. This makes sense to me as you would be warned that the customer you are creating a ticket for already has one open.

mantas:
Hi @Stubenhocker1399 

This is now merged as https://github.com/zammad/zammad/commit/5c1cbb966ea6e4a19414d6b6cfb95f7fa0add254 . Thank you for your contribution! 👍

PR #4440, opened 2022-12-21
Feature: Add STARTTLS support for LDAP connections

Mikrz:
The LDAP Integration does not currently support STARTTLS as an encryption method, and thus is unsupported by many common LDAP setups as LDAPS is deprecated as of [2003 per RFC3494](https://www.rfc-editor.org/rfc/rfc3494), with STARTTLS being built into the LDAPv3 standard.

I have added extremely basic support for STARTTLS in LDAP connections. This came from a manual patch while attempting to set up Zammad in my company infrastructure while migrating from OTRS. My Ruby knowledge started with this work, and so I apologize if there are better ways of implementing this solution, however I feel this solution is the simplest and most intuitive method without adding any UI modifications.

The encryption method and default port is determined by the protocol, which was already being split out from the host_url to determine if SSL should be used.


abotelho-cbn:
This would be hugely beneficial to us! Has there been any progress on this? Seems like a pretty simple PR.

MrGeneration:
Please don't bump. It doesn't help at all.

dominikklein:
@Mikrz Thanks for your contribution, it's a really good idea to add the `STARTTLS` support for LDAP. Nevertheless, to add this to the core product, we need some better solutions related to the UI.

So the idea would be to switch the current "SSL" field (no/yes) to an "Encryption/Security" field with the following options: `no`, `SSL`, and `STARTTLS`).
With this change, we could remove the "Protocol" (`ldaps://` or `ldap://`) from the host value, because it's no longer needed to detect the encryption.
For the existing installations, migration would be needed.

dominikklein:
Because of the needed UI changes, this PR can currently not be merged, we will handle the topic inside the following issue: https://github.com/zammad/zammad/issues/4746
@Mikrz  For you will receive some credits for your input here :-)

PR #4512, opened 2023-03-02
Email signatures trim trailing whitespace in text part fix

### LINKED ISSUE #2126: Email signatures trim trailing whitespace in text part
### Infos:

* Used Zammad version: 2.4.0
* Installation method: package
* Operating system: Ubuntu 16.04
* Database + version: MySQL

### Expected behavior:

* having no trailing whitespace trimmed in text emails

### Actual behavior:

* trailing whitespace is trimmed in text part of emails - thus the signature separator `-- ` (dash, dash space) gets into a `--` only and is not recognized by all email clients anymore.

* write an answer in Zammad like this:

```
Hello

-- 
This is the signature
```

Then the text part of the email should contain the `-- ` and not only the `--`.

**Edit** *sigh* also Github drops the space and thus both examples look the same. Check out the source to see the difference. It is about DASH DASH SPACE vs DASH DASH.

thojo0:
Fixes #2126
Decision was made because of the same regex-detection on line 389.
Test: send an E-Mail by ticket creation with signature


dominikklein:
@rolfschmidt and @fliebe92 any thoughts about this?

fliebe92:
> @rolfschmidt and @fliebe92 any thoughts about this?

The regular expression in general looks fine, but, I would like to see a test here.

fliebe92:
Hey @thojo0, welcome to the Zammad repository 🥳 I would like to say thanks for your pull request.

Is it possible for you to add a test case to prove that the reported issue is fixed with your change? This would be awesome.

We really appreciate your support, looking forward to hearing from you. 😄

thojo0:
How should I do the test case?
Write how to do it or/and what happend when I did it?

fliebe92:
> How should I do the test case?
> Write how to do it or/and what happend when I did it?

I was speaking about a unit test, sorry if I was unclear. We do have `spec/lib/core_ext/string_spec.rb` where your changed `html2text`is checked (in the matter of "does it work in the way we expect it to work?"). It would be nice if you could add at least one test case there.

fliebe92:
Hey @thojo0 and thanks for being patient with us.

We checked your changes and we decided to go a slightly different way. However, your PR will be handled soon.

Thanks for your support!

fliebe92:
Hey @thojo0, your changes are now added to `develop`: https://github.com/zammad/zammad/commit/9331892e8535cc42b269b57fbb8875d9ccbffe85

Thanks a lot, we're looking forward to more PRs from you!

PR #4518, opened 2023-03-11
Add S3 backend as Storage option Closes #660

### LINKED ISSUE #660: Add new AWS S3 as storage backend
Currently, database and filesystem is allowed for attachment storage. This is a request to add AWS S3 as a new backend

hggh:
I have added S3 storage backend support - I'm not a Rails person. I was not able to create a settings page that will only show the S3 access key, secret key and region configuration settings if Storage is S3. 

If you wish such a page, please give me some hints to do :)

I have already added some docs but not created yet a PR: https://github.com/hggh/zammad-admin-documentation/commit/79cd8aeb8c7c083c382c133bfd3fc68991bd2959

Closes: https://github.com/zammad/zammad/issues/660

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -170,6 +170,9 @@ gem 'viewpoint', require: false
 # integrations - S/MIME
 gem 'openssl'
 
+# storage - S3 backend
+gem 'aws-sdk-s3'
+
 # Translation sync
 gem 'byk', require: false
 gem 'PoParser', require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -94,6 +94,22 @@ GEM
     ast (2.4.2)
     autoprefixer-rails (10.4.13.0)
       execjs (~> 2)
+    aws-eventstream (1.2.0)
+    aws-partitions (1.717.0)
+    aws-sdk-core (3.170.0)
+      aws-eventstream (~> 1, >= 1.0.2)
+      aws-partitions (~> 1, >= 1.651.0)
+      aws-sigv4 (~> 1.5)
+      jmespath (~> 1, >= 1.6.1)
+    aws-sdk-kms (1.62.0)
+      aws-sdk-core (~> 3, >= 3.165.0)
+      aws-sigv4 (~> 1.1)
+    aws-sdk-s3 (1.119.1)
+      aws-sdk-core (~> 3, >= 3.165.0)
+      aws-sdk-kms (~> 1)
+      aws-sigv4 (~> 1.4)
+    aws-sigv4 (1.5.2)
+      aws-eventstream (~> 1, >= 1.0.2)
     binding_of_caller (1.0.0)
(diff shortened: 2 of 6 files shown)


hggh:
For the better settings, I have created a new branch for testing:
https://github.com/hggh/zammad/commit/84cdb957cc76d8a9d07587d2d8917ad8a017d8e1
Screenshot from 2023-03-14 20-31-35

I have used the proxy settings area, but the forms now automatic appears below my custom form and the drop-down is missing. :(

timnis:
Can this be enhanced to support custom S3 server like Minio? Basically adding server address field, if it's empty then refer to Amazon S3.

hggh:
> Can this be enhanced to support custom S3 server like Minio? Basically adding server address field, if it's empty then refer to Amazon S3.

good idea. I did a quick test and it works:
```
endpoint:         'http://localhost:9000',
 force_path_style: true,
```
I had to configure force_path_style to true, because I used as endpoint http://localhost:9000.

We can use ENV Variables to configure that "expert" setting or make to two new (in the system/storage provder overview) settings. What do you like? (i will work on that feature on the weekend)

Screenshot from 2023-03-20 19-35-14

martini:
Hi ho. I like this approach. I just wondering if somebody did any performance tests/compare to FS and DB here?

Would be really interesting (and just want to get a idea if it will scale on big systems or not)! :)

Thanks anyway for this nice work!

hggh:
> Can this be enhanced to support custom S3 server like Minio? Basically adding server address field, if it's empty then refer to Amazon S3.

@timnis added the options endpoint and path style access. The code now works with MINIO and AWS S3.

monotek:
Any reason this is stale?
I would like to see that comming too, as it would help to scale zammad even if file storage is used instead of db.
See for example the discussion in: https://github.com/zammad/zammad-helm/issues/196

Performance wise it should be ok, as s3 storage is used a lot nowdays for stuff like this.
Normaly only latency should be a slightly higher but that should'nt have to much of an impact for loading images and so on.

If this is considered a valid option storage support for other cloud providers like Azure (Blob storage) and Google (Cloud Storage) would be nice too, but that could be takkled later. 

I think AWS S3 and compatible providers like MinIO would be good first step.

mgruner:
@dominikklein @KoJie how to proceed here? This is not a small change.

dominikklein:
We internally already have some working state.
But this topic needs to be clarified in general.

monotek:
If this would work for /var dir too it would be realy nice and solve a lot of the problems we currently have with persistence & scaling.

Realy looking forward to see this soon.

tschaefer:
@hggh Thank you for the PR. We implemented S3-compatible services as a storage provider with the commit  https://github.com/zammad/zammad/commit/c66165f0b8cd3c967ff03b7b82be1f8381d291a5. We already had an open private PR for a longer time.

PR #4586, opened 2023-04-19
contrib/backup/functions: support backup of storage directory when it is a symlink

### LINKED ISSUE #4585: zammad_backup doesn't backup attachments if storage directory is a symlink
### Used Zammad Version

5.4.1

### Environment

- Operating system : CentOS Stream 8
- Database + version:  postgresql 14
- Elasticsearch version: N/A
- Browser + version: N/A

### Actual behaviour

In our Zammad installation, `/opt/zammad/storage` is a symlink to a directory on another disk. With this configuration, `zammad_backup` (from `contrib/backup`) doesn't process the attachment directory and produces a tar file that contains only the symlink.

### Expected behaviour

Attachements in the directory referenced by `/opt/zammad/storage` symlink saved in the tar file produced.

See above.

### Support Ticket

_No response_

yes

jouvin:
Fixes #4585

### DIFF
--- a/contrib/backup/functions
+++ b/contrib/backup/functions
@@ -155,9 +155,11 @@ function backup_files () {
       chown -R zammad:zammad ${ZAMMAD_DIR}/storage/
     fi
 
-      tar -C / -czf ${BACKUP_DIR}/${TIMESTAMP}_zammad_files.tar.gz ${ZAMMAD_DIR#/}/storage/ ${ZAMMAD_DIR#/}/var/
+    storage_dir=$(realpath ${ZAMMAD_DIR}/storage/)
+    tar -C / -czf ${BACKUP_DIR}/${TIMESTAMP}_zammad_files.tar.gz\
+	    ${storage_dir#/}
 
-      state=$?
+    state=$?
   fi
 
   if [ $state == '2' ]; then


jouvin:
Hi,

Any chance to see this PR merged? It's really annoying in our environment and the proposed change is minimal and should not break anything, IMO...

MrGeneration:
I'll have a look tomorrow, I'm sorry.

MrGeneration:
The suggested approach is good, thanks for your patience.
Right now I'm still verifying restoration which shouldn't be an issue.

One thing that bugs me is that the symlink is not saved. I do have a solution for that though.

jouvin:
Thanks! I must confess that I never tried the restore so something may goes wrong and may need an additional fix. As for saving the link, it is probably discussable: you may say that it is the site responsibility to provide the correct configuration, the restore taking the responsibility of putting the content inside the target directory, whether it is a directory or a symlink....

MrGeneration:
> As for saving the link, it is probably discussable: you may say that it is the site responsibility to provide the correct configuration, the restore taking the responsibility of putting the content inside the target directory, whether it is a directory or a symlink....

I do see your point.
The "issue" with the current restoration process is that the current approach automatically starts Zammad after restoring.
This brings potential issues when Zammad can't find it's files and may stay undiscovered for some time. For this reason I've added a detection method.

I'm already preparing the PR internally, may take me a couple more minutes. Stable backport will follow soon after that.

MrGeneration:
Thanks again for your contribution.
I'll be closing this PR due to our MR process - you can find my adjusted version with commit [https://github.com/zammad/zammad/commit/41cbf134ccd7d060e7e61021c363fa2b7ebc4e44

I am sorry, I seem to have messed up the PR on my end which is why your author information was lost. I'll try again on the stable approach. 🙏 It wasn't on purpose. :-(](https://github.com/zammad/zammad/commit/14074bfec82376e10ca33abd77db50797e910de4)

Your forked version can now safely be removed if you'd like.
I'll prepare the stable backport - packages (for stable) should be ready by tomorrow.

PR #4610, opened 2023-05-14
Remove rsync from docker files

monotek:
* get rid of /tmp/zammad and rsync to /opt/zammad in the docker container
  * migration of stuff from the filesystem should not be needed anymore as all data goes to postgres already

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,11 @@ FROM node:16.18.0-slim as node
 FROM ruby:3.1.3-slim AS builder
 ARG DEBIAN_FRONTEND=noninteractive
 ARG RAILS_ENV=production
-ARG ZAMMAD_TMP_DIR=/tmp/zammad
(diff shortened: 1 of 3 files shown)


dominikklein:
@mgruner something for you?

mgruner (approved):

mgruner:
@dominikklein I would suggest that @fliebe92 and @monotek handle this together when @monotek is back from vacation. There may be other related tasks such as backports to pre-release, stable and also changes in the zammad-docker-compose files.

monotek:
Imho this can be merged for develop now.
Should be cherry picked for stable when 6.0 comes out.
Helm chart and docker compose have to be changed afterwards.

fliebe92:
Merged to `develop` with https://github.com/zammad/zammad/commit/19c60c68665f4b2453879e00717ccf52fc01f3f1.

PR #4611, opened 2023-05-15
Draft: Replace 'gzip' with 'zstd' compression algorithm

dvanzuijlekom:
An idea originally conceived by @sigio and put to use in our local Zammad instances. We initially replaced `gzip` with `pigz`, to add multi-threading to the compressor, alleviating the workload of a single CPU and spreading it out over all of our cores instead. A feature which we *severely* missed in some of our *large* Zammad setups. We decided to further incorporate the Zstandard compression algorithm and to completely replace `gzip` with `zstd`, bringing the compression algorithm into the 21st century.

This feature should be considered as a (somewhat) *breaking change*, as the new backup script won't be able to restore the older gzipped files anymore with these changes in place. Either re-compress older backups, keep an archived copy of the `functions` script to be able to (more easily) restore older backups, or restore manually. It's not that difficult.

You'll probably want to clean up the old gzipped fileformat archives in your backup directory as well, as the new script only touches the new `.zst` files.

Package dependencies haven't been updated, as this is an optional script. `zstd` seems to be installed by default(?) on current versions of Debian and Ubuntu. It should be installed manually on distros such as RHEL/CentOS/etc.

### DIFF
--- a/contrib/backup/functions
+++ b/contrib/backup/functions
@@ -45,7 +45,7 @@ function delete_old_backups () {
   # Use -mmin to clean up files as -mtime (days) is too unprecise.
   #   However, add +60 minutes to allow for the backup script run time, so that
   #   backups created a few minutes more than a day ago are not already purged.
-  test -d ${BACKUP_DIR} && find ${BACKUP_DIR}/*_zammad_*.gz -type f -mmin +$(((60*24)*${HOLD_DAYS}+60)) -delete
+  test -d ${BACKUP_DIR} && find ${BACKUP_DIR}/*_zammad_*.zst -type f -mmin +$(((60*24)*${HOLD_DAYS}+60)) -delete
 }
 
 function get_db_credentials () {
@@ -112,11 +112,11 @@ backup_file_read_test () {
 
   if [ "${DEBUG}" == "yes" ]; then
     echo "I've been looking for these backup files: "
-    echo "- ${BACKUP_DIR}/${RESTORE_FILE_DATE}_zammad_files.tar.gz"
-    echo "- ${BACKUP_DIR}/${RESTORE_DB_DATE}_zammad_db.${DB_FILE_EXT}.gz"
+    echo "- ${BACKUP_DIR}/${RESTORE_FILE_DATE}_zammad_files.tar.zst"
+    echo "- ${BACKUP_DIR}/${RESTORE_DB_DATE}_zammad_db.${DB_FILE_EXT}.zst"
   fi
 
-  if [[ (! -r "${BACKUP_DIR}/${RESTORE_FILE_DATE}_zammad_files.tar.gz") || (! -r "${BACKUP_DIR}/${RESTORE_DB_DATE}_zammad_db.${DB_FILE_EXT}.gz") ]]; then
+  if [[ (! -r "${BACKUP_DIR}/${RESTORE_FILE_DATE}_zammad_files.tar.zst") || (! -r "${BACKUP_DIR}/${RESTORE_DB_DATE}_zammad_db.${DB_FILE_EXT}.zst") ]]; then
     echo -e "\n\n # ERROR - Cannot read on or more of my backup files. Double check permissions."
     echo -e " #-> RESTORE WAS NOT SUCCESSFUL"
     exit 3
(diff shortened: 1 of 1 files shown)


MrGeneration:
That's a very bad approach.
The backup and restore scripts are intentionally compatible from at least Zammad 3 to the current stable and that should be kept in mind. Additionally, `zstd` *is not* a default package which requires additional dependencies. This method also forces the migration path to have two exact same versions of Zammad which sometimes is not possible due to dropped OS support as example.

From my POV you have the following options:
- let the admin decide which approach to use (like we do with `FULL_FS_DUMP`: https://github.com/zammad/zammad/blob/develop/contrib/backup/config.dist#L10 )
- automatically detect the presence of `zstd` and dynamically use it (or not); Keep in mind that this approach *has to* have a fallback to `.gz` files in case the admin tries to restore an old system while having the dependencies installed

These changes also require documentational adjustments *and* a verification on all supported package installation distributions.

dvanzuijlekom:
> That's a very bad approach.

Thank you for your kind words, you're welcome. Also thank you for immediately deep-sixing my efforts and motivation to try and contribute to Zammad.

I see and understand where you're coming from though, I hadn't really thought about compatibility with (_very_) legacy versions, partly because that's not how _we_ use Zammad. We keep things in line with the released stable package/security updates and lifecycle, as recommended by Zammad. Furthermore, if one of our backups is over a couple of days old, it's usually not particularly useful to go back to that older version anyway, because we'd lose a lot of information accumulated in the meantime. It's also probably mostly due to the fact I had figured this script is in `contrib`, and not part of the Zammad core, so probably less important for support, portability and compatibility. I guess I was wrong.

> Additionally, `zstd` _is not_ a default package which requires additional dependencies.

I hadn't _really_ researched this, I just noticed it was installed on my own Zammad test setup on Debian 11.7, and I don't remember installing it explicitly myself, so assumed it would have probably either been installed by default or it came with a package/dependency installed for Zammad or its dependencies.

> From my POV you have the following options:
> 
>     * let the admin decide which approach to use (like we do with `FULL_FS_DUMP`: https://github.com/zammad/zammad/blob/develop/contrib/backup/config.dist#L10 )
> 
>     * automatically detect the presence of `zstd` and dynamically use it (or not); Keep in mind that this approach _has to_ have a fallback to `.gz` files in case the admin tries to restore an old system while having the dependencies installed

I can refactor/redesign the backup script to accommodate this type of behaviour, that'll take an investment of time from my part. I obviously don't have the entire CI/CD in place to test my changes with all of your supported platform flavours and/or Zammad versions, I can only test against the (current) stable version of Zammad _packages_ on Debian.

MrGeneration:
> Thank you for your kind words, you're welcome. Also thank you for immediately deep-sixing my efforts and motivation to try and contribute to Zammad.

Well sorry. I guess I went overboard with my wording. It wasn't supposed to slap you. Honestly.

> We keep things in line with the released stable package/security updates and lifecycle, as recommended by Zammad. Furthermore, if one of our backups is over a couple of days old, it's usually not particularly useful to go back to that older version anyway, because we'd lose a lot of information accumulated in the meantime.

Right and that's good, really.
Think of this use case, maybe you can understand my point better:
Migration from a very old version that hadn't been maintained well enough to a new host because the old one's no longer supported (this simply reduces overhead on migrating to a new system, this use case is technically aligned with our backup and restore documentation and allows clean starts (apart from the Zammad data of course).

Another use case you'd stop Zammad let's say in 2020 and due to legal reasons have to spin up your 2 year old backup for e.g. finance controls by your state. Yes, that could be solved by re-packing the backup. In that scenario the question would be if it's worth the effort to repack if you consider the required additional time. (that being said, if the scripts can detect that and adapt accordingly I'd personally be fine with that as long as old backups technically stay compatible)

> It's also probably mostly due to the fact I had figured this script is in contrib, and not part of the Zammad core, so probably less important for support, portability and compatibility.

I think the main issue is that it's the main method of backing up and restoring if you don't want to build something on your own. (Which you can always do of course)

> I can refactor/redesign the backup script to accommodate this type of behaviour, that'll take an investment of time from my part. I obviously don't have the entire CI/CD in place to test my changes with all of your supported platform flavours and/or Zammad versions, I can only test against the (current) stable version of Zammad packages on Debian.

That's fine. I am basically the CI at hand. (:

What I can do is offer my help and insights on this topic to improve the current as is (again, the idea isn't bad at all, especially when you want speedier backups and restores). I'll also volunteer to verify this against all package installations including cross migrations.

But I can understand if you don't want to invest further time.

mgruner:
@dvanzuijlekom thank you very much for your contribution! Looking forward to an updated version. Until now, I'll close the current PR. Feel free to reopen or open a new one when time is ready.

PR #4638, opened 2023-06-05
Uniform two factor string translation

maxxer:
The *Try another method* string is being translated with only the *T* capital all around except here

### DIFF
--- a/app/assets/javascripts/app/views/widget/two_factor_login/try_another_method.jst.eco
+++ b/app/assets/javascripts/app/views/widget/two_factor_login/try_another_method.jst.eco
@@ -1,5 +1,5 @@
 <h3 class="text-center">
-  <%= @T('Try Another Method') %>
+  <%= @T('Try another method') %>
 </h3>
 <div class="two-factor-auth">
   <div class="two-factor-auth-method-buttons">


mgruner:
@maxxer @dominikklein `Try Another Method` looks ok to me as title case which seems appropriate in a heading.

dominikklein:
I think the problem is that we have two places with the same string, but different first letters.

sheremet-va:
> @maxxer @dominikklein `Try Another Method` looks ok to me as title case which seems appropriate in a heading.

There is a CSS property for that (`text-transform: capitalize`), shouldn't we just use it to keep translation consistent? Or it doesn't work great with other languages?

dominikklein:
> There is a CSS property for that (text-transform: capitalize), shouldn't we just use it to keep translation consistent? Or it doesn't work great with other languages?

@dvuckovic @mgruner what do you think about this question?

From the current string perspective, I think title is missing it in this file: app/frontend/apps/mobile/pages/login/views/Login.vue:133

dvuckovic:
I guess we could use CSS for this, as it's presentational, sure. But using it just in one instance is probably not a good idea: it's confusing and prone to breakage. Probably a good point for the desktop rewrite though 😄

maxxer:
> There is a CSS property for that (`text-transform: capitalize`), shouldn't we just use it to keep translation consistent? Or it doesn't work great with other languages?

Indeed, that's not correct for the Italian language. I don't know about UI or graphical appearance, but the language itself doesn't contemplate the use of all capitalized words for a title, AFAIK.

mgruner:
I agree with @maxxer. We must leave this to the translators, that's why the source strings must be as correct as they can be.

dominikklein:
Then this place needs to be changed to Title-Case: `app/frontend/apps/mobile/pages/login/views/Login.vue:133` to align the two use cases.

dominikklein:
@maxxer Many thanks for bringing this up, it's now aligned with the following commit: https://github.com/zammad/zammad/commit/ecb60a73e8444d0611a491b140cc54b111d2c08e

PR #4681, opened 2023-06-24
docker entrypoint: add db option, improve error handling

valodzka:
1. POSTGRESQL_OPTIONS to allow passing options to postgresqul database like advisory_locking=false
2. Better check for user presence when for db:seed. In case of issue with User.any? seed will not be executed (detailed description can be found here https://github.com/zammad/zammad-docker-compose/issues/289).


valodzka:
@monotek is this tests failure related to the commit? From my understanding it is some issue with tests not with the commit.

mgruner (changes requested):
@valodzka thank you for your PR and sorry that it took so long to review.

I believe that the original code is actually correct. The case when something goes wrong with `User.any?` is when the database is present, but does not have schema information yet (container started for the first time). You can see this [in the log of the test run](https://github.com/zammad/zammad/actions/runs/5365278721/job/14739117951?pr=4681) where it raises this error:

```
zammad-init_1           | 2023-07-03T13:25:54.534192345Z /usr/local/bundle/gems/activerecord-6.1.7.3/lib/active_record/connection_adapters/postgresql/database_statements.rb:19:in `exec': PG::UndefinedTable: ERROR:  relation "users" does not exist (ActiveRecord::StatementInvalid)
```

In that case, the seeding should take place. Your change prevents this from happening, so I don't think we can go this way. Can you rebase your branch and remove this part, so that we can merge only the new variable?

valodzka:
@mgruner Ok, you are correct, I didn't considered this cases.

But still I'm pretty sure that seed can run second time in some cases. Most of it is harmless `create_if_not_exists` calls. But some like `Calendar.init_setup` or `Trigger.create_or_update `can change configured settings. I move extra check to seeds code. Is it better now?

mgruner:
Yes, thank you very much!

Issue #357, opened 2016-11-04
WhatsApp


martini:
Zammad wish of Marc Nilius

WhatsApp integration as Zammad channel like Twitter or Facebook

wagnst:
I would love to see Telegram integration. Should also be much easier to implement, as its oss.

thorsteneckel:
Hi @wagnst - thanks for your proposal. Could you please create a separate issue for this so we can discuss it there? Btw. I like it 👍

wagnst:
Hi @thorsteneckel, done thanks #362!

singular-error:
Telegram, Threema, What'sUp and Telephone contacts - they all have in common that they don't need a email-adress. So, what about a user-identification via one of some unique ids instead of email-adress? And of course, one user can have more ids / channels - but maybe no email.? I'd like to see this as a key feature.

vincentfuhler:
Whatsapp is very important to us

Johannes-GitHub:
Are there any news already?

thorsteneckel:
It's still prohibited by Whatsapp/Facebook to use their API in any other way than their own product(s). There are implementations out there but if they are used Whatsapp might suspend the user (you) and start a legal case against the provider (us?). So until there are no changes in their policy there won't be a solution/implementation for Whatsapp without risking too much. We'll keep all subscribers updated here in case of any changes. We are interested in an integration too but can't do anything about it. 😞

Wouter0100:
With the introduction of WhatsApp Business and their API, this could be a real thing - and would be VERY awesome. 

See https://www.whatsapp.com/business/api and https://developers.facebook.com/docs/whatsapp

Long story short: you'll have to run a Docker container which connects to WhatsApp servers and provides an API to get data from and to WhatsApp. Because the way WhatsApp is designed, something needs to be running to receive the messages - which is the Docker container.

Implementing the API in Zammad would be a real feature and a plus to implement WhatsApp without the risks of being blocked.

petervanbeugen:
I agree very much with Wouter. Having the WA functionality available in the short term would be fantastic.
I am willing to co-sponsor this !

java-guy:
Is there any news on the WhatsApp integration? 
We are interested in switching to zammad but WA is a must for us.

thorsteneckel:
[...]

PR #4735, opened 2023-07-28
Added passthrough to entrypoint for alternative `docker run` commands.

fleverest:
This PR will resolve three issues from the `zammad-docker-compose` repository: [Issue 355](https://github.com/zammad/zammad-docker-compose/issues/355
), [Issue 356](https://github.com/zammad/zammad-docker-compose/issues/356) and [Issue 358](https://github.com/zammad/zammad-docker-compose/issues/358). It will also make [PR #357](https://github.com/zammad/zammad-docker-compose/pull/357) unnecessary.

The fundamental issue is that running commands such as `docker exec -it zammad rake zammad:searchindex:rebuild` would fail, since the process does not initiate with the required environment variables (such as `ELASTICSEARCH_HOST`) which are defined (when absent) by the entrypoint script. This PR adds a passthrough to the end of the entrypoint script, which is a common practice. For example, this is used [in the postgres entrypoint](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L346).

With this change in place, we will be able to execute `elasticsearch-reindex` jobs et. al. via `docker run` commands. For example, with a `zammad-docker-compose` deployment we could run:
```
docker compose run zammad-scheduler rake zammad:searchindex:rebuild
```
as opposed to the old `exec` which currently doesn't work:
```
docker compose exec zammad-scheduler rake zammad:searchindex:rebuild
```

Thanks to the dev team for developing such an invaluable product!

### DIFF
--- a/contrib/docker/docker-entrypoint.sh
+++ b/contrib/docker/docker-entrypoint.sh
@@ -160,3 +160,6 @@ if [ "$1" = 'zammad-websocket' ]; then
 
   exec bundle exec script/websocket-server.rb -b 0.0.0.0 -p "${ZAMMAD_WEBSOCKET_PORT}" start
 fi
+
+# Pass all other container commands to shell
+exec "$@"
--- a/contrib/docker/docker-entrypoint.sh
+++ b/contrib/docker/docker-entrypoint.sh
@@ -110,11 +110,9 @@ if [ "$1" = 'zammad-init' ]; then
 
   # chown var
   chown -R zammad:zammad "${ZAMMAD_DIR}/var"
-fi
-
 
 # zammad nginx
-if [ "$1" = 'zammad-nginx' ]; then
+elif [ "$1" = 'zammad-nginx' ]; then
   check_zammad_ready
 
   # configure nginx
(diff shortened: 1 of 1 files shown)


mgruner:
@fliebe92 @dominikklein this looks good to me. Can you please have an eye on this one, and probably merge and back port this to `stable` in the next days?

It would be great if @monotek could also agree here.

mgruner:
@fleverest CI reports errors, can you have a look, please?

monotek (approved):
lgtm :)

fleverest:
@mgruner You're right - it's because the `zammad-init` command passes through to the end of the script. Changine each `if; fi` block to `if elif` should do the trick.

fliebe92:
Hi @fleverest. Thanks for your contribution to our repository, we accepted your pull request and the changes are available in `develop` already: https://github.com/zammad/zammad/commit/e4c6649865d111f3200eec35499eef414f43d5d2

The change is currently in the backporting phase to our `stable` branch as well.

Please note that due to our internal workflow, this pull request will be closed soon - no worries, there is no harm done by this. 😎 

I will post another update when `stable` is handled as well.

Thanks again!

fliebe92:
Changes for `stable` are done as well: https://github.com/zammad/zammad/commit/6dc6b0fae5089a7a81b1a5f107f972b9ca279247

Thanks for this really nice contribution. ❤️ 

Closing.

PR #4762, opened 2023-08-15
Update default overviews (seed)

YetAnotherGerrit:
Revamp of the default overviews based on my consulting experience.

### DIFF
--- a/db/seeds/overviews.rb
+++ b/db/seeds/overviews.rb
@@ -1,15 +1,39 @@
 # Copyright (C) 2012-2023 Zammad Foundation, https://zammad-foundation.org/
 
 overview_role = Role.find_by(name: 'Agent')
+
 Overview.create_if_not_exists(
-  name:      __('My Assigned Tickets'),
-  link:      'my_assigned',
+  name:      __('Unassigned tickets'),
+  link:      'all_unassigned',
   prio:      1000,
+  role_ids:  [overview_role.id],
+  condition: {
+    'ticket.owner_id' => {
+      operator:      'is',
+      pre_condition: 'not_set',
+    },
+  },
+  order:     {
+    by:        'created_at',
+    direction: 'ASC',
+  },
+  view:      {
+    d:                 %w[title customer group created_at],
+    s:                 %w[title customer group created_at],
+    m:                 %w[number title customer group created_at],
+    view_mode_default: 's',
+  },
+)
+
+Overview.create_if_not_exists(
+  name:      __('My new/open tickets'),
+  link:      'my_assigned',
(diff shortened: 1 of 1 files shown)


YetAnotherGerrit:
image

YetAnotherGerrit:
@martini informed me that changing the seed files might impact existing instances because translations will be removed from the translation tables. How can I prevent that?

Is there a file with deprecated strings, to keep them in translation?

dominikklein:
@mgruner maybe some questions for you about the old strings? Also, checking the overview names from the title case perspective?

db/seeds/overviews.rb
```diff
@@ -1,19 +1,16 @@
 # Copyright (C) 2012-2023 Zammad Foundation, https://zammad-foundation.org/
 
 overview_role = Role.find_by(name: 'Agent')
+
 Overview.create_if_not_exists(
-  name:      __('My Assigned Tickets'),
-  link:      'my_assigned',
+  name:      __('Unassigned tickets'),
+  link:      'all_unassigned',
```
mgruner, line 7: @YetAnotherGerrit Until now, we used [title case](https://en.wikipedia.org/wiki/Title_case) for the strings, as they usually appear on headings. Did you change that on purpose?

Regarding the old strings, you can probably just add a list of legacy strings at the end of the file with a leading comment line before them. The strings will do nothing there, but the extractor will find them.

mgruner (changes requested):

db/seeds/overviews.rb
YetAnotherGerrit: > Did you change that on purpose?

No, actually, it just felt more right to me. But I received multiple feedbacks already that they like "title case" more. I will update my PR back to title case. For me, the conceptional change is the important thing.

I will add the legacy string in the next commit as well. I will also update the translation catalog, so the CI-link runs successfully.

dominikklein: Can we bring this forward in some way?

martini:
I think we need to setup a refinement for this to clarify why this change is needed and what this change is improving (to have a common understanding). I'm open for a change but (at least for me) it's not clear what the origin problem is.

Note: We should avoid to change the default overviews often.

db/seeds/overviews.rb
YetAnotherGerrit: Currently this is holding me back:
https://community.zammad.org/t/reduce-file-name-length-of-vcr-cassettes-files/12336

There are workarounds, but they are time consuming to discover.

mgruner:
I will close this for now, as there seems to be no further activity / interest at present.

PR #4776, opened 2023-08-23
Fixes #4775 - healthcheck: return 504 if not healthy

### LINKED ISSUE #4775: Healthcheck: returns 200 even if unhealthy
### Used Zammad Version

6

### Environment

- Installation method: docker-compose

### Actual behaviour

Hi,
The healthcheck endpoint returns 200 even if .healthy=false.
It would help monitoring tools if the http status code would be >=400 when Zammad is unhealthy

### Expected behaviour

HTTP status code is 200 even if unhealthy

break it and observe

### Support Ticket

_No response_

no

Frankkkkk:
Hi,
By returning 504 if Zammad is not healthy, it is easy to add this url on monitoring tools that do not support parsing json (.healthy)
Cheers !

### DIFF
--- a/app/controllers/monitoring_controller.rb
+++ b/app/controllers/monitoring_controller.rb
@@ -45,7 +45,13 @@ curl http://localhost/api/v1/monitoring/health_check?token=XXX
       result[:token] = Setting.get('monitoring_token')
     end
 
-    render json: result
+    status = :ok
+    if !health_status.healthy?
+      status = :service_unavailable
+    end
+
+    render json: result, status: status
+
   end
 
 =begin


MrGeneration:
⚠️ This is a potential breaking change and may break existing variants depending on the system.
(This is just a side note from my side)

mgruner:
Hi @Frankkkkk. I can see your point, but from a REST PoV this seems to be wrong, because the HTTP status code refers to the result that was returned by the service just queried. 504 would mean that the monitoring service itself is not available / that a server side error occurred. That might cause issues with clients. Therefore I'm going to close this PR now; it will stay available as a reference for others. Thank you very much for your suggestion!

Frankkkkk:
Hi @mgruner , indeed it's _technically_ wrong, but more user-friendly IMHO. The end result of the monitoring service being available/zammad being down is the same.
Anyway, my 2¢
Cheers

PR #4850, opened 2023-10-03
added variable to deactivate IPv6

jv-barsuk:
Zammad docker-compose setup fails to start when IPv6 is not activated on the source system, because nginx and puma tries to listen on IPv6 addresses.

https://community.zammad.org/t/docker-setup-and-resources/12305

### DIFF
--- a/contrib/docker/docker-entrypoint.sh
+++ b/contrib/docker/docker-entrypoint.sh
@@ -115,12 +115,19 @@ if [ "$1" = 'zammad-init' ]; then
 elif [ "$1" = 'zammad-nginx' ]; then
   check_zammad_ready
 
+  IPV6_PATTERN="listen \[::\]:80;"
+  if [ "$ZAMMAD_IPV6" -ne 0 ]; then
+    IPV6_PATTERN="non_existent_ipv6_pattern"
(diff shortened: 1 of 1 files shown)


dominikklein:
@mgruner Maybe something you can check?

mgruner (changes requested):
Hello @jv-barsuk, thank you for your contribution. Right now I have concerns regarding this MR, mostly because it changes the behaviour. All existing systems which have ZAMMAD_IPV6 not set will switch to IPv4 only. That is not acceptable.

To change this, I would suggest that you update the PR as follows:

- use a variable `ZAMMAD_DISABLE_IPV6` instead which is not set by default
- make a separate `sed` call to remove the listen directive only if this variable is set, rather than a non-matching pattern
- change the puma invocation accordingly

With that, we'd have backwards compatible behaviour. What do you think?

mgruner:
I will close this PR for now. Please feel free to reopen / open another one with the updates. Thank you very much!

jv-barsuk:
I created a new pull request with your suggested changes, which can be found here: https://github.com/zammad/zammad/pull/4951

PR #4861, opened 2023-10-05
Add multi-arch docker builds including arm64

mgruner:
This adds support for multi-arch builds based on self hosted runners.
- We already have an experimental arm64 runner which still needs some overhaul to be production ready.
- The current approach also allows adding more architectures in future.
- Test builds available at https://hub.docker.com/repository/docker/zammad/zammad-multiarch-test/tags?page=1&ordering=last_updated


mgruner:
@monotek this is not 100% complete, but I would like to ask for your feedback. Do you agree with this approach in general? Any recommendations?

monotek:
I see guithhub artifact upload and donwload  for the maifest is used.
Is this faster as before?
If so, i'm fine with it :)

mgruner:
@monotek this data transfer is needed because the jobs to push the platform images must run first (as they are on different native platforms). Then the final jobs build the multiarch manifest lists from the generated images.

The whole process takes around 10 minutes, I believe this is pretty good. And we can add more platform jobs in future, in case that should become necessary. 

I was able to run zammad-docker-compose from the generated image on arm64 just fine. So if you agree, I would perform final clean-ups and then merge it to production. For existing docker-compose or kubernetes installations nothing should change. 🤞

monotek:
Sounds good. Lets do it :)

PR #4883, opened 2023-10-16
Remove timeout option from db connection configuration

thegcat:
The `timeout` connection option is only relevant for `sqlite3` connections, and it seems those are not supported anymore. This option can thus be removed.

### DIFF
--- a/config/database/database.yml
+++ b/config/database/database.yml
@@ -5,7 +5,6 @@ default: &default
   # For details on connection pooling, see Rails configuration guide
   # http://guides.rubyonrails.org/configuring.html#database-pooling
   pool: 50
-  timeout: 5000
   encoding: utf8
 
   ##### postgresql config #####


mgruner:
Hello @thegcat, thank you very much for your contribution. Can you briefly explain this, please? [In the docs](https://guides.rubyonrails.org/v7.0/configuring.html#database-pooling) it says:

> If you try to use more connections than are available, Active Record will block you and wait for a connection from the pool. If it cannot get a connection, a timeout error similar to that given below will be thrown.

And that does not seem to be limited to sqlite only.

thegcat:
> Hello @thegcat, thank you very much for your contribution. Can you briefly explain this, please?

I'm afraid brevity was not easy to achieve, sorry.

Please consult the [connection pool configuration options](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L95-L106) that do state specific `idle_timeout` and `checkout_timeout` options, but not `timeout` option. If you go down the rabbit hole looking for `AR::ConnectionTimeoutError`, you will [find it `raise`d in the connection pool code](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb#L128) that [ultimately uses `checkout_timeout`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L671). The [general DB config](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/database_configurations/database_config.rb) also doesn't carry a no adjective `timeout`.

The docs show the `timeout` option only on `sqlite3` connections, and that is also [the only place I was able to find it used](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L732-L741). I can't easily give a negative proof for this though, so I guess you'll have to believe me I haven't found other places that would use a `timeout` from the DB config.

Last but not least, I just realised the phrasing of my first message could be misleading. What I meant was "as Zammad doesn't support `sqlite3` the `timeout` config is superfluous".

Please also note that the [defaults for the DB config](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/database_configurations/hash_config.rb#L91-L104) already specify a 5s `checkout_timeout`, which would in spirit correspond to the 5000ms `timeout` specified in the Zammad documentation, so it would not need to be explicitly defined.

thegcat:
As a quick aside, we stumbled over this problem while using Zammad with a pgbouncer. We had some issues with Zammad keeping too many idle connections open and assumed the `timeout` option explicitly specified in the Zammad documentation would help us solve this problem, it turned out to be a red herring and to have no effect on a postgresql connection configuration whatsoever.

We hope we can avoid others also chasing this false assumption by removing the unused configuration option from the Zammad documentation.

mgruner:
@thegcat thank you so much for your detailed explanation, research and especially for sharing the improvement with the community! I will make sure this gets merged soon.

thegcat:
Thank you. I have made a note to try to improve the Rails documentation, it seems it can also be made more clear on this point.

FreeMinded:
> As a quick aside, we stumbled over this problem while using Zammad with a pgbouncer. We had some issues with Zammad keeping too many idle connections open and assumed the `timeout` option explicitly specified in the Zammad documentation would help us solve this problem, it turned out to be a red herring and to have no effect on a postgresql connection configuration whatsoever.

@thegcat we are currently hit by this exact issue. What was your solution?

MrGeneration:
https://github.com/zammad/zammad/pull/4883/files

For anything else, please use the community for technical questions (https://community.zammad.org)

PR #4900, opened 2023-10-20
WIP: Trigger to write a static Telegram message

Gogowitsch:
I couldn’t get the feature to work locally. It seems like no triggers are ever executed. Can you give me a pointer which command to execute? Maybe some queuing or cron command?

I have copied the Telegram feature from the Note feature. The idea is that a customizable string is configured in the trigger and both added as an article to the ticket as well as sent out to Telegram to the external user who created the ticket by writing a Telegram bot.

If you want to work with me in getting this feature to work and to write tests for it: let me know here. Thanks! 🙏


dominikklein:
Hi @fonata 

thanks for your effort.

Is your background job service running? Normally it should work when all needed services are running.

In general, about the feature, I think a telegram auto response could be something interesting, but this points more in the notification direction (e.g. like for SMS). But I think from a technical perspective it's very similar.

dominikklein:
@gogowitsch Feel free to bring this PR on a level we can check if it's something we see in the core. But currently it's not our focus and the PR has also not a state in which it can added to the core product.

PR #4920, opened 2023-11-08
Maintenance: Generate docker tags for Zammad major.minor and major versions

mgruner:
This change will
- add `major.minor` and `major` docker tags that can be used to get automatic updates with docker-compose and helm
- remove the dysfunctional `semver` tag configuration - does not work because we don't use semver tags with `v1.2.3`
- stop creating the duplicated `zammad-*` tags

### DIFF
--- a/.github/workflows/docker-release.yaml
+++ b/.github/workflows/docker-release.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - stable
       - develop
+      - develop-*
     tags:
       - '*'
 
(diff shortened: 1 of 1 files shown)


mgruner:
@monotek can you please have a look? With Zammad 6.2, I would like to start pushing and updating tags like `6.2` and `6` (like it is done for https://hub.docker.com/r/selenium/standalone-chrome/tags, for example).
This will allow users to receive patchlevel updates automatically from docker-compose or helm without having to update the .env all the time.

The `semver` config we had there didn't seem to work at all, and I would also like to remove the `zammad-*` tags if you agree. What do you think?

monotek (approved):
lgtm but ci fails :shrug:

mgruner:
Thanks @monotek. CI was improved in the meantime and should work now. I will then remove the debug comments and merge this.

PR #4951, opened 2023-11-28
added variable to deactivate IPv6

jv-barsuk:
Zammad docker-compose setup fails to start when IPv6 is not activated on the source system, because nginx and puma tries to listen on IPv6 addresses.

https://community.zammad.org/t/docker-setup-and-resources/12305

Original pull request: https://github.com/zammad/zammad/pull/4850

### DIFF
--- a/contrib/docker/docker-entrypoint.sh
+++ b/contrib/docker/docker-entrypoint.sh
@@ -115,20 +115,18 @@ if [ "$1" = 'zammad-init' ]; then
 elif [ "$1" = 'zammad-nginx' ]; then
   check_zammad_ready
 
-  IPV6_PATTERN="listen \[::\]:80;"
-  if [ "$ZAMMAD_IPV6" -ne 0 ]; then
-    IPV6_PATTERN="non_existent_ipv6_pattern"
-  fi
-
-
   # configure nginx
   sed -e "s#\(listen\)\(.*\)80#\1\2${NGINX_PORT}#g" \
       -e "s#proxy_set_header X-Forwarded-Proto .*;#proxy_set_header X-Forwarded-Proto ${NGINX_SERVER_SCHEME};#g" \
       -e "s#server .*:3000#server ${ZAMMAD_RAILSSERVER_HOST}:${ZAMMAD_RAILSSERVER_PORT}#g" \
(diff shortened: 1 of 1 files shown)


contrib/docker/docker-entrypoint.sh
```diff
@@ -123,6 +123,11 @@ elif [ "$1" = 'zammad-nginx' ]; then
       -e "s#server_name .*#server_name ${NGINX_SERVER_NAME};#g" \
       -e 's#/var/log/nginx/zammad.\(access\|error\).log#/dev/stdout#g' < contrib/nginx/zammad.conf > /etc/nginx/sites-enabled/default
 
+  IPV6_PATTERN="listen \[::\]:80;"
```
mgruner, line 126: We can probably move this into the `if`.

contrib/docker/docker-entrypoint.sh
```diff
@@ -134,7 +139,9 @@ elif [ "$1" = 'zammad-railsserver' ]; then
   echo "starting railsserver... with WEB_CONCURRENCY=${ZAMMAD_WEB_CONCURRENCY}"
 
   #shellcheck disable=SC2101
-  exec bundle exec puma -b tcp://[::]:"${ZAMMAD_RAILSSERVER_PORT}" -w "${ZAMMAD_WEB_CONCURRENCY}" -e "${RAILS_ENV}"
+  if [ "$ZAMMAD_DISABLE_IPV6"==1 ]; then
+    exec bundle exec puma -b tcp://0.0.0.0:"${ZAMMAD_RAILSSERVER_PORT}" -w "${ZAMMAD_WEB_CONCURRENCY}" -e "${RAILS_ENV}"
+  fi
```
mgruner, line 144: Here it seems that an `else` is missing, the default IPv6 case. To avoid redundancy, we could consider declaring a PUMA_BIND_ADDRESS with the default IPv6 value, and changing it in the `if` statement to the IPv4 value. Then only one puma call can be made afterwards.

It would also be good to add the variable to the top of the file where all variables are initialized (default to empty).

mgruner (changes requested):

dominikklein:
@mgruner Can/Should we do the last changes on our own or maybe @jv-barsuk you have some time to react?

Otherwise, I would close the PR.

Issue #5002, opened 2024-01-10
Core Workflow: Removing groups with re-adding some discards all permissions the user has


MrGeneration:
### Used Zammad Version

6.2.0.0e7f99e15.saas

### Environment

- Installation method: any
- Operating system (if you're unsure: `cat /etc/os-release` ): any
- Database + version:  any
- Elasticsearch version: any
- Browser + version: verified on current Vivaldi (Win)

### Actual behaviour

Agents cannot create or update any ticket if there's two core workflows in place:
- one that removes all available groups
- one that adds specific groups (depending on the use case)

Doing so will cause Zammad to throw

```
E, [2024-01-07T23:41:04.671145#12472-145300] ERROR -- : Invalid value '7' for field 'owner_id'! (Exceptions::ApplicationModel)
app/models/concerns/checks_core_workflow.rb:50:in `block in check_restrict_values'
app/models/concerns/checks_core_workflow.rb:45:in `each_key'
app/models/concerns/checks_core_workflow.rb:45:in `check_restrict_values'
app/models/concerns/checks_core_workflow.rb:40:in `validate_workflows'
app/controllers/tickets_controller.rb:165:in `block in create'
app/models/transaction.rb:17:in `block in execute'
app/models/transaction.rb:14:in `execute'
app/models/transaction.rb:26:in `execute'
app/controllers/tickets_controller.rb:85:in `create'
app/controllers/application_controller/has_download.rb:17:in `block (4 levels) in <module:HasDownload>'
app/controllers/application_controller/has_download.rb:16:in `block (3 levels) in <module:HasDownload>'
app/controllers/application_controller/has_download.rb:15:in `block (2 levels) in <module:HasDownload>'
app/controllers/application_controller/handles_transitions.rb:16:in `handle_transaction'
E, [2024-01-07T23:41:04.673090#12472-145300] ERROR -- : Error ID MoggpXLQ: Invalid value '7' for field 'owner_id'!
```

Even though the agent does have all the required permissions to do so.
The information regarding the users permission seems to be lost during the check.

![feature_group_removal_breaks_owner_check](https://github.com/zammad/zammad/assets/6549061/0a0b3c4d-9490-4839-bb30-0a1d4a6cf863)

### Expected behaviour

You can remove non relevant groups and add them as per requirement via Core Workflows without issues.

### Steps to reproduce the behaviour

- create a Core workflow that removes all groups
- create a Core workflow with lower priority that adds one or two specific groups
[...]

PR #5003, opened 2024-01-10
Core Workflows Fix for #5002

bahusafoo:
Created method merge_changes_into_restrict_values to resolve cases where perform_result does not have some of the submitted data while being passed to CoreWorkflow.perform().  This resolves an issue encountered when core workflows remove groups, and then add them back to the group selection object later on.

### DIFF
--- a/app/models/concerns/checks_core_workflow.rb
+++ b/app/models/concerns/checks_core_workflow.rb
@@ -1,4 +1,4 @@
-# Copyright (C) 2012-2024 Zammad Foundation, https://zammad-foundation.org/
+# Copyright (C) 2012-2023 Zammad Foundation, https://zammad-foundation.org/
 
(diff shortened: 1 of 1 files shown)


bahusafoo:
Code works nicely with any of these commits Checks seem to not like the use of "do <something> unless <something>" syntax - changed to use if not syntax instead.

rolfschmidt:
Hi @bahusafoo ,

normally the concern is only the bridge between the core workflow engine and the rails model. I think the real reason for this issue is that there was no rerun for the group after the option was re-enabled so it was still restricted and did not recover the original state. But thank you very much for your encouragement :) Hopefully it will work for you too when it gets merged to stable soon. If not feel free to create a new issue with steps for reproduction.

PR #5044, opened 2024-02-06
Fixes #5017 - No "From"-Field in Articles created on behalf of a customer with auto-generated login

### LINKED ISSUE #5017: No "From"-Field in Articles created on behalf of a customer with auto-generated login
### Used Zammad Version

6.2

### Environment

- Operating system: debian 11
- Database + version:  postgresql 13.13
- Elasticsearch version: 8.12
- Browser + version: firefox 121

### Actual behaviour

When creating tickets on behalf of customers with autogenerated "login" field (because they don't have to login), the "ticket_articles.from"-field is an empty string in the DB and doesn't get shown when clicking on the article.
Creating those tickets worked correctly in Zammad 4.1 (didn't try in Zammad 5.x).

customer without login:
without_login

customer with login (also created by the support team instead of the customer):
with_login

older ticket created with zammad 4.1:
without_login_4 1

### 

### Expected behaviour

Provide a fallback for these users by using just the Username, like in zammad 4.1, I think the commit removing this behaviour was 9f670d26aa, which commented out `transform_to_from(article)` in `Service::Ticket::Article::Create`.

create a user without email-address or password to create a auto-some-uuid-login
create a ticket for this user
click on the first article
there should be no "from"-field

### Support Ticket

_No response_

yes

fgeisser2:
If the customer for which a ticket is created has no email-address currently no creator is shown for the first article in the ticket.
This patch uses just the firstname and the lastname in the from-field of the article instead of using an empty string.
Fixes #5017 as now the name of the customer gets shown.

### DIFF
--- a/app/models/ticket/article/adds_metadata_general.rb
+++ b/app/models/ticket/article/adds_metadata_general.rb
@@ -63,7 +63,7 @@ module Ticket::Article::AddsMetadataGeneral
     type        = Ticket::Article::Type.lookup(id: type_id)
     is_customer = !author.permissions?('ticket.agent')
 
-    self.from = if %w[web phone].include?(type.name) && is_customer
+    self.from = if %w[web phone].include?(type.name) && is_customer && !author.email.empty?
                   Channel::EmailBuild.recipient_line "#{author.firstname} #{author.lastname}", author.email
                 else
                   "#{author.firstname} #{author.lastname}"


dominikklein:
@mantas what do you think?

mantas:
Looks good good to me! I'll run this in CI.

I've a feeling `!author.email.empty?` may get flagged by our Rubocop setup, but I'll take care of it :)

mantas:
Thank you @fgeisser2 ! This is now merged with https://github.com/zammad/zammad/issues/5017.

PR #5058, opened 2024-02-16
Group Permissions: Limit users to posting interally

tyler-wright:
Hello Zammad Team,

You have an amazing product here - thanks for open sourcing it. It's been a pleasure to poke around in.

### Overview of Proposed Feature

So far I've been evaluating it for a somewhat novel application, and I found a missing feature which is a blocker for me. Competing platforms like Zendesk and even open source competitors like osTicket have the ability to restrict users to only posting internally. 

I wasn't able to find this feature in Zammad - the permissions system doesn't seem to dig in a very granular fashion into the functionality that's available. However, you have such a great system for distinguishing between internal/external correspondence. It's all there - so I figured it can't be too hard to build the feature.

The main problem I'm trying to solve is to give access to Agents that shouldn't be able to accidentally send external correspondence. Instead, the Agents are able to assist more senior Agents by creating drafts and internal comments on tickets, subject to group access. The senior Agents can then review and respond using their external access permissions. 

I thought the implementation best into Group/Role access permissions, because it means that you can have some Agents with full access to some Groups, but only read or internal access on other Groups. So it adds more flexibility.

See below for a visual guide of what I've changed in this PR.

### Role Permissions

There is a new item in the Role Permissions that allows you to select 'External' for a Group.

Screenshot from 2024-02-16 17-53-06

In this example, this role can post externally for tickets in the Users group but not the Random Group, where they are limited to internal.

Below you can see a Role can be created called 'Internal Agent' where you can configure what groups this Agent has access to. This allows you to setup a role where all access will be internal to every Group, perfect for what I'm after.

Screenshot from 2024-02-16 17-53-37

Then below you can see what a Limited User's access looks like in the Ticket Zoom overview. It disables the visual buttons to toggle tickets internal/externally, and by default any type of channel will be marked as internal correspondence.

Screenshot from 2024-02-16 17-56-00

Screenshot from 2024-02-16 17-56-11

### What this PR doesn't address (yet)

1. **Tests** - I have not written any tests for this code, I could use some guidance on doing so. The testing section in your developer documentation is not terrible, but I haven't yet had time to setup the different front-end and back-end testing suites.
2. **API Permissions** - I am not sure if the external correspondence can be controlled via the API or whether this needs addressing
3. **Use of 'Full' Permissions** - I am not 100% certain whether changes in the `creates_ticket_articles.rb` will be sufficient to cover a case where a User or a Role has 'full' permissions, it might still prevent them from posting externally because `authorise!(ticket, :external?)` doesn't allow. I've run out of testing time.

### Next Steps

It would be great if this feature was implemented. If it's not on the roadmap, I'd be interested in any advice on how to create it as a package/add-on. I haven't yet explored that because I wanted to see if there was genuine interest in merging it into the product.

Thanks for your time!

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_action/internal.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_action/internal.coffee
@@ -2,6 +2,8 @@ class Internal
   @action: (actions, ticket, article, ui) ->
     return actions if !ticket.editable()
     return actions if ticket.currentView() is 'customer'
+    # Even if above satisfied, still don't display the external/internal toggle if user doesn't have external or full access to group
+    return actions if (!ticket.userGroupAccess('external') or !ticket.userGroupAccess('full'))
 
     if article.internal is true
       actions.push {
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_new.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_new.coffee
@@ -38,7 +38,8 @@ class App.TicketZoomArticleNew extends App.Controller
     @type = @defaults['type'] || 'note'
     @setPossibleArticleTypes()
 
-    if @ticket.currentView() is 'agent'
+    # If you're an agent, and you have external or full group access, you can access the internal selector
+    if @ticket.currentView() is 'agent' and (@ticket.userGroupAccess('external') or @ticket.userGroupAccess('full'))
       @internalSelector = true
 
     @textareaHeight =
@@ -424,6 +425,10 @@ class App.TicketZoomArticleNew extends App.Controller
     for articleTypeConfig in @articleTypes
       if articleTypeConfig.name is type
         config = articleTypeConfig
+
+    # Automatically select internal on the new article if the user doesn't have external or full access
+    if (!@ticket.userGroupAccess('external') and !@ticket.userGroupAccess('full'))
+      config.internal = true
 
     if config
       if config.internal
--- a/app/assets/javascripts/app/models/group.coffee
+++ b/app/assets/javascripts/app/models/group.coffee
@@ -44,6 +44,7 @@ class App.Group extends App.Model
     read: __('Read')
     create: __('Create')
     change: __('Change')
+    external: __('External') # External role allows posting articles with internal = false
     overview: __('Overview')
     full: __('Full')
 
--- a/app/controllers/concerns/creates_ticket_articles.rb
+++ b/app/controllers/concerns/creates_ticket_articles.rb
@@ -51,6 +51,11 @@ module CreatesTicketArticles # rubocop:disable Metrics/ModuleLength
       clean_params[:internal] = false
     end
 
+    # Do not allow creation if you don't have external or full group access
+    if clean_params[:internal] == false
+      authorize!(ticket, :external?) # TODO: This might stop those with 'full' access from posting, check it doesn't
+    end
+
     article = Ticket::Article.new(clean_params)
     article.ticket_id = ticket.id
     article.check_mentions_raises_error = true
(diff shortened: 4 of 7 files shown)


dominikklein:
@tyler-wright Thank you very much for the idea, I think we would first need to talk internally if this is a feature that we see in the product, and when we think it's interesting we need to check how this feature could be implemented. Then it would be possible to check together, what needs to be aligned in this PR.
Currently, it's difficult to say how long this will take.

tyler-wright:
@dominikklein No worries, let me know if I can assist in any way. I may run it in my production instance in the meantime and see how it works out. Currently Ive noticed that even if you set an email to be internal, it still ends to external recipients - so that's a bit counter-intuitive.

gb-123-git:
+1 For this Feature.

dominikklein:
@YetAnotherGerrit Maybe we should think if such an problem is something we need to solve inside the product (I'm not saying that the current solution in the PR is the way to go). But first we would need a product decision about such an feature.

YetAnotherGerrit:
Thank you for the contribution!

Usually the group structure should represent the team structure. So if a ticket is in the responsibility of a team that should not communicate directly to the customer, they have their own group. This might be because its a team of trainees, that should check with their seniors before answering, or it is a department that should focus on other stuff and a "customer service" team takes care of the external communication.

In my described approach it is easy to prevent the team from sending e-mails, by just disabling the outgoing e-mail-address on their group. But it currently does not prevent that team to write public notes, or SMS, or WhatsApp, and so on. It also disables triggers that need to send emails.

I see the need for a revamp regarding "internal" and "external" communication. But to bring that into live, we need to consider more than the UI. As you already recognized, API and E-Mail are things we need to think about. And other communication channels as well. And shared drafts. There are also use cases where Zammad is used completely without email addresses.

We will do some refinement about the consequences of that change and circle back to you. As we have an upcoming event that will block some time, it will not be within days, sorry!

meus-rls:
i like this feature a lot. thanks for your effort to describe it in detail and to commit it.
just as a comment: in our case what we also need is the _inverse_ feature of this one: an agent group that is not allowed to read and write internal comments.
background: we have a first and second level support. both are to be able to comment publicly. but only second level is supposed to discuss technical questions internally. first level support should not see the internal comments of second level support. 
as discussed in this commit: this is not configurable by existing means.

YetAnotherGerrit:
The `change` permission currently grants you everything you need to make adjustments on tickets (without the option to be an owner). Which is something I don't want to see changed. By introducing the `external` permission we force every Zammad admin to change the way, they configure their system. Every Zammad admin need to re-learn the permission system to achieve the same behaviour they are used to configure.

But it was always odd, that the `read` permission allows you to see internal communication without the possibility to participate in it.

So, my counter suggestion would be to change the behaviour of the `read` permission. Not for existing systems, that would be a breaking change, but for fresh instances. And to add a global option to activate that behaviour for existing systems as well.

This would allow users with the `read` permission to write internal notes without having `change` as a permission. From my perspective, this would solve your use case (and that of plenty of other Zammad users). Correct?

@meus-rls your use case might be similar but technically it is completely different. I suggest creating a feature request in the Community for it.

jsuchal:
I've already created this here https://community.zammad.org/t/agent-permission-only-to-comment-internally-on-ticket/16920

Another possibility (or probably a more backwards compatible way) to do this would be to expand core workflow with an action/option to be able to set article response types/visibility to "fixed" as you currently can with other fields. This will also be much more powerful in terms of conditions that could be used for such visibility logic.

dominikklein:
@YetAnotherGerrit Sounds to me something we need to tackle on our own at some point, with all consequences and so on. Will maybe not that easy to finalize this inside a PR.

@tyler-wright  For now I would close this PR, but thank your for the feature idea and trying to bring this forward.
But for sure you can also try to bring it forward on your own with having the direction of the described idea from @YetAnotherGerrit .

PR #5065, opened 2024-02-23
Maintenance: Drop nulldb adapter.

mgruner:
Possible solution for https://github.com/zammad/zammad-helm/issues/251. Opinions?


monotek:
Are we sure we don't need it anymore?
I try to remember for what exact reason i've added the nulldb adapter.
Seems to be some workaround to fix an issue with the docker builds: https://github.com/zammad/zammad/commit/04e2d8ccc7947620f581409d536835ae7a7dd5cf

At least i can't find a "rake assets:precompile without db" in the code with a quick search :shrug:

mgruner:
@monotek yes, I believe that was the only reason. And as you can see by the changes in the build script, and the successful `docker-ci` job, building works without nulldb now. Ok?

monotek (approved):

mgruner:
Merged in ed3a9d1, thanks.

PR #5087, opened 2024-03-14
Maintenance: Drop zammad.ready file from docker entrypoint.

mgruner:
Zammad 6.3 will drop the `var/` folder, so that no share is needed any more for it. That's a good chance to improve the container waiting in the docker compose context as well.

Rather than placing a file in the `var` share to signal readiness, we can simply call the rails stack and ask it to fail if migrations are pending. Other parts in the `zammad-init` container such as `Translation.sync` can happen later and are not a requirement for services to start. This will also reduce start time / upgrade downtime significantly.

zammad-helm will soon handle it in a similar way (see https://github.com/zammad/zammad-helm/pull/243).

### DIFF
--- a/contrib/docker/docker-entrypoint.sh
+++ b/contrib/docker/docker-entrypoint.sh
@@ -35,7 +35,7 @@ ESCAPED_POSTGRESQL_PASS=$(echo "$POSTGRESQL_PASS" | sed -e 's/[\/&]/\\&/g')
 export DATABASE_URL="postgres://${POSTGRESQL_USER}:${ESCAPED_POSTGRESQL_PASS}@${POSTGRESQL_HOST}:${POSTGRESQL_PORT}/${POSTGRESQL_DB}${POSTGRESQL_OPTIONS}"
 
 function check_zammad_ready {
-  until RAILS_CHECK_PENDING_MIGRATIONS=1 bundle exec rails r true &> /dev/null; do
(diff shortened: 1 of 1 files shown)


monotek (commented):
The initial idea behind this whole thing was to reduce logs with error messages during startup in docker-compose.

The helm chart should not use such file at all an do the right thing out of the box (restart the container if it can't reach one of it´s dependencies like postgres).

Maybe we can move to this pattern in docker-compose too?

mgruner:
> The initial idea behind this whole thing was to reduce logs with error messages during startup in docker-compose.

The PR provides such behaviours. The service containers will print a few lines like 

```
waiting for init container to finish install or update...
waiting for init container to finish install or update...
waiting for init container to finish install or update...
```

Until everything is ready. And then they start up. No error messages! IMHO that is very user friendly behaviour, isn't it?

> 
> The helm chart should not use such file at all an do the right thing out of the box (restart the container if it can't reach one of it´s dependencies like postgres).
> 
> Maybe we can move to this pattern in docker-compose too?

It might be possible to control the restart policy of the containers somehow, but the current proposal is better IMHO: it uses less resources, and people are perhaps not used to restarting containers in such context than in k8s. It might also conflict with customizations that users made.

So I would suggest to stick with this proposal.

monotek (approved):

mgruner:
https://github.com/zammad/zammad/commit/c4d0f943ca343ec6f6771bb953fcc372af9b3d2a

Issue #5101, opened 2024-03-27
Fetching LdapSource for unknown users fails


infeix:
### Used Zammad Version

6.2.0-1711349647.bc0c70d0.focal

### Environment

- Installation method: [e.g. source, package]
package

- Operating system (if you're unsure: `cat /etc/os-release` ): [e.g. debian 10.4, ubuntu 20.04]
NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.6 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

- Database + version:  [e.g. postgresql 9.3, mysql 5.7, mariadb 10.3]
postgresql 9.4

- Elasticsearch version: [e.g. 7.17]
elasticsearch/stable,now 7.17.18 amd64

### Actual behaviour

ERROR in production.log:

I, [2024-03-27T11:23:54.692386#997-142260]  INFO -- : Skipping. Found an existing user for login 'woholi' synced by LDAP.
E, [2024-03-27T11:23:54.922850#997-142260] ERROR -- : ImportJob 'Import::Exchange' failed: undefined method `source' for nil:NilClass

    return if !%r{^Ldap::(\d+)$}.match?(user.source)
                                            ^^^^^^^
E, [2024-03-27T11:23:54.922940#997-142260] ERROR -- : undefined method `source' for nil:NilClass

    return if !%r{^Ldap::(\d+)$}.match?(user.source)
                                            ^^^^^^^ (NoMethodError)
app/models/ldap_source.rb:13:in `by_user'
lib/sequencer/unit/import/common/user/skip/is_ldap_source.rb:14:in `process'
lib/sequencer/unit/import/common/model/mixin/skip/action.rb:42:in `process'
lib/sequencer/unit/base.rb:247:in `process'
lib/sequencer.rb:76:in `block (4 levels) in process'
lib/mixin/start_finish_logger.rb:9:in `log_start_finish'
lib/sequencer.rb:75:in `block (3 levels) in process'
[...]

PR #5102, opened 2024-03-27
Fix bug #5101

infeix:
This small change did the job for me.
But I don't have a unit test.


fliebe92:
Hello @infeix. We're working on the aforementioned issue, and we'll implement a slightly different approach, even with a test case. Therefore, we're going to close this pull request. However, thanks a lot for your report and this pull request and all the effort you put into it. ❤️

PR #5143, opened 2024-04-23
fixes #5142| fix db check in preinstall.sh

### LINKED ISSUE #5142: contrib/packager.io/preinstall.sh does not check the db adapter setting and fails if both psql and mysql are installed
### Used Zammad Version

6.3

### Environment

- Installation method: deb package
- Operating system (if you're unsure: `cat /etc/os-release` ): Debian 12 Bookworm
- Database + version: MariaDB 10.11
- Elasticsearch version: 7.10
- Browser + version: n/a

### Actual behaviour

When running Zammad with Mysql and also having `psql` installed on the same system the `preinstall.sh` script fails as it tries to connect to a non existing postgresql database

### Expected behaviour

the script should also check the adapter setting in the `database.yml` and use the correct utility to check the availability of the database

- Install Zammad with MySQL/MariaDB
- Install psql-client on the same Server
- try to upgrade through the debian packages

### Support Ticket

_No response_

yes

jr-timme:
Make sure the correct database is used depending on the adapter setting

Fixes: https://github.com/zammad/zammad/issues/5142


mgruner:
Excellent fix, thank you so much for your contribution @jr-timme!

PR #5210, opened 2024-06-11
Consistent capitalization for WhatsApp

maxxer:
Found during translation


mgruner:
Thank you very much @maxxer

maxxer:
Thank you for taking care of this, I don't (yet) have set up Zammad dev env!

PR #5255, opened 2024-07-09
Fixed 209 in the feature requests in the forum

MauritzFunke:
Maybe some help needed with tests. I already added a feature request in the community forum but its still pending. It has the id 209

### DIFF
--- a/app/assets/javascripts/app/models/ticket_article.coffee
+++ b/app/assets/javascripts/app/models/ticket_article.coffee
@@ -13,9 +13,9 @@ class App.TicketArticle extends App.Model
       { name: 'sender_id',      display: __('Sender'),      tag: 'select',   multiple: false, null: false, relation: 'TicketArticleSender', default: '' },
       { name: 'internal',       display: __('Visibility'),  tag: 'radio',  default: false,  null: true, options: { true: 'internal', false: 'public' } },
       { name: 'created_by_id',  display: __('Created by'),  relation: 'User', readonly: 1 },
-      { name: 'created_at',     display: __('Created'),     tag: 'datetime', readonly: 1, searchable: false },
-      { name: 'updated_by_id',  display: __('Updated by'),  relation: 'User', readonly: 1, searchable: false },
-      { name: 'updated_at',     display: __('Updated'),     tag: 'datetime', readonly: 1, searchable: false },
+      { name: 'created_at',     display: __('Created'),     tag: 'datetime', readonly: 1, searchable: true },
+      { name: 'updated_by_id',  display: __('Updated by'),  relation: 'User', readonly: 1, searchable: true },
+      { name: 'updated_at',     display: __('Updated'),     tag: 'datetime', readonly: 1, searchable: true },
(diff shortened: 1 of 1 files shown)


MauritzFunke:
Here is the feature request https://community.zammad.org/t/add-created-at-filter-to-articles/14892/1

dominikklein:
@rolfschmidt Maybe some thought if this is a good idea?

dominikklein:
@martini What do you think about such a change?

mantas:
I've a feeling such change would be confusing and have unexpected results. This may open a pandora box of querying among ticket articles in a single ticket. Or we could hide this behind a flag for people who know what they're doing.

I'm not sure about the use case mentioned in the community. IIRC Overviews show all timeunits in a ticket during it's lifetime. Not timeunits on a given day. While such overview would fetch tickets touched in a given day. It'd show time units logged in other days too.

Listing time units in a specific day is definitely an important metric though.

dominikklein:
When we don't see any use case we should close this idea or? :-) @YetAnotherGerrit Don't know if this is something you heard already before about.

MrGeneration:
The use case is to see accounted time per day (at least that's what I understand is the ultimate goal), however, searching articles based on their created_at does not accomplish this. Accounted times are not part of articles and thus I'd say this PR does not "close" the FR mentioned above.

YetAnotherGerrit:
Accounted time should be part of articles. But that is a much more complex task to achieve in the future. I don't understand the value of this MR. Does this add 3 new attributes to the search index?

dominikklein:
@MauritzFunke Maybe some additional input from your side, otherwise we would close the PR for now.

MauritzFunke:
Sorry for the late reply, the goal would be to get metrics about the time used on tickets per user and other metrics to somewhat monitor productivity.

MrGeneration:
But these attributes are available in the elasticsearch index and thus can be used with e.g. Grafana...?
Our default example dashboards use these actually....

https://docs.zammad.org/en/latest/appendix/reporting-tools-thirdparty.html

MauritzFunke:
Last time I looked I either overlooked the reporting tools section or it wasn't there

The metric I'm trying to get is time accounted per agent per month. Previously I had the issue with tickets with overlapping months:

27.04.2025: Customer creates ticket
28.04.2025: Agent does step 1-2 needed to solve the ticket but needs input from customer (2 hours)
02.05.2025: Customer answers
02.05.2025: Agent does final steps to solve ticket (1 hour)

Searching for the ticket shows the time units in April but the last article is in May and should be accounted for in May.

If I look correctly in the [documentation](https://docs.zammad.org/en/latest/install/elasticsearch/indexed-attributes.html#ticket) there is no search index for time units in articles.


Edit:
If I see correctly, articles also include the information. I'll try to get the information I need and reply if there are some unexpected issues

MrGeneration:
image

ticket articles have nothing to do with time tracking functionality at all. At least not attribute wise.

dominikklein:
Thanks for the contribution idea.

I think currently there is no plan to merge this and seems to be it's solvable with an external reporting tool like Grafana.

PR #5288, opened 2024-08-05
Show reply all button when cc

### LINKED ISSUE #4325: When replying to tickets with external addresses in CC, the "Reply to all" option is hidden
Used Zammad version: 5.2.3
Installation method (source, package, ..): any
Operating system: any
Database + version: any
Elasticsearch version: any
Browser + version: Microsoft Edge, Version 105.0.1343.33 (Official Build) (64-Bit)
Ticket: #10117381, #10121985, #10163461, #10171863, #10210383

### Expected behavior:

* an agent creates a ticket by sending an email (with external addresses in cc, at least one) to the helpdesk. The "Reply to all" option remains visible and can be used.

### Actual behavior:

* An agent creates a ticket by sending an email to the helpdesk with an email address in the cc field. The "Reply to all" option remains visible and can only be used if the cc field contains internal (other agents) email addresses.

### Steps to reproduce the behaviour:

* Agent XY sends an email from his address (with external addresses in the cc field, at least one) to the helpdesk (example@zammad.com) and creates a ticket with it. The ticket created has only four options, and the "Reply All" option is missing.

nicktencate:
This PR fixes #4325 however as I was looking at this code I became unsure of the desired functionality.

I was also unable to find the original PR that added this functionality in the first place.

I'm not sure if the debate should be here or in the issue, so first let me offer my apologies for this triple failure on my part :)

TLDR: When do we show the 'Reply-all' button in an email ticket? The current code counts how many non-agents (foreignRecipients) are in the to/from/cc fields combined. If this total is more than 1, show the 'Reply-all' button.

I assume this was an attempt to spare various agents a lot of email - but the current code does not actually remove any email addresses or spare them from getting an email; except if they are CC'd in a ticket, then they are un-CC'd since there is no 'Reply-all' button.

Personally - if the email address was included in the to/from/cc headers, I would like that person to get an email. I think this is the expected behavior from most of our agents. In this case, most of this code can be removed and a simple counter over the to/from/cc with > 2 will suffice.

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/article_action/email_reply.coffee
@@ -15,6 +15,8 @@ class EmailReply extends App.Controller
 
       # check if reply all needs to be shown
       recipients = []
+      recipientsCC = []
+
       if article.sender.name is 'Customer'
         if article.from
           localRecipients = emailAddresses.parseAddressList(article.from)
@@ -28,6 +30,7 @@ class EmailReply extends App.Controller
         localRecipients = emailAddresses.parseAddressList(article.cc)
         if localRecipients
           recipients = recipients.concat localRecipients
+          recipientsCC = recipients.concat localRecipients
 
       # remove system addresses
       localAddresses = App.EmailAddress.all()
@@ -48,7 +51,7 @@ class EmailReply extends App.Controller
               foreignRecipients.push recipient
 
       # check if reply all is needed
-      if foreignRecipients.length > 1
+      if foreignRecipients.length > 1 || recipientsCC.length > 0
         actions.push {
           name: __('reply all')
           type: 'emailReplyAll'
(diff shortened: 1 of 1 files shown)


dominikklein:
@rolfschmidt @mantas some direction which could be valid or completely wrong?

nicktencate:
Valid. def.

mantas:
Hmm. But if CC address happens to be so-called local address, then it will show `Reply All`. But on clicking such address would be filtered out.

If I'm following correctly, it'd make sense to turn `Reply All` button only if there's a foreign address in CC?

dominikklein:
> If I'm following correctly, it'd make sense to turn Reply All button only if there's a foreign address in CC?

Yes, sounds for me without a deeper look like the correct direction.

dominikklein:
@mantas When I understand you currently the current PR is not doing it in this way?

mantas:
> @mantas When I understand you currently the current PR is not doing it in this way?

Looking at my old comment, I reading it this way.

dominikklein:
@nicktencate Thanks for contribution, for now I will close this PR. But when you have any plans to improve this PR with the given input from @mantas, we can re-open it.

PR #5291, opened 2024-08-06
Fix deletion of mails successfully filtered by postmaster

### LINKED ISSUE #5049: Unprocessable mails that also match a `PostmasterFilter` with the action `{"x-zammad-ignore"=>{"value"=>true}` don't get deleted upon reprocessing
### Used Zammad Version

6.2

### Environment

- Installation method: any
- Operating system: any
- Database + version:  any
- Elasticsearch version:any
- Browser + version: any

### Actual behaviour

When (for any given reason) an email ends up in the location for unprocessable mails (currently `var/spool/unprocessable_mail`) and it also matches a `PostmasterFilter` with the action `{"x-zammad-ignore"=>{"value"=>true}`, it will be ignored (as expected) upon reprocessing. However, the `.eml` file will not be deleted from the directory.

This leads to the mail being displayed under `Monitoring` until an admin goes into the file-system and manually deletes it. 

### Expected behaviour

Unless the reprocessing of an unprocessable mail throws an error, its `.eml` file is deleted from the directory.

Have a `PostmasterFilter`:
```
 match: {"from"=>{"operator"=>"matches regex", "value"=>"foo@example.com"}},
 perform: {"x-zammad-ignore"=>{"value"=>true}},
```
Have an `.eml` file:

```eml
From: <foo@example.com>
To: <bar@example.com>
Subject: Hello, world!

Hello, world!
```
Place the `.eml` file into `var/spool/unprocessable_mail` and run:

`rails r 'Channel::EmailParser.process_unprocessable_mails'`

### Support Ticket

Ticket#10150795

yes

binsky08:
Fixes #5049

### DIFF
--- a/app/models/channel/email_parser.rb
+++ b/app/models/channel/email_parser.rb
@@ -175,7 +175,9 @@ returns
     # check ignore header
     if mail[:'x-zammad-ignore'] == 'true' || mail[:'x-zammad-ignore'] == true
       Rails.logger.info "ignored email with msgid '#{mail[:message_id]}' from '#{mail[:from]}' because of x-zammad-ignore header"
-      return
+
+      # technical correctly processed
+      return [true]
     end
 
     ticket       = nil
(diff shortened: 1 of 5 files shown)


binsky08:
fixed with e38fe78d287a7288e744d7dc084e7dd6d2842ec2

PR #5294, opened 2024-08-08
Migrate issue links to global usable ids to prevent broken links after moving repos

### LINKED ISSUE #5015: Gitlab/Github UI loses links when they are renamed
### Used Zammad Version

6.2. develop

### Environment

- Installation method: any
- Operating system: any
- Database + version:  any
- Elasticsearch version: any
- Browser + version: MS Edge v.120.0.2210.144 (64-bit)

### Actual behaviour

Renaming GitHub issues (changing name or path/repo) added to a Zammad article will cause Zammad to update the link displayed in the right sidebar the first time the page is reloaded afterward. Renaming a GitLab issue that has been added to a Zammad ticket will result in the issue being removed completely, although the issue previously connected is still displayed in the icon. 

### Expected behaviour

Renaming GitHub issues (changing name or path/repo) added to a Zammad article will cause Zammad to update the link displayed in the right sidebar the first time the page is reloaded afterward. Renaming a GitLab issue that has been added to a Zammad ticket should also be updated, as it is with GitHub issues. 

1) enable GitLab Zammad integration
2) create and add an GitLab issue to one Zammad ticket
3) change in GitLab under: Settings --> General --> Advanced --> Change path (changing the URL)--> click on "Save path" and go back to Zammad. If you only once changed the path, the GitLab integration looks like this:

Screenshot 2024-01-22 134431

### Support Ticket

Ticket#10149033

yes

binsky08:
Fixes #5015 

frontend tests needs to be validated, was not able to execute them on my dev instance

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_git_issue.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_git_issue.coffee
@@ -4,8 +4,10 @@ class App.SidebarGitIssue extends App.Controller
 
   constructor: ->
     super
+    @issueLinks     = []
     @issueGids      = []
     @issueData      = []
+    @error          = ''
     @providerIdentifier = @provider.toLowerCase()
 
   sidebarItem: =>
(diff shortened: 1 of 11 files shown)


binsky08:
too heavy migration. simpler alternative approach in progress

PR #5295, opened 2024-08-09
Allow pointer events on input:readonly fields to make it copy-/selectable

### LINKED ISSUE #5172: Read only Ticket attribute can not be copied properly
### Used Zammad Version

63

### Environment

- Installation method: any
- Operating system (if you're unsure: `cat /etc/os-release` ): any
- Database + version:  any
- Elasticsearch version: any
- Browser + version: any

### Actual behaviour

I created a custom Ticket attribute and can not mark it and copy it with ctr+c.
I have to mark the name of the ticket attribute and the ticket attribute beneath it to copy it. 

2024-05-15 09_21_01-Zammad Helpdesk - #64097 - testing organisation 2 – Opera

### Expected behaviour

I would expect to be able to copy the custom ticket attribute by marking the content of it, like when it is writable.

2024-05-15 09_23_59-Zammad Helpdesk - #64097 - testing organisation 2 – Opera

- Create a custom ticket attribute 
2024-05-15 09_26_32-Zammad Helpdesk - Subscription – Opera

- Make it visible via core workflow, and set it to read only for one role
- Try to copy the content of the ticket attribute with the read only permissions

### Support Ticket

Ticket#10154990

no

binsky08:
Fixes #5172

The commit introduces an additional css rule to modify the cursor and pointer event of the readonly input elements directly. Since changing it in line [2101](https://github.com/binsky08/zammad/blob/6e71bbfc68d4d26f7aa6bf0f89f640ca5f778d7b/app/assets/stylesheets/zammad.scss#L2101) could introduce unexpected side effects in other places, this change should only improve the ux without damaging anything else.

Change preview:

[issue-5172-readonly-input.webm](https://github.com/user-attachments/assets/ee467f8b-1464-4b25-b009-c5cfc3fedb57)

### DIFF
--- a/app/assets/stylesheets/zammad.scss
+++ b/app/assets/stylesheets/zammad.scss
@@ -2100,6 +2100,11 @@ fieldset > .form-group {
   &.is-readonly .controls:not(.ignore-readonly) {
     pointer-events: none;
     cursor: not-allowed !important;
+
+    input:read-only.form-control {
+      pointer-events: auto;
+      cursor: not-allowed;
+    }
   }
 
   &.form-group--nested {


dominikklein:
Fixed with: https://github.com/zammad/zammad/commit/b361246c9fb4bc811ed16237d9ad13638ba555b9

Issue #5225, opened 2024-06-16
SAML SSL Verification is broken


strifel:
### Used Zammad Version

6.3

### Environment

- Installation method: docker
- Operating system (if you're unsure: `cat /etc/os-release` ): debian 12

### Actual behaviour

The SAML TLS verification seems to check things on save wrongly.

1. It does check the IDP SSO URL. That is only ever accessed by the client. This should therefore not be checked by the server.
2. It does say to check the SAML IDP-Certificate. This is wrong. The certificate does not seem to be used for the check.
3. It [does check](https://github.com/zammad/zammad/blob/develop/lib/user_agent.rb#L517-L521) redirect URIs ([that are valid](https://developer.mozilla.org/en-US/docs/web/http/headers/location)) not correctly as it does not accept relative redirect URIs.

### Expected behaviour

The SAML check passes

### Steps to reproduce the behaviour

For:

1. Enter an IDP SSO Url: Break it (for example by using Authentik that does relative redirect URIs)
2. See 1, Read error message
3. Use for example Authentik that does relative redirects.

### I'm sure this is a bug and no feature request or a general question.

yes

tschaefer:
The check verifies the TLS connection/certificate to the SAML server. The idea of the check is to support the admin during the Zammad SAML configuration.

In practice, it is not needed for the Zammad server to reach the SAML server, as the SAML communication is handled between the user browser - any modern browser checks the TLS connection/certificate - and the SAML server. In this case, the check is "useless" and will fail with a connection error.

For the initial purpose of the check the current state is good enough.

strifel:
Currently this button pretends to be important and should definitely checked ("Die Deaktivierung der SSL-Überprüfung stellt ein Sicherheitsrisiko dar und sollte nur vorübergehend verwendet werden. Verwenden Sie diese Option auf eigene Gefahr!"). But as it stands it randomly breaks with some IDPs (see point 3) and does not have any security effects (as you said).
When it then breaks it even gives wrong error messages (see point 2).

I do not see where this helps the admin, it only confuses them and costs them a lot of time (as it did for us).

The button text should definitely reflect that it has no security effects.

PR #5403, opened 2024-11-10
Fix misleading language used for SAML TLS verification

strifel:
As brought up in #5225 the language used for the SAML TLS verification is quite misleading.
This PR changes the language to clearly state that the SSL verification setting does have no security impact.
It also changes the language of the error message to clearly state that the HTTPs connection to the IDP SSO target URL is the one that breaks.

### DIFF
--- a/app/models/setting/validation/saml/tls.rb
+++ b/app/models/setting/validation/saml/tls.rb
@@ -34,9 +34,9 @@ class Setting::Validation::Saml::TLS < Setting::Validation::Base
     Rails.logger.error("SAML: TLS verification failed for '#{url}': #{resp.error}")
 
     if resp.error.starts_with?('#<OpenSSL::SSL::SSLError')
-      __('The verification of the TLS connection failed. Please check the SAML IDP certificate.')
+      __('The verification of the TLS connection to the IDP SSO target URL failed. Please check the SAML IDP HTTPs certificate.')
     else
-      __('The verification of the TLS connection is not possible. Please check the SAML IDP connection.')
+      __('The verification of the TLS connection to the IDP SSO target URL is not possible. Please check the connection.')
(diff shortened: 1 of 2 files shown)


dominikklein:
@fliebe92 Maybe you could check if this text changes make sense?

YetAnotherGerrit:
@strifel can you please defend `Deactivating it has no security impact.`? From my perspective it increases security if you validate the host's certificate.

strifel:
@YetAnotherGerrit 

> @strifel can you please defend `Deactivating it has no security impact.`? From my perspective it increases security if you validate the host's certificate.

As stated by @tschaefer in #5225 

> In practice, it is not needed for the Zammad server to reach the SAML server, as the SAML communication is handled between the user browser - any modern browser checks the TLS connection/certificate - and the SAML server. In this case, the check is "useless" and will fail with a connection error.

and 

> The check verifies the TLS connection/certificate to the SAML server. The idea of the check is to support the admin during the Zammad SAML configuration.


I do understand this statements and the code to only check the endpoint of the SAML server once, when configuring as it does never need reach this url again.
This means that there is no verification of this certificate during authentication (and can therefore change from the first check).
This is not a security issue as the certificate (as stated by @tschaefer) is checked by the browser.
The integrity of the information provided by the browser is checked by verifying it against the "IDP-certificate" which is as far as I can tell not controlled by this setting.

strifel:
@dominikklein Is there anything left blocking this from getting mergef?

fliebe92:
Hi @strifel and thanks for this pull request.

We discussed it again internally, and we'll merge it soon to `develop`. Due to our internal workflow, I have to close this pull request, but no worries, the commit author is still you.

Thanks again!

PR #5449, opened 2024-12-19
Telegram add +user info

Chinoman10:
Sorry for no tests and if the commit messages don't really follow your guidelines, but I wanted to develop something quickly and I'm not super experienced with Rust myself...

This PR is to solve this Feature Request:
https://community.zammad.org/t/add-more-user-information-from-telegram/16390


dominikklein:
@Chinoman10 thanks for your contribution idea. I need to say that the current changes do not match the standard feature implementation. For example, we can not relay on "custom" attributes that are not inside the core.
In general, the needed information are also already present, maybe only not inside the UI. What we could discuss if the added information in `note` could be improved for the situation that new username exists.

But when you currently need the Telegram-ID of the User, you can do the following already: 

`User.find(?).authorizations`

Chinoman10:
I understand your point; and I figured as much, which is why I do a check to see if the property exists first, and only then populate it. But I understand this isn't a common practice.

Perhaps we can just keep the note improvement then? The way I made it, the user ID is always present in the note, regardless if a username exists or not.

dominikklein:
Maybe you can explain again a little bit more what is the use case behind? Because you can definitiv find this information of the ID. E.g. the id will also be the login of this user, when no telegram username exists.

dominikklein:
For now, we will close the merge request, because when needed, you can find the needed information already. Feel free to sent some additional use cases where the information in the note field is needed and we can check it again.

Chinoman10:
> Maybe you can explain again a little bit more what is the use case behind? Because you can definitiv find this information of the ID. E.g. the id will also be the login of this user, when no telegram username exists.

@dominikklein I only saw your comment now, sorry for that.
The use case is simple:
- When a ticket is opened by the Telegram integration, the user's username (when exists) is populated as a note: `Telegram @username`;
- Adding it to the Note field isn't ideal, but at least it's there;
- The problem is that when a username isn't available, we don't even have the TG ID populated anywhere.

Ideally we'd like to have the username+ID when available, or just the ID is no username exists/is available.

Right now the support agents have to ask the person's ID/username which feels retarded to the users, since they're chatting with someone already (a support bot).

Let me know if you need further clarifications; I can also ask the client and/or the support team leader to come and speak their opinions (_I'm only in charge of infra and have support experience, but I'm a busy founder rn, so I'm not "in the trenches" on a daily basis as they are_).

MrGeneration:
What exactly is the use case to "know the customers Telegram username" ?
The whole sense of a helpdesk system like Zammad (and any other!) is that you have a centralized communication platform.

This sounds like someone's going to send them messages next to Zammad... and if that happens: Why would you?

Chinoman10:
> What exactly is the use case to "know the customers Telegram username" ? The whole sense of a helpdesk system like Zammad (and any other!) is that you have a centralized communication platform.
> 
> This sounds like someone's going to send them messages next to Zammad... and if that happens: Why would you?

No, not at all.
The use-case is simple: the customers who open tickets are paying (or free) video-game players who play on Telegram (directly).
The support staff needs to know their Telegram information in order to run queries to find relevant game information (how much they've spent to assess their 'VIP tier', if they are missing in-game rewards as claimed by them, how much in-game currency they have, and so on).

Noone's sending DMs to the users parallel to Zammad, as you have mentioned, that's stupid and it breaches the currently implemented support staff guidelines.

dominikklein:
@Chinoman10 Thanks for the information. For me, one missing point was that it's related to the agent's "daily" business. First, I was thinking you meant in edge cases from the "admin" perspective, which often has access to this additional information in the background. 

Personally I think as a fallback, we can add the Telegeram-ID, when no username is present, then the agent can see it. Let me reopen the PR and we can handle it in the next weeks.

Chinoman10:
> @Chinoman10 Thanks for the information. For me, one missing point was that it's related to the agent's "daily" business. First, I was thinking you meant in edge cases from the "admin" perspective, which often has access to this additional information in the background.
> 
> Personally I think as a fallback, we can add the Telegeram-ID, when no username is present, then the agent can see it. Let me reopen the PR and we can handle it in the next weeks.

Only displaying the ID when the username isn't present isn't great, since users can change their username over time, as opposed to IDs which are permanent; for this reason, app/bot developers always use the ID as a primary key in DB tables (_so if the only information they have is the username, they have to do an additional step to get the ID from the username_).
I really liked the idea of what I tried doing in this PR: populate a field with a fixed name if it exists, ignore otherwise; and always add the ID on the notes (_keep existing behaviour, simply add to it_).

Thank you for looking into this.

PR #5506, opened 2025-03-05
Fallback to /proc/meminfo when lshw is unavailable

stonerl:
In Docker containers, `lshw` is often missing, causing Zammad’s system report to return `null` for total memory.

As a fallback, use `/proc/meminfo`. Note that values may slightly differ due to reserved memory.


dominikklein:
@mgruner Can you check this one?

mgruner (approved):

stonerl:
@mgruner I addressed the linter errors.

mgruner:
Merged with c895d6f7c17182c6aa058911fa5d8d94acfd7f21, thank you very much @stonerl

PR #5521, opened 2025-03-14
Dynamically set real IP from Docker gateway

stonerl:
Retrieve the Docker network gateway in docker-entrypoint.sh and inject it into the Nginx configuration to ensure correct client IP handling.

Without this, all client sessions would incorrectly show Docker's gateway IP instead of the actual client IP, when using a second proxy in front of the Nginx container.

### DIFF
--- a/contrib/docker/docker-entrypoint.sh
+++ b/contrib/docker/docker-entrypoint.sh
@@ -25,6 +25,7 @@ set -e
 : "${RAILS_LOG_TO_STDOUT:=true}"
 : "${RAILS_TRUSTED_PROXIES:=127.0.0.1,::1}"
 : "${ZAMMAD_DIR:=/opt/zammad}"
+: "${ZAMMAD_EXT_PROXY:=false}"
 : "${ZAMMAD_RAILSSERVER_HOST:=zammad-railsserver}"
 : "${ZAMMAD_RAILSSERVER_PORT:=3000}"
 : "${ZAMMAD_WEBSOCKET_HOST:=zammad-websocket}"
@@ -122,13 +123,30 @@ elif [ "$1" = 'zammad-nginx' ]; then
   # There is a workaround needed to support DNS resolution of upstream container names with short TTL:
   #   we set the proxy pass directly with a variable including the URL (!), rather than just referring to the
   #   upstream {} definition. For details, see https://tenzer.dk/nginx-with-dynamic-upstreams/.
-  sed -e "s#\(listen\)\(.*\)80#\1\2${NGINX_PORT}#g" \
-      -e "s#proxy_set_header X-Forwarded-Proto .*;#proxy_set_header X-Forwarded-Proto ${NGINX_SERVER_SCHEME};#g" \
-      -e "s#proxy_pass http://zammad-railsserver;#set \$zammad_railsserver_url http://${ZAMMAD_RAILSSERVER_HOST}:${ZAMMAD_RAILSSERVER_PORT}; proxy_pass \$zammad_railsserver_url;#g" \
-      -e "s#proxy_pass http://zammad-websocket;#set \$zammad_websocket_url http://${ZAMMAD_WEBSOCKET_HOST}:${ZAMMAD_WEBSOCKET_PORT}; proxy_pass \$zammad_websocket_url;#g" \
+  sed_args=(
+    -e "s#\(listen\)\(.*\)80#\1\2${NGINX_PORT}#g" \
+    -e "s#proxy_set_header X-Forwarded-Proto .*;#proxy_set_header X-Forwarded-Proto ${NGINX_SERVER_SCHEME};#g" \
+    -e "s#proxy_pass http://zammad-railsserver;#set \$zammad_railsserver_url http://${ZAMMAD_RAILSSERVER_HOST}:${ZAMMAD_RAILSSERVER_PORT}; proxy_pass \$zammad_railsserver_url;#g" \
+    -e "s#proxy_pass http://zammad-websocket;#set \$zammad_websocket_url http://${ZAMMAD_WEBSOCKET_HOST}:${ZAMMAD_WEBSOCKET_PORT}; proxy_pass \$zammad_websocket_url;#g" \
+    -e "s#server_name .*#server_name ${NGINX_SERVER_NAME};#g" \
+    -e "s#client_max_body_size .*#client_max_body_size ${NGINX_CLIENT_MAX_BODY_SIZE};#g" \
+    -e 's#/var/log/nginx/zammad.\(access\|error\).log#/dev/stdout#g'
+  )
+
+  if [ "${ZAMMAD_EXT_PROXY}" = "true" ]; then
+    GATEWAY=$(ip route | awk '/default/ {print $3}')
+    echo "Detected gateway: ${GATEWAY}"
+
+    sed_args+=(
+      -e "s#server_name .*;#server_name ${NGINX_SERVER_NAME};\n  # Trust Docker network gateway\n  set_real_ip_from ${GATEWAY};\n  real_ip_header X-Forwarded-For;\n  real_ip_recursive on;#g" \
+    )
+  else
+    sed_args+=(
       -e "s#server_name .*#server_name ${NGINX_SERVER_NAME};#g" \
-      -e "s#client_max_body_size .*#client_max_body_size ${NGINX_CLIENT_MAX_BODY_SIZE};#g" \
-      -e 's#/var/log/nginx/zammad.\(access\|error\).log#/dev/stdout#g' < contrib/nginx/zammad.conf > /etc/nginx/sites-enabled/default
+    )
+  fi
+
+  sed "${sed_args[@]}" < contrib/nginx/zammad.conf > /etc/nginx/sites-enabled/default
 
   #
   # Once we can use an nginx version >= 1.27.3, we can drop the proxy_pass workaround above and


MrGeneration:
This sounds use case specific, where you have more than one proxy in place and thus don't have Zammads nginx proxy container in the front line. In my opinion this shouldn't be done "by default".

stonerl:
Hm, I'm running Zammad's default Docker Compose with a bare-metal Nginx in front as a proxy. Is this a specific use case? I thought that is more or less the default. But I could add a ENV to toggle this behavior.

MrGeneration:
Zammads default setup is a 1 Proxy in front, not two. This implies several configurations that deviate from "our default" so to say.

stonerl:
I understand, thanks for clarifying. If I were to set this up behind an environment variable, for instance, `ZAMMAD_EXT_PROXY`, would this PR go in, or should I keep it in my local setup?

stonerl:
@MrGeneration I added a new ENV `ZAMMAD_EXT_PROXY` by default it is set to `false`

dominikklein:
@mgruner Maybe you can check this one.

ralf401:
There is also a related [documentation PR](https://github.com/zammad/zammad-documentation/pull/624). If this got merged I would also merge it for documentation, if there are no objections.

dominikklein:
@mgruner Any chance that we can check this one here? Can we bring this forward? Otherwise we should maybe close it.

mgruner:
@stonerl thank you very much for your contribution, and sorry for the delay so far. This seems to be a complex matter. 

Let me try to check if I understood the problem correctly: Zammad does not detect the client IP correctly, because the docker network gateway IP address and/or the nginx container IP address is not known via `RAILS_TRUSTED_PROXIES`. Furthermore, containers don't have static IP addresses in docker (at least by default) - these might change when container restarts happen.

Before we go deeper into possible solutions: Did you try adding the network gateway IP address via `RAILS_TRUSTED_PROXIES`? Would this technically work?

A different case might be using Traefik: https://community.zammad.org/t/zammad-uses-local-ip-of-the-reverse-proxy-traefik-instead-of-the-client-ip/10481/3 - from the report there, the actual traefik container IP gets passed to the stack.

stonerl:
@mgruner No worries.

Yes, I already added the host IP to `RAILS_TRUSTED_PROXIES` but that did not work. I also cannot switch to Traefik (and also have not tested), since we do run other services on that server, that “require” Nginx.

stonerl:
Sorry, I forgot to add more details.

So we are running an external Nginx proxy in front of Zammad. The client IPs are always reported as the Dockers gateway `178.x.x.x` I already added the host IP to `RAILS_TRUSTED_PROXIES` because I thought that might fix it, but it did not. That is why I came up with this solution.

Currently, I run a cronjob on the host that runs every hour to patch the nginx config inside the container (In case the container gets restarted). But on my test system I'm using the config option I opened the PR for here.

mgruner:
@stonerl is the (reported) gateway IP a static one, or is it also changing?

And I didn't want to suggest that you switch to Traefik, just wanted to add it as another use case for the problem.

stonerl:
@mgruner The reported one is always static (the gateway of the Zammad network, in our case `172.18.0.1`).

> And I didn't want to suggest that you switch to Traefik, just wanted to add it as another use case for the problem.

I did not assume that 😄, no worries.

mgruner:
> @mgruner The reported one is always static (the gateway of the Zammad network, in our case `172.18.0.1`).

Hm. What's then in the `X-Forwarded-For` of the requests, do you know or can find it out? Does it contain the IP address of the nginx container, or the gateway address, or both?

This header is used by https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html together with the `RAILS_TRUSTED_PROXIES` configuration to determine the actual client IP.

Another approach for a solution that is not tied to nginx internals might be that we resolve any hostnames in that variable in `config/environments/production.rb` - there is already some related code in there. But it would require DNS resolution capabilities for these hostnames. 🤷

mgruner:
> Another approach for a solution that is not tied to nginx internals might be that we resolve any hostnames in that variable in `config/environments/production.rb` - there is already some related code in there. But it would require DNS resolution capabilities for these hostnames. 🤷

Hm, but this would also not work correctly when there are nginx container restarts with IP change…

stonerl:
@mgruner If it does help, this is the config for the Nginx proxy:

```
server {
    listen 134.xx.xx.xx:80 default_server;
    server_name zammad.xxx.de;

    location / {
        return 301 https://$host$request_uri;
    }
}

server {
    listen 134.xx.xx.xx:443 ssl http2;
    server_name zammad.xxx.de;

   # Upload limit per file
   client_max_body_size 20M;

    location / {
        proxy_pass  http://127.0.0.1:8080;
        proxy_set_header Host   $host;

        # Avoid CSRF Token errors
        proxy_set_header  X-Forwarded-Proto https;

        # Pass real client IP to Zammad
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

        # WebSockets support
        proxy_set_header  X-Forwarded-Ssl on;
        proxy_set_header  X-Forwarded-Port $server_port;
        proxy_set_header  X-Forwarded-Host $host;

        # Buffer size limits to prevent issues
        proxy_buffer_size 256k;
        proxy_buffers 8 512k;
        proxy_busy_buffers_size 512k;

        # Timeout settings
        proxy_connect_timeout 60s;
        proxy_send_timeout 60s;
        proxy_read_timeout 60s;

        # Enable gzip
        gzip on;
        gzip_types text/plain text/xml text/css image/svg+xml application/javascript application/x-javascript application/json application/xml;
        gzip_proxied any;
    }

    # WebSocket Proxy (legacy and action cable)
    location /ws {
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "Upgrade";
        proxy_set_header Host $http_host;
        proxy_set_header CLIENT_IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto https;
        proxy_read_timeout 86400;
        proxy_buffering off;
        proxy_pass http://127.0.0.1:8080;
    }

    location /cable {
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "Upgrade";
        proxy_set_header Host $http_host;
        proxy_set_header CLIENT_IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto https;
        proxy_read_timeout 86400;
        proxy_buffering off;
        proxy_pass http://127.0.0.1:8080;
    }

    # Cache static files
    location ~ ^/(assets/|robots.txt|humans.txt|favicon.ico|apple-touch-icon.png) {
        expires max;
        proxy_pass http://127.0.0.1:8080;
    }

    # modern configuration
    ssl_protocols TLSv1.2 TLSv1.3;
    ssl_ecdh_curve X25519:prime256v1:secp384r1;
    ssl_ciphers ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-CHACHA20-POLY1305;
    ssl_prefer_server_ciphers off;

    # HSTS (ngx_http_headers_module is required) (63072000 seconds)
    add_header Strict-Transport-Security "max-age=63072000" always;

    # Enable OCSP Stapling
    ssl_stapling on;
    ssl_stapling_verify on;
    resolver 134.x.x.1 134.x.x.2 valid=300s;
    resolver_timeout 5s;

    ssl_session_timeout 1d;
    ssl_session_cache shared:MozSSL:10m;  # about 40000 sessions
    ssl_session_tickets off;

    ssl_certificate /etc/ssl/certs/zammad_xxx_de.crt;
    ssl_certificate_key /etc/ssl/private/zammad_xxx_de.key;
    ssl_trusted_certificate /etc/ssl/certs/xxx_de_chain.pem;
}
```

mgruner:
@stonerl I tried to reproduce your issue locally. For that, I used an nginx-proxy-manager container in `network_mode: host`, and connected it to the Zammad stack by defining a proxy host which redirects to `localhost:8080`. I was able to see the issue you described: instead of the client IP, the gateway IP of the Zammad network showed up instead.

Then, I added this gateway IP via `RAILS_TRUSTED_PROXIES="172.19.0.1"` to `.env`, and afterwards the addresses were shown correctly again. Since this gateway address is static and does not change on container restarts, this seems to be an acceptable solution to me.

Could you please let me know if this works for you as well, or otherwise, how it misbehaves if you configure `RAILS_TRUSTED_PROXIES` properly?

mgruner:
There is now a new issue for the problem with dynamic IP addresses: #5680.

mgruner:
I believe https://github.com/zammad/zammad/issues/5680 provides an acceptable solution for dealing with dynamic proxy  IP addresses, and static network gateway addresses can be specified like before.

Closing this proposal now. Thanks a lot for your contribution @stonerl, it helped getting this situation improved.

PR #5536, opened 2025-03-29
Add support for "is less than" and "is greater than" operators

jsuchal:
This adds support for `is less than` and `is greater than` conditions for integer fields. This was surprisingly easy to add for triggers/schedulers.

I would need help with testing since I am not (yet) familiar with the code base.

This was requested here https://community.zammad.org/t/less-than-and-greater-than-conditions-for-integers/4358 and also yesterday on webinar :)


jsuchal:
Seems like a flickering test failure.

rolfschmidt:
@dominikklein @mantas @dvuckovic what do you think about this one?

In general, I think the operators for the integer field are shitty. We should definitely enhance this. The only big question is do we only want **gt** and **lt** or also → **lte** and **gte** ? (I am voting for 4 variants)

The work diff shouldn't be that much. We also need some tests and core workflow support, but then this one would be a nice addition too.

rolfschmidt:
**2025-04-25 Daily**: we want this with 4 op's.

Issue #3782, opened 2021-10-01
Accented characters are not considered in search


benrubson:
Hi,

### Infos:

* Used Zammad version: 4.1.0-1632223866.38f67335.buster amd64
* Installation method (source, package, ..): package
* Operating system: Debian 10
* Database + version: mariadb-server 10.3.29-0+deb10u1 amd64
* Elasticsearch version: elasticsearch-oss 7.10.2 amd64
* Browser + version: Safari, Firefox (lastest versions)

### Expected behavior:

* Accented characters are considered in research, for example looking for _probl**e**m_ will return tickets containing _probl**e**m_, _probl**è**m_, _probl**é**m_... words.

### Actual behavior:

* Accented characters are not considered in research, for example looking for _probl**e**m_ will not return tickets containing _probl**è**m_, _probl**é**m_... words.
Which of course is a problem :) Possible workaround is to look for _probl?m_, but it's not really user-friendly.

### Steps to reproduce the behavior:

* Create a ticket containing _problèm_ word. And then look for _problem_.
* The other way also triggers the issue : create a ticket containing _problem_ word. And then look for _problèm_.

### Possible solution:

As per my findings, and if research is fully handled by Elasticsearch, `asciifolding` filter option in Elasticsearch could be the trick.

Many thanks 👍

MrGeneration:
Sorry, but this seems to be a feature request.
Please note that this repository is the wrong place for feature requests.
Instead, please refer the Zammad community over at: https://community.zammad.org .

In case the community provides enough votes (hearts on the first article, so feels the same way like you), we will move the feature request to this repository.

If you require a feature *now* or want to sponsor functions, please feel free to contact our sales team: sales [at] zammad [dot] com.

benrubson:
https://community.zammad.org/t/consider-accented-characters-in-research/7805

benrubson:
Many thanks @mantas for your interest into this 👍

YetAnotherGerrit:
Will be solved with https://github.com/zammad/zammad/pull/5537 in Zammad 7.0

PR #5537, opened 2025-03-29
Add support for optional Elasticsearch asciifolding analyzer

jsuchal:
This adds optional setting for searching in elasticsearch ignoring diacritics/special characters. This is imo expected behavior in almost all cases when user searches for text. 

Feature was requested here https://community.zammad.org/t/consider-accented-characters-in-research/7805 and here https://github.com/zammad/zammad/issues/3782

Example:

Ticket with title/content `Ružovučký koníček a sôvä spia` should be matched against queries like `ruz`, `sova`, `konicek` and vice-versa. Basically - diacritics (umlauts, accents...) should be ignored.

Example showing this works both ways
image

### DIFF
--- a/db/migrate/20250329155232_add_es_asciifolding_setting.rb
+++ b/db/migrate/20250329155232_add_es_asciifolding_setting.rb
@@ -1,0 +1,16 @@
+class AddEsAsciifoldingSetting < ActiveRecord::Migration[7.2]
+  def up
+    # return if it's a new setup
+    return if !Setting.exists?(name: 'system_init_done')
+
+    Setting.create_if_not_exists(
+      title:       __('Elasticsearch S Configuration'),
+      name:        'es_asciifolding',
+      area:        'SearchIndex::Elasticsearch',
+      description: __('Define if asciifolding analyzer should be used in Elasticsearch.'),
+      state:       false,
+      preferences: { online_service_disable: true },
+      frontend:    false
+    )
+  end
+end
--- a/db/migrate/20250329155232_add_es_asciifolding_setting.rb
+++ b/db/migrate/20250329155232_add_es_asciifolding_setting.rb
@@ -1,3 +1,5 @@
+# Copyright (C) 2012-2025 Zammad Foundation, https://zammad-foundation.org/
+
 class AddEsAsciifoldingSetting < ActiveRecord::Migration[7.2]
   def up
     # return if it's a new setup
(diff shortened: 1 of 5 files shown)


dominikklein:
@rolfschmidt Any thoughts about this one?

rolfschmidt:
@dominikklein looks cool, maybe we can even default that behaviour if both ways are supported. I did not play around with that filters yet so not sure if they bring any side effects. I would need to do some tests to understand it more.

---

https://www.elastic.co/docs/reference/text-analysis/analysis-asciifolding-tokenfilter
https://www.elastic.co/docs/reference/text-analysis/analysis-lowercase-tokenfilter

jsuchal:
@rolfschmidt I would advise against going for a global default on this. If zammad wants to support global users using non latin characters this will not work for them. 

For users that use latin character this is probably a good default. I don't really remember a case when someone wanted to search really precisely and diacritics were significant. And I do have a lot of background in search engines.

rolfschmidt:
> @rolfschmidt I would advise against going for a global default on this. If zammad wants to support global users using non latin characters this will not work for them.
> 
> For users that use latin character this is probably a good default. I don't really remember a case when someone wanted to search really precisely and diacritics were significant. And I do have a lot of background in search engines.

But can't you still do an exact search via `.keyword` if you want to?

image

image

mantas:
IIRC asciifolding analyzer would just ignore non-latin characters. It should be safe to enable. But it needs testing with some non-latin scripts to be sure.

rolfschmidt:
> IIRC asciifolding analyzer would just ignore non-latin characters. It should be safe to enable. But it needs testing with some non-latin scripts to be sure.

Currently, there are 3 options:

1) Ignore this PR, since you could solve it via rails console (es_model_settings).
2) Merge the PR, but then it is a hidden 1% setting which nearly nobody will use.
3) Merge the PR and set as default `true`.

For 3. it would be interesting how many people would turn that off. I think it's all about a good default state for this stuff. In general rebuild the ES for zammad is common, and I think it is not a big deal.

--

My opinion → 3: Merge and activate this behaviour, so default on. Maybe we could add a documentation section for non-latin users to consider disabling this on installation, if they don't like it.

dvuckovic:
Just my 2 cents, spoken from personal experience: Serbian language uses both Cyrillic and Latin scripts interchangeably. However, only the Latin alphabet can be "asciifolded", in Cyrillic it makes no sense.

In case the indexed content is in Cyrillic, it normally has to be transliterated first into Latin, and then possibly "asciifolded". (But this is a whole different set of problems.)

For content in Latin, however, it *is* something that is normally done by users when searching, nobody changes their keyboard input just for the search.

So 100% for the default value of `true` from my side. But do allow for it to be disabled in case of any issues, providing a good example in the docs.

mantas:
I'm also 💯  for default `true`. And merging, this is a very nice addition in many latin languages.

PR #5545, opened 2025-04-02
Handle multiple dns server in resolve conf

valodzka:
Current docker-entrypoint.sh fails with nginx syntax error if resolv.con contains > 1 dns server.
This fixes the issues by taking only first server

### DIFF
--- a/contrib/docker/docker-entrypoint.sh
+++ b/contrib/docker/docker-entrypoint.sh
@@ -114,7 +114,7 @@ elif [ "$1" = 'zammad-nginx' ]; then
   # configure nginx
 
   # Ensure that nginx has a short TTL so that recreated containers with new IP addresses are found.
-  NAMESERVER=$(grep "nameserver" < /etc/resolv.conf | awk '{print $2}')
+  NAMESERVER=$(grep -m1 "nameserver" < /etc/resolv.conf | awk '{print $2}')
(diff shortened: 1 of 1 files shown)


rkoehn:
I have a similar problem (even slightly more complex) problem. My docker generated `resolv.conf` looks like this:
```
zammad@471b0f446d2a:~$ cat /etc/resolv.conf
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.

nameserver 192.168.178.1
nameserver 8.8.8.8
search fritz.box

# Based on host file: '/etc/resolv.conf' (legacy)
# Overrides: [nameservers]
```
In this case, not only two nameservers are used, but additionally the comment `# Overrides: [nameservers]` is picked.

Potentially safer solution would be to first remove all comments (`grep -v "#" `):
```
NAMESERVER=$(grep -v "#" < /etc/resolv.conf | grep -m1 "nameserver" | awk '{print $2}')
```

rkoehn:
@valodzka , I think this problem may have been fixed in https://github.com/zammad/zammad/issues/5553 ?

Related commit: https://github.com/zammad/zammad/commit/a77cc70b7ba53dd8529c13563ca81caf49effc06

valodzka:
Yes, looks like so

PR #5582, opened 2025-04-23
Feature: Modernize developer workflow.

Stackout:
Modernize Developer Workflow with Docker-Based Dev Mode
This PR improves the development experience by allowing contributors to run Zammad entirely within Docker—no need to install Ruby, Node, pnpm, or other tools on the host machine.

✅ What’s included

- Adds conditional installation of dev-only tools (pnpm, forego, golang, etc.) when RAILS_ENV=development.
- Preserves the small image size for production.
- Ensures full frontend + backend development is possible from within a single container.
- Automatically runs pnpm install on container startup to support the Vite dev server.
- Allows Vite to bind to 0.0.0.0 via ZAMMAD_BIND_IP so it's accessible externally.

🤝 Why this is useful

- No local setup is needed—just clone and run ./dev.sh.
- Great for contributors who prefer isolated dev environments.
- More consistent onboarding and testing experience.

🙏 Request for feedback
I tried to keep things clean, minimal and consistent, but I may be missing something. Please let me know if there are better places to hook this in or any conventions I should align with.

Thanks!

### DIFF
--- a/.dev/.env.dist
+++ b/.dev/.env.dist
@@ -1,0 +1,27 @@
+RAILS_ENV=development
+REDIS_URL=redis://localhost:6379
+
+# Zammad configuration.
+
+ZAMMAD_BIND_IP=0.0.0.0
+DB_ADAPTER=postgresql
+DB_PASSWORD=zammad
+REDIS_URL=redis://redis:6379
+ELASTICSEARCH_URL=http://elasticsearch:9200
+DB_ADAPTER=postgresql
+DB_HOST=zammad-postgresql
+DB_NAME=zammad
+DB_USER=zammad
+DB_PASSWORD=zammad
+
+# PostgreSQL configuration.
+
+POSTGRESQL_PASS=zammad
+POSTGRESQL_USER=zammad
+POSTGRESQL_HOST=zammad-postgresql
+POSTGRESQL_PORT=5432
+POSTGRESQL_DB=zammad
+POSTGRESQL_OPTIONS=
+
+# Elasticsearch configuration.
+ELASTICSEARCH_HOST=elasticsearch
--- a/.dev/docker-compose.yml
+++ b/.dev/docker-compose.yml
@@ -1,0 +1,65 @@
+services:
+  zammad-postgresql:
+    image: postgres:17
+    environment:
+      POSTGRES_USER: zammad
+      POSTGRES_PASSWORD: zammad
+      POSTGRES_DB: zammad
+    volumes:
+      - zammad_pg_data:/var/lib/postgresql/data
+    ports:
+      - '5432:5432'
+
+  redis:
+    image: redis:7
+    volumes:
+      - zammad_redis_data:/data
+    ports:
+      - '6379:6379'
+
+  zammad-elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.17.6
+    environment:
+      - discovery.type=single-node
+      - ES_JAVA_OPTS=-Xms1g -Xmx1g
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    mem_limit: 2g
+    ports:
+      - '9200:9200'
+    volumes:
+      - zammad_elasticsearch_data:/usr/share/elasticsearch/data
+      - ./es-plugins.sh:/usr/share/elasticsearch/plugins_install.sh
+    entrypoint: ["/bin/bash", "-c", "/usr/share/elasticsearch/plugins_install.sh && /usr/local/bin/docker-entrypoint.sh"]
+
+  zammad:
+    image: zammad-dev:latest
+    command: pnpm run dev
+    healthcheck:
+      test: ["CMD-SHELL", "curl -fs http://localhost:3000 || exit 1"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    build:
+      context: ..
+      args:
+        COMMIT_SHA: ${COMMIT_SHA:-dev}
+        RAILS_ENV: "development"
+    depends_on:
+      - zammad-postgresql
+      - redis
+      - zammad-elasticsearch
+    env_file: .env
+    ports:
+      - '3000:3000'
+      - '6042:6042'
+      - '3036:3036'
+    volumes:
+      - ..:/opt/zammad
+
+volumes:
+  zammad_pg_data:
+  zammad_redis_data:
+  zammad_elasticsearch_data:
--- a/.dev/es-plugins.sh
+++ b/.dev/es-plugins.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+
+PLUGINS=(
+  "ingest-attachment"
+  # Add more here if needed in future
+)
+
+for plugin in "${PLUGINS[@]}"; do
+  if ! elasticsearch-plugin list | grep -q "$plugin"; then
+    echo "Installing plugin: $plugin"
+    elasticsearch-plugin install --batch "$plugin"
+  else
+    echo "Plugin already installed: $plugin"
+  fi
+done
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,8 @@ RUN contrib/docker/setup.sh builder
 FROM ruby:3.3.8-slim-bookworm
 ARG DEBIAN_FRONTEND=noninteractive
 ARG ZAMMAD_USER=zammad
-ENV RAILS_ENV=production
+ARG RAILS_ENV=production
+ENV RAILS_ENV=${RAILS_ENV}
 ENV RAILS_LOG_TO_STDOUT=true
 ENV ZAMMAD_DIR=/opt/zammad
 
@@ -30,6 +31,26 @@ COPY --from=builder ${ZAMMAD_DIR} .
 COPY --from=builder /usr/local/bundle /usr/local/bundle
 COPY --from=builder ${ZAMMAD_DIR}/contrib/docker/docker-entrypoint.sh /
 RUN contrib/docker/setup.sh runner
+
+RUN if [ "$RAILS_ENV" = "development" ]; then \
+  apt-get update && \
+  apt-get install -y --no-install-recommends \
+    nodejs \
+    npm \
+    curl \
+    ca-certificates \
+    git \
+    golang && \
+  npm install -g corepack && \
+  corepack enable && \
+  corepack prepare pnpm@10.9.0 --activate && \
+  mkdir -p /tmp/go/src/github.com/ddollar && \
+  git clone https://github.com/ddollar/forego.git /tmp/go/src/github.com/ddollar/forego && \
+  cd /tmp/go/src/github.com/ddollar/forego && \
+  GOPATH=/tmp/go GO111MODULE=off go build -o /usr/local/bin/forego && \
+  chmod +x /usr/local/bin/forego && \
+  rm -rf /tmp/go /var/lib/apt/lists/*; \
+fi
 
 USER zammad:zammad
 ENTRYPOINT ["/docker-entrypoint.sh"]
--- a/dev.sh
+++ b/dev.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -euo pipefail
+
+PROJECT_NAME=zammad-dev
+COMPOSE_FILE=".dev/docker-compose.yml"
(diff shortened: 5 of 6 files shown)


contrib/docker/docker-entrypoint.sh
```diff
@@ -49,6 +49,8 @@ function check_zammad_ready {
 
 # zammad init
 if [ "$1" = 'zammad-init' ]; then
+  echo "📡 DATABASE_URL: $DATABASE_URL"
```
Stackout: I'll remove this I had this in here for debugging purposes.

dev.sh
```diff
@@ -0,0 +1,29 @@
...
+done
+
+echo "🔧 Building and starting containers..."
+COMMIT_SHA=$(git rev-parse HEAD) RAILS_ENV=development docker-compose up -d --remove-orphans --build
+
+echo "⚙️  Running one-shot DB initializer..."
+docker compose run --rm zammad zammad-init
+
+echo "✅ Zammad is running at: http://localhost:3000"
+echo "🧠 Default login: admin@example.com / test"
```
Stackout: I'll also remove this bit here, I made an assumption initially writing this that wasn't true.

es-plugins.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -e
+
+PLUGINS=(
+  "ingest-attachment"
```
Stackout: I had a problem running this via docker initially, so I felt adding an es-plugins.sh was a good hook to setup the plugins.

Stackout:
I will keep this as a single commit with zero noise, and ensure the commit is atomic and stands on its own. Thanks for considering this PR!

dominikklein:
@mgruner @dvuckovic Any thoughts?

mgruner (changes requested):
@Stackout thanks a lot for your proposal. Looks very promising! Before we look at further details, I would ask if we can move all new files to a dedicated subdirectory like `.dev/docker` or similar, to keep the codebase clean and not confuse people with a development focused `docker-compose.yml`.

Stackout:
@mgruner Thanks for the feedback! I’ve moved all development-related files into .dev/docker as requested to keep the root clean. I did keep dev.sh at the root to preserve the ease-of-use for developers — just clone and run. Let me know if you'd prefer that moved as well.

.dev/.env.dist
```diff
@@ -0,0 +1,27 @@
+RAILS_ENV=development
+REDIS_URL=redis://localhost:6379
```
mgruner, line 2: Duplicate, see below

.dev/.env.dist
```diff
@@ -0,0 +1,27 @@
...
+
+POSTGRESQL_PASS=zammad
+POSTGRESQL_USER=zammad
+POSTGRESQL_HOST=zammad-postgresql
+POSTGRESQL_PORT=5432
+POSTGRESQL_DB=zammad
+POSTGRESQL_OPTIONS=
+
+# Elasticsearch configuration.
+ELASTICSEARCH_HOST=elasticsearch
```
mgruner, line 27: Should this be zammad-elasticsearch?

.dev/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    environment:
+      POSTGRES_USER: zammad
+      POSTGRES_PASSWORD: zammad
+      POSTGRES_DB: zammad
+    volumes:
+      - zammad_pg_data:/var/lib/postgresql/data
+    ports:
+      - '5432:5432'
+
+  redis:
```
mgruner, line 13: Let's call this `zammad-redis` to synchronize the service names with https://github.com/zammad/zammad-docker-compose/blob/master/docker-compose.yml

.dev/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    image: redis:7
+    volumes:
+      - zammad_redis_data:/data
+    ports:
+      - '6379:6379'
+
+  zammad-elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.17.6
```
mgruner, line 21: Let's use the same image as in https://github.com/zammad/zammad-docker-compose/blob/master/docker-compose.yml here, please. Then we should also not need the plugin install script any more, AFAIR.

.dev/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    ports:
+      - '9200:9200'
+    volumes:
+      - zammad_elasticsearch_data:/usr/share/elasticsearch/data
+      - ./es-plugins.sh:/usr/share/elasticsearch/plugins_install.sh
+    entrypoint: ["/bin/bash", "-c", "/usr/share/elasticsearch/plugins_install.sh && /usr/local/bin/docker-entrypoint.sh"]
+
+  zammad:
+    image: zammad-dev:latest
+    command: pnpm run dev
```
mgruner, line 39: I run this on an arm Mac. It mounts my node_modules from the host, and then vite keeps erroring out:

```plain
forego      | starting vite.1 on port 7100
vite.1      | /opt/zammad/node_modules/.pnpm/rollup@4.36.0/node_modules/rollup/dist/native.js:64
vite.1      | 		throw new Error(
vite.1      | 		      ^
vite.1      | 
vite.1      | Error: Cannot find module @rollup/rollup-linux-arm64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
vite.1      |     at requireWithFriendlyError (/opt/zammad/node_modules/.pnpm/rollup@4.36.0/node_modules/rollup/dist/native.js:64:9)
vite.1      |     at Object.<anonymous> (/opt/zammad/node_modules/.pnpm/rollup@4.36.0/node_modules/rollup/dist/native.js:73:76)
```

.dev/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+      - '3000:3000'
+      - '6042:6042'
+      - '3036:3036'
+    volumes:
+      - ..:/opt/zammad
+
+volumes:
+  zammad_pg_data:
+  zammad_redis_data:
+  zammad_elasticsearch_data:
```
mgruner, line 65: Let's align the volume names with https://github.com/zammad/zammad-docker-compose/blob/master/docker-compose.yml, please. There is no `zammad` prefix needed, as they are prefixed with the stack name.

.dev/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
+services:
```
mgruner, line 1: Perhaps it would be useful to also add the stack name here, and make it configurable via `.env`. Otherwise it will be `dev` now always, due to the folder name.

.dev/.env.dist
```diff
@@ -0,0 +1,27 @@
+RAILS_ENV=development
```
mgruner, line 1: Can we move the files to `.dev/docker/` perhaps? That may be easier to identify.

.dev/es-plugins.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
mgruner, line 1: IIRC we can drop this after switching to ES 8.

Dockerfile
```diff
@@ -31,6 +32,26 @@ COPY --from=builder /usr/local/bundle /usr/local/bundle
 COPY --from=builder ${ZAMMAD_DIR}/contrib/docker/docker-entrypoint.sh /
 RUN contrib/docker/setup.sh runner
 
+RUN if [ "$RAILS_ENV" = "development" ]; then \
+  apt-get update && \
```
mgruner, line 36: Let's perhaps move this to a script in `.dev/docker` as well, to keep this file shorter.

dev.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+echo "🧹 Removing previous containers..."
+docker-compose -f "$COMPOSE_FILE" down --volumes --remove-orphans || true
+
+echo "🔧 Building and starting containers..."
+COMMIT_SHA=$(git rev-parse HEAD) RAILS_ENV=development docker-compose -f "$COMPOSE_FILE" up -d --remove-orphans
+
+echo "⚙️  Running one-shot DB initializer..."
+docker compose -f "$COMPOSE_FILE" run --rm zammad zammad-init
```
mgruner, line 23: This does run the DB init for me, but then hangs with `zammad-init waiting for elasticsearch server to be ready…` repeating forever - maybe because of wrong hostname (see .env)?

dev.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
mgruner, line 1: Would be good to also move this to `.dev/docker/`. We can document all this in the developer manual `doc/`.

mgruner (changes requested):
@Stackout again, thanks for your proposal! I tried it the first time, and sent you some first feedback. There might be more in future as we go deeper into the details.

At this point I cannot give you a guarantee that this will get merged. I will present it to my colleagues and ask for their feedback as well, to understand if this is a direction the team wants to take. But so far it does look promising to me!

Dockerfile
mgruner: There should probably be at least `./dev/docker/start.sh` and `stop.sh` as well. Maybe also a `restart.sh` or a `rebuild.sh`, but those could also be added in a follow-up step, if they are needed (eg. to run after dependency update).

dev.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -euo pipefail
+
+PROJECT_NAME=zammad-dev
```
mgruner, line 5: Unused? Maybe you wanted to use this as stack name? Then perhaps move it to .env?

.dev/docker-compose.yml
Stackout: Thank you! I'll work on getting this building on my arm Mac as well. I only tested on Linux.

.dev/.env.dist
Stackout: Sure thing!

dev.sh
Stackout: Yeah, I'll remove this. No problem!

dev.sh
Stackout: For sure, I liked the ease of use having it in the root to just get everything running. I came back to running it today and had an issue with PIDs, even on a fresh reboot.

.dev/docker-compose.yml
Stackout: I'm new to Zammad, so I apologize I should've seen this zammad-docker repo and aligned everything that you've done there first. Thank you so much for the feedback!

dev.sh
Stackout: I'm going to pull this over to my ARM Mac and get it building there. Most folks are probably not using Ubuntu/linux for dev work. I just happened to be working on my Linux box. I'll make some updates over the weekend. No worries if this doesn’t make it in—this has already been super helpful for avoiding local dependency installs. Definitely a cross platform bug. here.

mgruner: Colleages have signalled interest in this. There are already some private solutions, so your PR could finally be a starting point for an official, shared solution that we can improve together, step by step.

.dev/.env.dist
mgruner, line 27: Perhaps another thought about this .env file. I'm not sure if it is really needed, as everything is really internal to the stack.
Another idea might be to move these settings to a shared, static definition in the .yml file like we do in zammad-docker-compose which is used by the services. Perhaps at least for those settings where a modification does not make sense (such as ES hostname). But perhaps for all?
Having the ability to pass in custom stuff for the local developer's stack in addition to that might still be a good idea.
Just an idea, though.

Stackout: Got it! I'll go back and review your zammad-docker repo and follow prior art there. I'm working on adding a feature to Zammad and did this work so I could get it up and running and tear it down when I don't need it since I'm usually in 50+ repos at a time, Docker is a saving grace for development 🙏

mgruner:
Hey @Stackout. We managed to add a first, basic level of support for devcontainer in https://github.com/zammad/zammad/commit/9ebdbc20c594e61415ba0d8b977969d6ad063e40. I guess this will be the way forward when optimizing dev workflow in future. Please feel free to try this out, and collaborate on this basis if you see room for improvement.

Thanks for your work on this, and for nudging us in this direction.

@dominikklein FYI

PR #5614, opened 2025-05-12
backup - implemented using pigz as a optional alternative to gzip for file (tar) and database backups

zerwes:
This PR allows using [pigz](https://zlib.net/pigz/) as a alternative to good old gzip.
Especially on large setups this can speed up the backup, as pigz is capable of using multiple threads for compressing. Good old gzip is using just one CPU, regardless of how many might be available ..

We encountered issues after [migrating all of our jira issues into zammad](https://github.com/Rosa-Luxemburgstiftung-Berlin/jira2zammad).

The speed up should apply for both storage provider cases, `DB` or `File`.

After the merge of the PR, I can add another one, documenting the changes, to zammad-documentation.


MrGeneration:
I'd suggest that the script actually detects if `pigz` is available.
Because your approach doesn't ensure this, if I just "blindly" set your setting to yes, it will fail to execute successfully.

`pigz` is not a default binary in distribution installations - at least not on minimal and server environments as far as I'm aware. As the `pigz` archive throws out a `gzip` compatible archive, this shouldn't be a problem - even if restore and backup setups would have mixed setups.

Apart from that, I like the approach.

zerwes:
> I'd suggest that the script actually detects if `pigz` is available. Because your approach doesn't ensure this, if I just "blindly" set your setting to yes, it will fail to execute successfully.
> 
> `pigz` is not a default binary in distribution installations - at least not on minimal and server environments as far as I'm aware. As the `pigz` archive throws out a `gzip` compatible archive, this shouldn't be a problem - even if restore and backup setups would have mixed setups.
> 
> Apart from that, I like the approach.

In general, I consider users to be responsible, so if one says 'use gigz' I assume he will ensure pigz is installed, but you are right, pigz will not be installed by default, so a little check and fallback to a sane behavior will not hurt ...

Implemented in d3d115094

MrGeneration:
I'd prefer either, but not both. Imho, from your logic: If admin demands pigz to be used but it's not installed, it would be a user error and thus should make the script to exit with an error code.

Really, I'd prefer that switching to happen automatically without any configuration.
A hint in the documentation should be added after.

If you prefer, we can have @YetAnotherGerrit decide.

zerwes:
>  If admin demands pigz to be used but it's not installed, it would be a user error

would be OK for me too; the other way - hawing a valid working backup with a warning issued might be a good thing, if the exit code of the backup is not monitored (yes, I know, again a user error ...); I am fine with both approaches: warning and sane settings or hard and severe exit wit a error

> I'd prefer that switching to happen automatically without any configuration.

You mean something like "pigz is installed, so we will use it, otherwise fall back to good old gzip"? IMHO not that flexible and somehow patronizing ...

> A hint in the documentation should be added after.

:+1: as I wrote: as soon as the PR is settled, the method is decided and merged, I can create the corresponding PR in zammad-documentation.

YetAnotherGerrit:
Thank you for your contribution @zerwes ! <3

As long as `pigz` creates archives that can be restored with `gzip` no option should be needed. So if `pigz` is detected, the script can use it, otherwise it falls back to `gzip`.

Then the documentation just needs a hint/tipp that `pigz` can be installed to improve performance.

And if there are other alternatives for `gzip` (in the future) they can be added to the scripts as well. But that is for another day.

zerwes:
> As long as pigz creates archives that can be restored with gzip no option should be needed. So if pigz is detected, the script can use it, otherwise it falls back to gzip.

OK, straight ahead way implemented in 4f6a1a0
And yes, pigz uses the same compression alg, just a different implementation. 

> And if there are other alternatives for gzip (in the future) they can be added to the scripts as well. But that is for another day.

I started first a quite generic approach, allowing the users to choose between plain tar, gzip or any other external program to choose using `-I "...."`
https://github.com/zammad/zammad/compare/develop...Rosa-Luxemburgstiftung-Berlin:zammad:backup-generic-tar-compression
But generic sometimes ends up in too complicated ...

So I think for now the 2 options are sufficient ....

> Then the documentation just needs a hint/tipp that pigz can be installed to improve performance.

I will create a PR ...

zerwes:
https://github.com/zammad/zammad-documentation/pull/641

MrGeneration:
I prepared this PR internally in our Gitlab instance, tested and verified the functionality and adjusted minor things.
The PR is now subject to our internal Scrum stuff. I will sync the state back on here if anything moves, unless a colleague is faster than me. :-)

mgruner:
Thanks a lot @zerwes and @MrGeneration!

PR #5616, opened 2025-05-12
add mail header Resent-*

loppermann-de:
add a option to filter mails into groups where mail was redirected to


mantas:
Hm. Maybe we should add whole Resent-* headers family while at it.

loppermann-de:
Not a bad idea, I have added all of them.

mantas:
Thank you for your contribution @loppermann-de!  This is now merged with above commit.

PR #5621, opened 2025-05-14
Update EN password change notification

Gerharddc:
The current notification text for a password change may sound a bit strange to a native English speaker. This change tries to fix that.


dominikklein:
@mgruner @dvuckovic Maybe something you could check from a language perspective.

dvuckovic:
@Gerharddc Thanks for your contribution, your change will be included in the next release (probably Zammad 7.0).

Issue #5628, opened 2025-05-19
User/Ticket APIs are not idempotent causing duplicate/undeterministic results under concurrent load


jsuchal:
### Used Zammad Version

6.5.0-47

### Environment

- Installation method: [package, v 6.5.0-47]
- Operating system (not relevant]
- Database + version:  [not relevant]
- Elasticsearch version: [not relevant]
- Browser + version: [not relevant]

### Actual behaviour

We are in a process of migrating tickets from legacy system to zammad. We do this via API, since there are tens of thousands of tickets and we need to do this in parallel + retain history, etc. We have several workers importing tickets/ticket history in parallel. Sometime due to concurrency multiple workers try to create a new user (which is the same because it has communication history in multiple tickets).

This method runs before a user gets created via API (and GUI) and is very undeterministic in this case.

https://github.com/zammad/zammad/blob/develop/app/models/user.rb#L818-L846 

It basically generates a login by random - which causes duplicate users being created, etc. its basically a no-go.

Also during load sometimes zammad timeouts and returns an error leaving the import in an undeterministic state - we don't know if a ticket was created or not. 

The issue here is fundamental. The APIs are not idempotent. See https://martinfowler.com/articles/patterns-of-distributed-systems/idempotent-receiver.html 

Best case - you will have duplicate tickets.
Worst case -  you end up with duplicate users with "random" logins. 

As for users API we managed to monkeypatch the user api with 

![Image](https://github.com/user-attachments/assets/9e4b66d6-7d59-45e8-8c09-4f1a1051bfa0)

Since login must be unique this works as expected - if you try to create a user with the same login it fails. The job will retry and find that the user exists and continues.

### Expected behaviour

Ideally the APIs should be idempotent - e.g. we could generate a "uuid" on worker/client side before the record gets pushed to zammad via API. There should be a unique constraint on the zammad record to not allow creation of duplicate records and return a 403 Already Exists http code. Since the responsibility for generating unique uuids in on the API client side, this is expected behavior and the client knows that the record was created and can continue.

It will basically work the same way as we did with the login monkey-patch above. Since login has unique constraint we were able to "exploit it". For tickets it's tricky - we could reuse the unique ticket "number" and try to push it but not sure if this will not break zammad elsewhere.

This API change could be completely optional, if the API client is not sending the unique uuids, there is no change for current API users. 

### Steps to reproduce the behaviour

- Run multiple workers that try to create user with same login via API on zammad. 
[...]

PR #5631, opened 2025-05-20
No user creation when login is already taken, closes #4940

### LINKED ISSUE #4940: No user creation when login is already taken (via App/REST API, on create action)
### Used Zammad Version

6.2.x (develop)

### Environment

- Installation method: any
- Operating system (if you're unsure: `cat /etc/os-release` ): any
- Database + version:  any
- Elasticsearch version: any
- Browser + version: any

### Actual behaviour

Currently, when you create a user for which the login is already taken, there is a mechanism that will add a suffix at the end of the given login name.

For example, a user with a login/username `john.doe` exists already in the system and afterwards an additional user creation (e.g. via API) with the login/username `john.doe` will be triggered. This will lead to the login/username `john.doe1`.
afterward

### Expected behaviour

The user creation should fail with an error message.

For this the loop protection in the user model needs to be removed:

https://github.com/zammad/zammad/blob/7dbd1c1b15b967dfa45df98c8f80076572db1358/app/models/user.rb#L931-L936

For example with the user REST-API:
- Create user with login/username `john.doe`.
- Create afterwards a second user also with login/username `john.doe`.
- See that the login for this user is  `john.doe1`.

### Support Ticket

Ticket#10143308

yes

jsuchal:
#4940 #5628

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -823,14 +823,8 @@ try to find correct name
     end
 
     # check if login already exists
-    base_login = login.downcase.strip
-
-    alternatives = [nil] + Array(1..20) + [ SecureRandom.uuid ]
-    alternatives.each do |suffix|
-      self.login = "#{base_login}#{suffix}"
-      exists = User.find_by(login: login)
-      return true if !exists || exists.id == id
-    end
+    exists = User.find_by(login: login)
+    return true if !exists || exists.id == id
 
     raise Exceptions::UnprocessableEntity, "Invalid user login generation for login #{login}!"
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -806,7 +806,6 @@ try to find correct name
   end
 
   def check_login
-
     # use email as login if not given
     if login.blank?
       self.login = email
@@ -825,6 +824,9 @@ try to find correct name
     # check if login already exists
     exists = User.find_by(login: login)
     return true if !exists || exists.id == id
+
+    # we leave the email uniqueness check for ensure_uniq_email
+    return true if email.present?
 
     raise Exceptions::UnprocessableEntity, "Invalid user login generation for login #{login}!"
   end
--- a/spec/models/concerns/can_csv_import_user_examples.rb
+++ b/spec/models/concerns/can_csv_import_user_examples.rb
@@ -160,7 +160,7 @@ RSpec.shared_examples 'CanCsvImport - User specific tests', :aggregate_failures 
         it 'does import users' do
           expect { result }.to change(User, :count).by(2)
           expect(first_user).to have_attributes(
-            login:     'user-simple-import1',
+            login:     'user-simple-IMPORT1',
             firstname: 'firstname-simple-import1',
             lastname:  'lastname-simple-import1',
             email:     'user-simple-import1@example.com',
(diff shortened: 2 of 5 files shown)


app/models/user.rb
```diff
@@ -823,14 +822,11 @@ def check_login
...
+    return true if !exists || exists.id == id
 
-    alternatives = [nil] + Array(1..20) + [ SecureRandom.uuid ]
-    alternatives.each do |suffix|
-      self.login = "#{base_login}#{suffix}"
-      exists = User.find_by(login: login)
-      return true if !exists || exists.id == id
-    end
+    # we leave the email uniqueness check for ensure_uniq_email
+    return true if email.present?
```
jsuchal, line 829: this is needed since other parts of code expect activerecord errors when uniqueness check fails (e.g. email)

spec/models/concerns/can_csv_import_user_examples.rb
```diff
@@ -160,7 +160,7 @@
         it 'does import users' do
           expect { result }.to change(User, :count).by(2)
           expect(first_user).to have_attributes(
-            login:     'user-simple-import1',
+            login:     'user-simple-IMPORT1',
```
jsuchal, line 163: i believe we should not touch imported logins in any way.

jsuchal:
@dominikklein this is ready for review.

app/models/user.rb
```diff
@@ -823,14 +822,11 @@ def check_login
     end
 
     # check if login already exists
-    base_login = login.downcase.strip
+    exists = User.find_by(login: login)
```
mantas, line 825: This could be a `User.exists?`

app/models/user.rb
```diff
@@ -823,14 +822,11 @@ def check_login
...
+    return true if !exists || exists.id == id
```
mantas, line 826: On the other hand, maybe we should have a proper check with unique index and `validates_uniqueness_of` validation?

mantas: Actually, there's already an index for this. Yet no app-side check.

jsuchal: @mantas yes this is a weird pattern in the code, not sure why there is a exception thrown for this but i haven't touched this logic by-design. i don't have deep knowledge about what this might cause.

mantas:
After a deeper inspection, I'm not sure if this is viable change. The problem is quite complex.

- Sometimes Zammad allows non-unique emails.

- Email is used as a fallback for `login` attribute

- `login` must be unique

I guess you see where this is going :) That login iterator was making sure that accounts with non-unique emails can be created anyway.

@dominikklein @mgruner @dvuckovic any ideas?

jsuchal:
> After a deeper inspection, I'm not sure if this is viable change. The problem is quite complex.
> 
> * Sometimes Zammad allows non-unique emails.
> * Email is used as a fallback for `login` attribute
> * `login` must be unique
> 
> I guess you see where this is going :) That login iterator was making sure that accounts with non-unique emails can be created anyway.

Oh, that makes sense. In that case the fix should be something more like this mentioned here? https://github.com/zammad/zammad/issues/5628

dominikklein:
@mantas I did a look inside the topic again. Yes, for sure, when `user_email_multiple_use` is used, we need to have some mechanism that it's working in some way.

I think we need to decide if we stay with this loop for this situation, or we remove the loop and always use this `SecureRandom.uuid` addition.

dominikklein:
@jsuchal Thanks for your contribution :-) We merged it with https://github.com/zammad/zammad/commit/aaed6481e6ce5d6bf5e2d6cc3e3f4658d0de78d3

PR #5635, opened 2025-05-25
Sanitize FQDN by removing port to comply with RFC 5321 (HELO/EHLO)

athoik:
This change removes any port number from the FQDN before it is used in the SMTP HELO/EHLO command. According to RFC 5321 section 4.1.1.1, the domain name used in HELO/EHLO must be a valid FQDN without a port.

In some environments, the FQDN may include a port (e.g., xyz.example.com:8080), which leads to SMTP errors such as:

501 5.5.4 Invalid domain name

To address this, we strip the port using:

fqdn = fqdn.split(':').first

This ensures compatibility with SMTP servers and avoids invalid domain errors during mail delivery.

### DIFF
--- a/app/models/channel/driver/smtp.rb
+++ b/app/models/channel/driver/smtp.rb
@@ -57,6 +57,8 @@ class Channel::Driver::Smtp < Channel::Driver::BaseEmailOutbound
           fqdn = domain
         end
       end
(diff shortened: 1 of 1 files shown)


mantas:
Hey @athoik 

Thank you! Let me run this through internal CI...

mantas:
Hey @athoik 

Thank you for your contribution! This is now merged with above commit.

PR #5723, opened 2025-07-30
Create it.yml

albanobattistella:

### DIFF
--- a/i18n/text_modules/it.yml
+++ b/i18n/text_modules/it.yml
@@ -1,0 +1,115 @@
+---
+- name: cnr - Non sono riuscito a contattarti...
+  keywords: ''
+  content:
+    Non sono riuscito a contattarti telefonicamente. Potresti inviarmi una breve email e
+farmi sapere quando sarai di nuovo disponibile?
+  note: ''
+- name: iq - Se hai domande...
+  keywords: ''
+  content: Se avete domande, non esitate a contattarmi
+  note: ''
+- name: ifq - in caso di ulteriori domande...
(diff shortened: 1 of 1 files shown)


dominikklein:
@mgruner Are we adding such things?

mgruner:
@dk I'm not sure. Text modules have some limitations (no language dependency, no migration updateability) which we might need to solve first.

dominikklein:
Ah, true, I was thinking it's like with notification mails, but here it will create new data on initial setup, or?

Hm ok, then the decision is more difficult.

dk:
> @dk I'm not sure. Text modules have some limitations (no language dependency, no migration updateability) which we might need to solve first.

 I doubt you wanted to tag me so I'd like to remind that @ tags and spams people, in that particular case it was me. Good luck with your project )

mgruner:
Oh, I'm so sorry for the confusion!

mgruner:
@dominikklein yes, for new installations all text module files would be imported.

dominikklein:
@mgruner Like discussed in the chat, I think it would only be imported when `it` is detected as the language for the first admin use in the wizard.

So we can check if it's working as expected and then merge it.

mgruner:
You are right, @dominikklein.

mgruner:
Applied in 2ebac4a4d2ba1c5a76a0509c20f579fc319b5e38 with a small syntax fix. Thanks a lot @albanobattistella and sorry for the delay!

mgruner:
This will be included in Zammad 7.0.

PR #5753, opened 2025-08-27
Basic development docker compose setup

r3cha:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,5 @@ schema.graphql
 
 # Packages
 *.zpm
+
+.pnpm-store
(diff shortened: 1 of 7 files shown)


dominikklein:
@r3cha Thanks for the nice contribution idea. In the last weeks, we already added some dev container support:

https://github.com/zammad/zammad/tree/develop/.devcontainer

It would be nice if you could check if this is working for you or if you see any problems. We are also always open to improvements here, because this is something new that we added, and we are currently building some knowledge.

We need to add this part in our current development documentation; this is currently outstanding.

PR #5795, opened 2025-09-23
Fix misspelling in ticket_update in swedish

JonasMandren:
Fixed a misspelling of the word "Updated" in swedish.


dominikklein:
@JonasMandren Thanks for the contribution.

Translations of Zammad are processed centrally in our [Weblate instance](https://translations.zammad.org/). This allows non-developers to review and update translation texts of the language they actually speak.

PR #5801, opened 2025-10-01
Update shares and approvals

tamerWagih:


mgruner:
@tamerWagih we would like to understand more about your use case.

For now, I'll close this PR, as it is quite large and without any context.

PR #5842, opened 2025-11-08
Rename "contains" operators (e.g. Tags)  #4916

### LINKED ISSUE #4916: Rename "contains" operators (e.g. Tags)
### Used Zammad Version

6.1

### Environment

- Installation method: [e.g. source, package]
- Operating system (if you're unsure: `cat /etc/os-release` ): [e.g. debian 10.4, ubuntu 20.04]
- Database + version:  [e.g. postgresql 9.3, mysql 5.7, mariadb 10.3]
- Elasticsearch version: [e.g. 7.17]
- Browser + version: [e.g. chrome 83, safari 14, firefox 105]

### Actual behaviour

Many people have different understandings of the **contains all not** and the **contains one not** operator.

### Expected behaviour

Based on the discussions it came out that it might make sense to rename these operators.

We need new namings for:
- contains one
- contains all
- contains one not
- contains all not

-

### Support Ticket

_No response_

yes

AbdallahSalah003:
### Summary

This PR updates all occurrences of the following operators to improve clarity and consistency based on community feedback in issue #4916:

| Old Name           | New Name        |
|--------------------|-----------------|
| contains one       | includes any    |
| contains all       | includes all    |
| contains one not   | excludes any    |
| contains all not   | excludes all    |

### Changes

- Updated all backend, frontend, and specs reference to the renamed operators 
- Regenerated the translation catalog (`i18n/zammad.pot`) using:
  ```bash
  rails generate zammad:translation_catalog

### Related Issue

Fixes #4916

### DIFF
--- a/app/assets/javascripts/app/controllers/_ui_element/_application_selector.coffee
+++ b/app/assets/javascripts/app/controllers/_ui_element/_application_selector.coffee
@@ -29,14 +29,14 @@ class App.UiElement.ApplicationSelector
       'integer$': [__('is'), __('is not'), __('is less than'), __('is less than or equal to'), __('is greater than'), __('is greater than or equal to')]
       '^radio$': [__('is'), __('is not')]
       '^select$': [__('is'), __('is not')]
-      '^multiselect$': [__('contains all'), __('contains one'), __('contains all not'), __('contains one not')]
+      '^multiselect$': [__('includes all'), __('includes any'), __('excludes all'), __('excludes any')]
       '^tree_select$': [__('is'), __('is not')]
-      '^multi_tree_select$': [__('contains all'), __('contains one'), __('contains all not'), __('contains one not')]
+      '^multi_tree_select$': [__('includes all'), __('includes any'), __('excludes all'), __('excludes any')]
       '^autocompletion_ajax_external_data_source$': [__('is'), __('is not')]
       '^input$': [__('contains'), __('contains not'), __('is any of'), __('is none of'), __('starts with one of'), __('ends with one of')]
       '^richtext$': [__('contains'), __('contains not')]
       '^textarea$': [__('contains'), __('contains not')]
-      '^tag$': [__('contains all'), __('contains one'), __('contains all not'), __('contains one not')]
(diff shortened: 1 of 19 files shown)


dominikklein:
@rolfschmidt Do you remember the last idea about this topic?

dominikklein:
@dvuckovic Any thoughts? :-D

fliebe92:
I like the improved wording. :>

dvuckovic:
@dominikklein It is the most voted proposal in [the linked issue](https://github.com/zammad/zammad/issues/4916), after all. We should just confirm the consensus with all stakeholders, before we proceed.

However, I currently don't see any migration for existing systems: if there are any records referring to the "old" operators in their conditions, how would they ever know about the "new" ones? I'm afraid they will just stop functioning.

```rb
> Trigger.last.condition
=> {"ticket.tags" => {"operator" => "contains one", "value" => "tag1, tag2, tag3"}}
```

fliebe92:
True, migration is missing, but required.

martini:
I like the approach. But keep in mind, that this is an heavy breaking change in the API too. I would love to see some backward compatibility. @YetAnotherGerrit 

Regarding migration: I'm sure that users may have invalid saved conditions, so migration may fail. Keep in mind, wrong conditions (maybe because processes which expect still the older or new style of conditions) will lead to data loos.

dvuckovic:
Yes, one option could be to stay pragmatic, and just rename user-facing labels, but not the actual structure that gets saved/parsed.

Another approach could be a compatibility layer, that maps old operator names to new ones, and possibly seamless migration of existing entries on any update (that doesn't fail).

mantas:
I'd vote to change user-facing labels only. They don't map 1:1 to underlying commands in other languages than English anyway. And makes it easier if we want to fine-tune them later on.

YetAnotherGerrit:
Yes, for the reasons @martini described, I agree with Mantas.

dominikklein:
@AbdallahSalah003 do you want to try only changing the visible labels?

Issue #5893, opened 2025-12-17
Waiting for Customer + Checklist enabled causes JS crash


justnx:
### Used Zammad Version

6.5.2

### Environment

- Installation method: package
- Operating system Debian 12 (bookworm)
- Database + version:  Postgresql 15
- Elasticsearch version: 7.17.29
- Browser + version: chrome latest, Firefox latest

### Actual behaviour

When ticket has custom state "Waiting for Customer" and Checklist feature is enabled, changing to that state the ticket UI does not work and throwing a frontend Javascript error and the ticket state is not applied and ticket can't be safed anymore. It seems its loosing last state after processing custom state chain.

### Expected behaviour

Ticket state should update normally.

### Steps to reproduce the behaviour

1. Enable the checklist feature in Zammad.
2. Create a ticket and change custom state to "Wating for Customer"
3. Open the ticket and try to change its state via the drowndown or save.

JS error in browser console:

Stack trace:
```
Uncaught TypeError: can't access property "state_type", App.TicketState.find(...) is null
    submitChecklist https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:71
    e https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:70
    submit https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:71
    e https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:70
    s https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:9
    dispatch https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:2
    handle https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:2
    add https://censored-domain.de/assets/application-2c1fb08e2b971a2a55ab4bf3705d9ec291b276b1a325aa0e7569aa4c6b801682.js:2
[...]

PR #5894, opened 2025-12-17
Fix: Waiting for Customer + Checklist enabled causes JS crash

justnx:
Fixes Waiting for Customer + Checklist enabled causes JS crash #5893 
Also see: https://community.zammad.org/t/custom-state-not-working/16277/10

### DIFF
--- a/app/assets/javascripts/app/controllers/_ai/provider.coffee
+++ b/app/assets/javascripts/app/controllers/_ai/provider.coffee
@@ -27,10 +27,10 @@ class AiProviderSettings extends App.Controller
   constructor: ->
     super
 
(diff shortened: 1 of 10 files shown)


mantas:
Hey @justnx 

Thank you for looking into this!

I wonder if existential operators in `ticket?.stateType?.name` are needed. Looks like simply adding a check for `next_state_id` would be enough to solve the issue? It'd be great to keep PR to the smallest possible footprint :) Thanks!

justnx:
Hi @mantas 

sure thing, there yah go.

P.S. I just tried to add extra safety the vue style ;-)

PR #5929, opened 2026-01-26
Maintenance: Ensure devcontainer ports are opened locally only

MrGeneration:
Disclaimer up front:
Because of the way I work with vscode, I cannot use the internal approach for devcontainers and thus rely on `devcontainer up ...` commands. So I am not entirely sure if this is consistent with those who have a clean workspace and thus not the "problems" I have.

---

This PR ensures that docker opens all ports for localhost connections only.
The current as is opens them to `0.0.0.0` and thus makes the relevant ports available to anyone in the same network.

This thus might be security relevant, if a developer e.g. tests stuff locally with more sensitive information (e.g. bug reproduction).

I'd be interested to learn, if the devcontainer with the reopen in container option does open port `3000` on localhost. In my environment I am forced to explicitely enable the port forwarding (and have to use `ZAMMAD_BIND_IP` environments to be able to talk to the dev server at all. Again, my setup is... different).


mgruner:
@dvuckovic something for you?

dvuckovic:
@MrGeneration Merged in 652977c489363119c8ee671cbddebd94e7850395, thanks for your contribution!

dvuckovic (approved):

dvuckovic:
> I'd be interested to learn, if the devcontainer with the reopen in container option does open port 3000 on localhost. In my environment I am forced to explicitely enable the port forwarding (and have to use ZAMMAD_BIND_IP environments to be able to talk to the dev server at all. Again, my setup is... different).

To answer your question: (in VS Code, at least) reopening folder in container does *not* expose services running in the devcontainer  to the outside network; they are available exclusively on `localhost`, not the wider network. I believe this is by design.

So, if someone actually wants to access services running in the devcontainer (e.g. puma) from outside their dev-box (for example, to test the mobile view on a phone), they would need to explicitly forward the service somehow. I had luck using SSH tunnels for this, since I also want almost always to have a proper SSL termination exposed to any target devices. Again, this is highly specific, and not something we provide out-of-box due to the security implications you outlined, and sheer number of different setups out there.

With your changes, we are just aligning the behavior of other devcontainer *stack* services with the actual devcontainer.

MrGeneration:
Thanks for your insight.

> With your changes, we are just aligning the behavior of other devcontainer stack services with the actual devcontainer.

So basically adding the ENV with `ZAMMAD_BIND_IP: 0.0.0.0` would be the last thing needed to align devcontainer environments that are not started via VS Code directly? If so, I'll happily provide another alignment. If that's feasible of course. :)

dvuckovic:
I'm not sure, to be honest. If you have a particular idea, feel free to propose it.

I tried setting this env var via `export ZAMMAD_BIND_IP=0.0.0.0`, but it doesn't seem to help exposing the services running in devcontainer to the outside.

Perhaps the reason is that the devcontainer uses Docker network interface only:

```console
$ ifconfig
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.18.0.8  netmask 255.255.0.0  broadcast 172.18.255.255
```

I would imagine it will require some more changes to the stack to actually work. And it may depend on the platform used 🤷 

But I very much prefer the other, explicit way of tunneling the service to the outside. I think it's safer, and will result in a more usable DX.

MrGeneration:
Thanks for your insights. I believe I explained it poorly on the ENV regard.

In my devcontainer setup (outside of VS Code execution), the lack of the ENV will cause me not being able to access the dev-server from *outside* of the devcontainer.

While cycling through the compose on my end your end, I noticed that the default compose doesn't expose any ruby related ports... my addon compose has this though:

```
services:
  devcontainer:
    environment:
      ZAMMAD_BIND_IP: '0.0.0.0'
    volumes:
      - ../../..:/workspaces:cached
    ports:
      - "127.0.0.1:3000:3000"
      - "127.0.0.1:3001:3001"
      - "127.0.0.1:3036:3036"
      - "127.0.0.1:6042:6042"

```

I will need to re-visit this topic in the explicit repository scope (so vanilla vanilla approach).
Should the stack not be available outside of VS Code scope (but with devcontainer), shall I add a PR if that's the case?

If that's outside of your scope, that's fine!

dvuckovic:
At the moment, I can se the port forwarding that goes into https://github.com/zammad/zammad/blob/ed6ff4b3e42e744360e7062f3b6fa51d54d41e99/.devcontainer/default/devcontainer.json#L23 definition. However, we did notice services are also automatically forwarded when started in the devcontainer itself. And in case of conflicts, a random local port is assigned.

If you are still using the devcontainer stack in an unconventional way, we can evaluate your changes down the line.

Issue #5886, opened 2025-12-11
Translation to Catalan uses wrong locale code


mgruner:
### Used Zammad Version

6.5

### Environment

- irrelevant

### Actual behaviour

Zammad has a locale named "Català", which uses the non-existent locale identifier `es-ca`. This shows up in Weblate as "Spanish (es-ca)", because Weblate cannot identify the target language: https://translations.zammad.org/projects/zammad/zammad-development/es_CA/

### Expected behaviour

Catalan is not a variant of Spanish, it's a separate language: https://en.wikipedia.org/wiki/Catalan_language. This confuses translators which want to contribute to it, since they don't find the language on Weblate.

### Steps to reproduce the behaviour

- Select the Catalan locale
- check the saved code, or the entry in locales.yml:

```yml
- locale: es-ca
  alias: ''
  name: Català
  active: true
  dir: ltr
```

### Support Ticket

Ticket#10207339

### I'm sure this is a bug and no feature request or a general question.

no

mgruner:
### Technical Recommendation

- We can't simply change the locale code in place: user profiles would have to be updated, knowledge bases would have to be updated and this would cause URL changes in the public knowledge bases. 
- Therefore we suggest to create a new `ca` Locale, deprecate `es-ca` via Release Notes, and remove it after one or two versions.

MrGeneration:
Sorry but why can't a db migration update this exactly?
Burding this on users (especially customers) is a very bad user experience, because it can be needed very delayed to the actual update where this change becomes necessary.

Especially if developers know at which parts this needs to be changed...? It's hard to understand why this can't be automated (so what the downsides of this action would be). I guess even rails commands might help for admins if db migrations are no option... but manual user interaction?

mgruner:
👋 the main reason against this is that there might be Catalan KBs out there where the URLs would change all of a sudden. We believe this is a no-go.

mgruner:
We can still provide a migration for user profiles when the old locale gets removed or so.

MrGeneration:
> 👋 the main reason against this is that there might be Catalan KBs out there where the URLs would change all of a sudden. We believe this is a no-go.

That's a very fair point - couldn't there be a permanent redirect on application servers end? Just as an idea to solve it. Maybe it's an option as suggestion for the webserver configuration to reduce code maintenance.

PR #5945, opened 2026-02-08
Follow up #5886 - migrate user preferences to new Catalan locale code

MrGeneration:
I know it was decided against it, but it bugged me so much, that I did a little research.
[Back in 2022](https://github.com/zammad/zammad/blob/18206b495c3e97f21456d3701c7c248a95c875fd/db/migrate/20220308072741_rename_tw_locale.rb#L4), this kind of migration has happened.

This PR will touch affected user preferences only and should not have a negative impact on the knowledge base (at least I personally don't see one, but I might miss a vector).

I believe that this will improve the users experience and takes off burdens from administrators of an instance.

### DIFF
--- a/db/migrate/20260209120000_migrate_catalan_user_preference.rb
+++ b/db/migrate/20260209120000_migrate_catalan_user_preference.rb
@@ -1,0 +1,14 @@
+# Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
+
+class MigrateCatalanUserPreference < ActiveRecord::Migration[8.0]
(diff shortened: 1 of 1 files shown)


dominikklein:
@mgruner @mantas Maybe something you could check.

mgruner:
@MrGeneration thanks a lot! Looks good to me.

@mantas can you handle this, please, with a backport to pre-release?

mantas:
Hey @MrGeneration 

Nice to see your contribution! It's now merged with above commit.

PR #5952, opened 2026-02-16
Maintenance: Fix typos

urbalazs:
This PR fixes some typos and standardizes wording.

Please check, if the strings in `display` key should be translatable or not. The strings `ticket.agent` and `ticket.customer` looks different in translator point of view.

See in  `db/seeds/settings.rb`:

```
  description: __('Defines the session timeout for inactivity of users. Based on the assigned permissions the highest timeout value will be used, otherwise the default.'),
  options:     {
    form: [
      {
        display:   __('Default'),
        null:      false,
        name:      'default',
        tag:       'select',
        options:   options,
        translate: true,
      },
      {
        display:   __('admin'),
        null:      false,
        name:      'admin',
        tag:       'select',
        options:   options,
        translate: true,
      },
      {
        display:   __('ticket.agent'),
        null:      false,
        name:      'ticket.agent',
        tag:       'select',
        options:   options,
        translate: true,
      },
      {
        display:   __('ticket.customer'),
        null:      false,
        name:      'ticket.customer',
        tag:       'select',
        options:   options,
        translate: true,
      },
    ],
  },
```

### DIFF
--- a/db/seeds/schedulers.rb
+++ b/db/seeds/schedulers.rb
@@ -256,7 +256,7 @@ Scheduler.create_if_not_exists(
   created_by_id: 1,
 )
 Scheduler.create_if_not_exists(
-  name:          __('Update exchange oauth 2 token.'),
+  name:          __('Update exchange OAuth2 token.'),
   method:        'ExternalCredential::Exchange.refresh_token',
   period:        10.minutes,
   prio:          1,
--- a/db/seeds/settings.rb
+++ b/db/seeds/settings.rb
@@ -307,18 +307,18 @@ Setting.create_if_not_exists(
   frontend:    true
 )
 Setting.create_if_not_exists(
-  title:       __('Websocket backend'),
+  title:       __('WebSocket backend'),
   name:        'websocket_backend',
   area:        'System::WebSocket',
-  description: __('Defines how to reach websocket server. "websocket" is default on production, "websocketPort" is for CI'),
+  description: __('Defines how to reach WebSocket server. "websocket" is default on production, "websocketPort" is for CI.'),
   state:       Rails.env.production? ? 'websocket' : 'websocketPort',
   frontend:    true
 )
 Setting.create_if_not_exists(
-  title:       __('Websocket port'),
+  title:       __('WebSocket port'),
   name:        'websocket_port',
   area:        'System::WebSocket',
-  description: __('Defines the port of the websocket server.'),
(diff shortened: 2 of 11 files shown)


mantas:
Looks interesting. One point regarding changes in seed files. An accompanying migration (see db/migrate/...) is needed to update existing systems :)

mgruner:
@dvuckovic @mantas can one of you please take this further? Balazs is from the Hungarian team translating Zammad. We cannot expect them to perform more technical development tasks.

db/seeds/settings.rb
```diff
@@ -6017,7 +6017,7 @@
 )
 
 Setting.create_if_not_exists(
-  title:       __('UI Desktop Beta Switch'),
+  title:       __('UI Desktop BETA Switch'),
```
dvuckovic, line 6020: We would reword this as `Desktop BETA UI Switch`, it reads better, and aligns more with other strings.

db/seeds/settings.rb
```diff
@@ -6038,7 +6038,7 @@
   title:       __('UI Desktop BETA Switch Roles'),
```
dvuckovic, line 6038: `Desktop BETA UI Switch Roles`

db/seeds/settings.rb
dvuckovic, line 6029: `Desktop BETA UI Switch Admin Menu`

dvuckovic (approved):
@urbalazs Thanks for your efforts!

We will further improve some strings, and add the missing migration @mantas mentioned, so existing systems can be aligned to new wording as well.

As for the `session_timeout` setting: the current idea is to align display strings with their actual, translatable permission *labels* (i.e. `Admin interface` for `admin`, `Agent tickets` for `ticket.agent`, `Customer tickets` for `ticket.customer`), pending an internal review. The technical permission names will be hidden under a help icon, right next to the label, but they shall remain untranslatable.

There is nothing needed from your side at this point, stay tuned!

db/seeds/schedulers.rb
```diff
@@ -256,7 +256,7 @@
   created_by_id: 1,
 )
 Scheduler.create_if_not_exists(
-  name:          __('Update exchange oauth 2 token.'),
+  name:          __('Update exchange OAuth2 token.'),
```
dvuckovic, line 259: `Update Exchange OAuth2 token.`

dvuckovic:
@urbalazs Merged, thanks again for your contribution!

PR #5956, opened 2026-02-18
Fixes #5955 - Inline images from quoted emails cause invalid base64 GraphQL error.

### LINKED ISSUE #5955: Inline images from quoted emails cause invalid base64 GraphQL error in desktop view
### Used Zammad Version

7.1.x (develop branch, current HEAD)

### Environment

- Installation method: Docker
- Operating system: Debian Trixie (Docker base image ruby:3.4.8-slim-trixie)
- Database + version: PostgreSQL 16.1
- Elasticsearch version: 8.17.0
- Browser + version: Chrome 133, Firefox 135

### Actual behaviour

When replying to or forwarding a ticket email that contains inline images (embedded screenshots, company logos, etc.), a red error flash appears:

> Variable $files of type [UploadFileInput!]! was provided invalid value for 0.content (invalid base64)

The error is transient — the reply still sends, but the UX is jarring and confusing for agents.

### Expected behaviour

No error. Inline images in quoted content should be recognized as already-persisted server resources and not re-uploaded.

1. Receive an email ticket that contains inline/embedded images (e.g. a screenshot pasted into Outlook, or a company logo in the email signature)
2. Open the ticket in desktop-view (`/desktop`)
3. Click "Reply" or "Forward"
4. The quoted content includes the inline images — a red error flash appears immediately

### Root cause

In `app/frontend/shared/components/Form/fields/FieldEditor/features/image-handler/ImageHandler.vue`, the `handleUpload()` guard and `uploadCacheExists` computed only match `/api/v1/attachments/`. However, inline images from quoted emails use `/api/v1/ticket_attachment/:ticket_id/:article_id/:id?view=inline` URLs. These fall through to the else branch, which sends the raw URL path as base64 `content` to the GraphQL mutation. The backend's `BinaryStringType` then fails on `Base64.strict_decode64`.

### Fix

Broaden the guard from `/api/v1/attachments/` to `/api/v1/` — any image served by the Zammad API is already persisted server-side and doesn't need client-side re-uploading. A PR with the fix will follow.

Deenyoro:
## Summary

- Fixes a transient red error flash (`Variable $files of type [UploadFileInput!]! was provided invalid value for 0.content (invalid base64)`) when replying to or forwarding an email with inline images (embedded screenshots, company logos, etc.)
- Single file change, 2 lines modified

## Root Cause

When an email with inline images is quoted in a reply/forward, the backend replaces `cid:` references with `/api/v1/ticket_attachment/:ticket_id/:article_id/:id?view=inline` URLs. The `ImageHandler.vue` component's `handleUpload()` guard and `uploadCacheExists` computed only matched `/api/v1/attachments/`, so `ticket_attachment` URLs fell through to the else branch which sent the raw URL path as base64 `content` to the GraphQL mutation. `BinaryStringType` then failed on `Base64.strict_decode64`.

## Fix

Broaden the guard from `/api/v1/attachments/` to `/api/v1/` — any image served by the Zammad API is already persisted server-side and doesn't need client-side re-uploading.

## Testing

Built a clean Docker image from upstream develop with only this patch applied, deployed to a full stack (PostgreSQL 16.1, Elasticsearch 8.17.0, Redis 7, Memcached 1.6, Rails, Nginx) and verified:

- [x] Fix compiled into production Vite bundle (FieldEditorInput chunk) — confirmed `/api/v1/` guard in minified JS
- [x] Reply to email ticket with inline images — no error flash
- [x] Forward email ticket with inline images — no error flash  
- [x] New image uploads via editor toolbar — still works normally
- [x] `uploadCacheExists` computed correctly identifies server-persisted images (no dimming)
- [x] No regression in desktop-view or mobile app asset serving
- [x] Audited all `/api/v1/` routes — none serve user-uploadable content that would need re-uploading

### DIFF
--- a/app/frontend/shared/components/Form/fields/FieldEditor/features/image-handler/ImageHandler.vue
+++ b/app/frontend/shared/components/Form/fields/FieldEditor/features/image-handler/ImageHandler.vue
@@ -27,7 +27,7 @@ const isResizing = ref(false)
 const imageLoaded = ref(false)
 const failedDueToCors = ref(false)
 const isDraggable = computed(() => props.node.attrs.isDraggable)
-const uploadCacheExists = computed(() => props.node.attrs.src.startsWith('/api/v1/attachments/'))
+const uploadCacheExists = computed(() => props.node.attrs.src.startsWith('/api/v1/'))
 const uploadInsideOtherEditor = computed(
   () => props.node.attrs.src.startsWith('blob:') && !props.node.attrs.content,
 )
@@ -42,7 +42,7 @@ const getNodeStyle = () =>
   }, {})
 
 const handleUpload = () => {
-  if (props.node.attrs.src.startsWith('/api/v1/attachments/')) return
+  if (props.node.attrs.src.startsWith('/api/v1/')) return
   if (props.node.attrs.src.startsWith('blob:') && !props.node.attrs.content) return
 
   if (props.node.attrs.src.startsWith('file:')) {


mantas:
Looks pretty good. But I think it may be better to keep scope limited to `api/v1/attachments` + `api/v1/ticket_attachments`.

dominikklein:
Yes agree, we should not be that generic.

But we also need to check what happens on save. Normally, we should not have any references to other articles.

So maybe we need a different approach/solution for this situation.

fliebe92:
Hi @Deenyoro. Thanks for the pull request! However, after some internal discussion, we noticed that this is not a proper way to fix it.

We'll try to address it soon. Stay tuned!

PR #5962, opened 2026-02-21
Skip redis and database connections during asset precompilation

cryptomilk:
Asset precompilation only needs to compile CSS, JavaScript, and images. It does not require Redis (for ActionCable) or PostgreSQL/MySQL database connections. This patch modifies initializers to skip these services when running the assets:precompile rake task, allowing the build to succeed in environments where these services are not available.

I'm currently packaging zammad for openSUSE [here](https://build.opensuse.org/package/show/home:gladiac:apps/zammad). This is one of my patches trying to get it built.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -22,6 +22,13 @@ module Zammad
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 8.0
 
+    # Check if the current process is running assets:precompile only
+    def self.assets_precompile?
+      defined?(Rake) &&
+        Rake.respond_to?(:application) &&
+        Rake.application.top_level_tasks.any? { |t| t.include?('assets:precompile') }
+    end
+
     Rails.autoloaders.each do |autoloader|
       autoloader.ignore            "#{config.root}/app/frontend"
       autoloader.do_not_eager_load "#{config.root}/lib/core_ext"
--- a/config/initializers/db_preflight_check.rb
+++ b/config/initializers/db_preflight_check.rb
@@ -1,5 +1,8 @@
 # Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
+# Skip database preflight check during asset precompilation (no DB needed)
+return if Zammad::Application.assets_precompile?
+
 # Rails' constant auto-loading resolves to 'rails/initializable' instead
 require 'zammad/application/initializer/db_preflight_check'
 
--- a/config/initializers/zzz_action_cable_preferences.rb
+++ b/config/initializers/zzz_action_cable_preferences.rb
@@ -1,5 +1,8 @@
 # Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
+# Skip ActionCable configuration during asset precompilation (no Redis needed)
+return if Zammad::Application.assets_precompile?
+
 require_relative '../../lib/zammad/service/redis'
 
 Rails.application.config.action_cable.cable = {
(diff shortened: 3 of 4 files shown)


MrGeneration:
Just out of interest, why are you building suse packages when there's already some from the vendor?

cryptomilk:
I prefer to use distribution packages instead of vendoring.

config/initializers/db_preferences.rb
```diff
@@ -1,5 +1,8 @@
 # Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
+# Skip database preferences during asset precompilation (no DB needed)
```
mgruner: Why is this needed here? It just configures the DB, should not connect to it, right?

mgruner:
This is a very interesting approach. We already had a similar use case in `config/pre_initializers/load_asset_gems.rb`, where we used a custom approach freezing the CLI arguments so that we can detect if assets:precompile is running. (This ArgvHelper is also used for other purposes in the task handling).

Your solution seems to be more elegant. Maybe it would be good to consider centralizing this logic, e.g. in `Zammad::Application.is_assets_precompile_call?` or similar? 

@mantas FYI

config/initializers/db_preferences.rb
cryptomilk: Fixed.

cryptomilk:
> Your solution seems to be more elegant. Maybe it would be good to consider centralizing this logic, e.g. in `Zammad::Application.is_assets_precompile_call?` or similar?

Done.

config/application.rb
```diff
@@ -22,6 +22,11 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 8.0
 
+    # Check if the current process is running assets:precompile
+    def self.assets_precompile?
+      defined?(Rake) && Rake.application.top_level_tasks.any? { |t| t.include?('assets:precompile') }
```
mantas: I wonder what would happen if someone gives multiple commands, one of them is `assets:precompile`, but another one does require DB connection.

Maybe it'd be better check if `assets:precompile` is the only given command?

Also, it may better to name this check in a more generic way. Then it could be easily extended in the future if there're more rake tasks that can turn this check off.

cryptomilk: Done.

cryptomilk:
Ping ...

fliebe92:
@mantas could you try to finish this one today/tomorrow, please?

mgruner:
@fliebe92 we might need a follow-up here to simplify the packager.io builds, if they really don't require postgresql any more. Currently they provide a service container for this.

mgruner:
We should consider if this somehow conflicts with the existing ZAMMAD_SAFE_MODE handling in some way.

config/initializers/zzz_action_cable_preferences.rb
```diff
@@ -1,5 +1,8 @@
 # Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
+# Skip ActionCable configuration during asset precompilation (no Redis needed)
+return if Zammad::Application.assets_precompile?
+
```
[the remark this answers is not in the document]
cryptomilk: I disagree :-)

config/initializers/db_preflight_check.rb
```diff
@@ -1,5 +1,8 @@
 # Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
+# Skip database preflight check during asset precompilation (no DB needed)
+return if Zammad::Application.assets_precompile?
+
```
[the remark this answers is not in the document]
cryptomilk: I disagree :-)

mantas:
Hi @cryptomilk 

I just checked and it looks like existing SafeMode handles assets precompilation too. Following command run just fine with Postgres, Redis and Memached turned off.

```
ZAMMAD_SAFE_MODE=1 RAILS_ENV=production bundle exec rake assets:precompile
```

Closing this PR for now.

PR #5963, opened 2026-02-24
Fix Rack compatibility for newer Rack versions

cryptomilk:
Rack::Utils.add_cookie_to_header has been deprecated and removed in version 3.10.

This patch makes the monkey-patch conditional - it only applies if the method exists, allowing compatibility with both Rack 2.x and 3.x+.

### DIFF
--- a/lib/core_ext/rack/utils.rb
+++ b/lib/core_ext/rack/utils.rb
@@ -7,16 +7,23 @@ module Rack
 
     module_function
 
-    singleton_class.alias_method :original_add_cookie_to_header, :add_cookie_to_header
+    if Rack::Utils.respond_to?(:add_cookie_to_header)
+      # Rack 2.x
+      singleton_class.alias_method :original_add_cookie_to_header, :add_cookie_to_header
 
-    # https://github.com/rack/rack/blob/2.2.3/lib/rack/session/utils.rb#L223-L262
-    def add_cookie_to_header(header, key, value)
-
-      if value.is_a?(Hash)
-        value[:secure] = ::Session.secure_flag?
+      # https://github.com/rack/rack/blob/2.2.3/lib/rack/session/utils.rb#L223-L262
+      def add_cookie_to_header(header, key, value)
+        value[:secure] = ::Session.secure_flag? if value.is_a?(Hash)
+        original_add_cookie_to_header(header, key, value)
       end
+    else
+      # Rack 3.1+ — add_cookie_to_header was removed; set_cookie_header! is now used
+      singleton_class.alias_method :original_set_cookie_header!, :set_cookie_header!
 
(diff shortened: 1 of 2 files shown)


mantas:
I wonder if we still need this monkey patch in the first place. Looks like it was added during migration to Rails 6. But maybe latest Rails no longer relies on it...

mantas:
After further testing, it looks like our code is still using this monkeypatch. At the same time, our dependencies block Rack 3 usage, so it should be not a problem either.

For now I think we should keep the code as-is. Once our dependencies make us use Rack 3, we can rethink our approach here.

@mgruner what do you think?

mgruner:
@mantas ok. How do we track this?

mantas:
Good question.

I see Sprockets is one of the gems requiring Rack 2. I've a feeling we may be stuck with it till we retire the old UI. Not sure if it's worth to track something that far in the future.

On top of that, once we eventually upgrade Rack, this monkeypatch will simply prevent Zammad from working. I'm pretty sure we won't forget this.

cryptomilk:
What's the issue with updating sprockets? I'm not fully there yet. But I have a patch to uses sprockets 4.x.

mantas:
We're working on a new UI that uses modern bundling tools. Not sure if there's any will to update the old but trusty build chain for the old UI. Which will be gone sooner or later anyway.

mgruner:
I would be interested in using sprockets 4. We don't know when exactly we'll be able to drop it, so if it's no major effort it would be hygiene maintenance.

mgruner:
Maybe in a separate PR?

cryptomilk:
I have zammad running now. It has Rack 3.x and Sprockets 4.x

cryptomilk:
Ping

fliebe92:
Hi @cryptomilk. Trying to discuss it today how we can proceed there. Sorry the inconvenience.

mantas:
Hey @cryptomilk . Would you mind to provide a PR (or extend this one) which contains both Rack update and this fix in a single commit?

Also, the fix should set `secure` flag in newer Rack versions too. We're still relying on it. Unfortunately Rails' `force_ssl` is not an option for us due to the nature of packaged software running in many different environments.

cryptomilk:
You mean that I should add this commit to https://github.com/zammad/zammad/pull/5967 ?

mantas:
But it looks like that PR does not include Rack update?

Ideally, we'd like 2 separate and self-contained PRs:

- Rack update. Including a bump of that monkeypatch to work with the newer Rack version.

- Sprockets update. Which needs more internal testing wether our build pipeline can handle the newer version without too much work.

mantas:
This was addressed in another commit. Thank you for bringing our attention to this matter!

PR #5966, opened 2026-02-25
Make aws-sdk-s3 and S3 storage provider optional

cryptomilk:
Guard require 'aws-sdk-s3' so S3 provider file loads cleanly when the gem is not installed. It is marked as optional in the Gemfile.

### DIFF
--- a/app/models/store/provider/s3.rb
+++ b/app/models/store/provider/s3.rb
@@ -1,6 +1,10 @@
 # Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
-require 'aws-sdk-s3'
+begin
+  require 'aws-sdk-s3'
+rescue LoadError
+  # Gem not installed, S3 storage provider will not be available
+end
 
 module Store::Provider::S3
 


config/initializers/store_provider_check.rb
```diff
@@ -5,6 +5,7 @@
   next if !Rails.env.production?
 
   storage_provider = Setting.get('storage_provider')
+  next if storage_provider.blank?
```
fliebe92: How is it possible that `storage_provider` is set to empty? It's a dropdown, and you have to select a value. There is also a default set.

cryptomilk: I think this happened when zammad was started before db:seed was run. Do you prefer to remove it?

fliebe92: I was just wondering about this added line, because it should not happen in real life 😁

Would be fine for me to remove it. ~~About the loading of the related gem, totally fine, I'll try to wrap this up today.~~

app/models/store/provider/s3.rb
```diff
@@ -1,6 +1,10 @@
 # Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
-require 'aws-sdk-s3'
+begin
+  require 'aws-sdk-s3'
+rescue LoadError
+  # Gem not installed, S3 storage provider will not be available
+end
```
fliebe92, line 7: After thinking more about it, how is this even happening?

Could you describe what happened in your setup, @cryptomilk? Thanks!

fliebe92: `require: false` in the `Gemfile` just controls Bundlers behavior of auto-requiring the gem, not if it is installed or not. There's no `group` and no `optional: true`, so, it should be installed all the time, it's just not used.

cryptomilk: I'm packaging zammad for openSUSE and have removed aws-sdk-s3 from the Gemfile.

https://build.opensuse.org/package/show/home:gladiac:apps/zammad

cryptomilk: Should I add an s3 group?

fliebe92: We don't need a group, no. I guess I'll try it with your current approach. I was just wondering about how to even face the situation.

fliebe92:
Hi @cryptomilk and thanks again for your time invested into it. After some internal discussion, we noticed that the way the S3 backend in Zammad is requiring the needed gem is not the way we are usual using.

Based on this, we improved the S3 backend with https://github.com/zammad/zammad/commit/310cefb4ad580bf3a57134ddf8557056f6c99083. I'll try to backport it to `stable` branch as well, stay tuned.

Closing for now. 😄

fliebe92:
Merged to `stable` as well: https://github.com/zammad/zammad/commit/ee2b8445b7f5983ed3ffd4218077db01eec6ff22

PR #5969, opened 2026-02-26
Fixes #5968 - Increase OTRS import HTTP timeouts

### LINKED ISSUE #5968: OTRS import HTTP timed out
### Used Zammad Version

6.5.2

### Environment

- Installation method: Docker
- Operating system (if you're unsure: `cat /etc/os-release` ): n/a
- Database + version:  postgresql 17.7
- Elasticsearch version: 8.19.11
- Browser + version: chrome 145.0.7632.116 arm64 macos

### Actual behaviour

On initiating a migration from web ui importing from OTRS running on the same docker host i got a timeout in `lib/import/otrs/requester.rb` causing the import to fail.

After bumping the default timeouts to:
`
      read_timeout: 600,
      total_timeout: 1200,
`

import went through without an issue. OTRS and Zammad were running on the same host machine. Machine is a bit older E5-2609 v3 @ 1.90GHz. It might be a good idea to set this higher.

### Expected behaviour

Import works without issue.

Possibly importing from a slower OTRS ressource.

### Support Ticket

_No response_

yes

cygnusb:
## Summary

- Increases `read_timeout` from 120s to 600s in `lib/import/otrs/requester.rb`
- Increases `total_timeout` from 360s to 1200s
- Fixes OTRS migration failures on slower hardware or with larger datasets

Closes #5968

## Test plan

- [x] Verify OTRS import completes successfully on slower systems
- [x] Verify no regression for fast systems (timeouts are upper bounds, not delays)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/import/otrs/requester.rb
+++ b/lib/import/otrs/requester.rb
@@ -119,8 +119,8 @@ module Import
           params,
           {
             open_timeout:  10,
-            read_timeout:  120,
-            total_timeout: 360,
+            read_timeout:  600,
+            total_timeout: 1200,
             user:          Setting.get('import_otrs_user'),
             password:      Setting.get('import_otrs_password'),
           },


dominikklein:
Thanks for the contribution :-) We will handle this with an addition, so I will close this PR for now, but it will be available soon.

PR #6007, opened 2026-03-12
Sort expert header list alphabetically and add Exchange auto-reply headers

HanSolo72:
…aders

The list of expert headers used in postmaster filters is currently defined as a static array and displayed in the same order in which it appears in the source code. As the list grows, locating a specific header becomes increasingly difficult.

This change introduces an alphabetical sorting of the expert header options before they are returned from @defaults. Sorting is performed by the visible header name to improve usability in the UI.

In addition, two Microsoft Exchange specific headers have been added:

X-Auto-Response-Suppress

X-MS-Exchange-Inbox-Rules-Loop

These headers are commonly used by Microsoft Exchange / Microsoft 365 to mark automatically generated responses such as out-of-office replies or inbox rule loops. Having them available in the expert header list allows administrators to create more reliable postmaster filters to detect and handle automated replies.

The changes include:

Alphabetical sorting of groups.expert.options

Addition of the headers:
* X-Auto-Response-Suppress
* X-MS-Exchange-Inbox-Rules-Loop

This improves both usability of the expert header selector and the ability to detect automatic email responses in Exchange environments.

### DIFF
--- a/app/assets/javascripts/app/controllers/_ui_element/postmaster_match.coffee
+++ b/app/assets/javascripts/app/controllers/_ui_element/postmaster_match.coffee
@@ -77,12 +77,10 @@ class App.UiElement.postmaster_match
             value:    'x-priority'
             name:     'X-Priority'
           },
-
           {
             value:    'organization'
             name:     'Organization'
           },
-
           {
             value:    'x-original-to'
             name:     'X-Original-To'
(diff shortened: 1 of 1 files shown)


derpixler:
Thank you for this contribution, @HanSolo72! Both changes are very welcome:

Alphabetical sorting of the expert header list is a small but meaningful usability improvement — it makes the list much easier to navigate as it grows over time.

Adding `X-Auto-Response-Suppress` and `X-MS-Exchange-Inbox-Rules-Loop` is a practical addition that helps administrators handle automated replies from Exchange/M365 environments more reliably.
This has been merged into develop and is part of the 7.1 milestone. Thanks for taking the time to identify the issue, implement the fix, and provide a thorough description — much appreciated!

PR #6027, opened 2026-03-19
Force kill leftover gpg-agents

pebosi:
After running zammad for several weeks without restart, there are several (> 2K) gpg-agent processes still running. Running zammad latest 6.5.x via docker.

this is a draft to get rid of those processes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ ENV RAILS_ENV="production" \
 # Use `postgresql-client` meta-package to have the latest `pg_dump` that works even with the latest PostgreSQL versions.
 #   https://github.com/zammad/zammad/issues/6009
 RUN apt-get update -qq && \
-    apt-get install -y postgresql-common && \
+    apt-get install -y tini postgresql-common && \
     /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y && \
     apt-get install --no-install-recommends -y curl libimlib2 libpq5 nginx gnupg postgresql-client && \
     rm -rf /var/lib/apt/lists /var/cache/apt/archives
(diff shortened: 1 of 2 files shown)


pebosi:
Figured out, it's the deployment via docker stack deploy itself, so build a tini-wrapper image here: 
[pebosi/zammad-tini](https://hub.docker.com/r/pebosi/zammad-tini/tags)

dominikklein:
@fliebe92 @mgruner @mantas Any thoughts about a change in this area?

mgruner:
@dominikklein I have seen `tini` in other images before, but have no personal experience in using it. We should be careful here.

fliebe92:
No experience with `tini` at all, sorry. About the still running gpg-agents, maybe we need to check what's going on in more detail, not sure though.

cluck:
Increase the verbosity of gpgconf to find out what's failing, because either there is a deadlock or gpgconf isn't called with the right context. Maybe also use `--status-fd FD` to replace the arbitrary sleep+timeout, and obtain gpgconfs' verdict.

Example call: `LANG=C LC_ALL=C GNUPGHOME=/whatever/path gpgconf --verbose --output /dev/stderr --status-fd 1 --kill gpg-agent`. This will redirect all output to stderr and produce **[GNUPG:] SUCCESS** or **[GNUPG:] FAILURE** on stdout.

mgruner:
Please see https://github.com/zammad/zammad-docker-compose/issues/600. Perhaps this could be a solution also for the current issue?

mgruner:
This should be solved with https://github.com/zammad/zammad-docker-compose/pull/602

PR #6031, opened 2026-03-21
Normalize email recipients for consistent display and delivery

### LINKED ISSUE #4938: Ticket zoom reply via email does not show fullname in To field
### Used Zammad Version

6.1

### Environment

- Installation method: [e.g. source, package]
- Operating system (if you're unsure: `cat /etc/os-release` ): [e.g. debian 10.4, ubuntu 20.04]
- Database + version:  [e.g. postgresql 9.3, mysql 5.7, mariadb 10.3]
- Elasticsearch version: [e.g. 7.17]
- Browser + version: [e.g. chrome 83, safari 14, firefox 105]

### Actual behaviour

Only mail addresses are shown on reply in ticket zoom.

image

### Expected behaviour

When replying to the customer via ticket zoom, it should include the full name of the customer when shown before.

image

- Open ticket zoom.
- Click on email reply.
- Add Nicole Braun and save article.

### Support Ticket

_No response_

yes

HanSolo72:
This PR introduces a central EmailRecipientNormalizer to ensure consistent
formatting of email recipients across email delivery and API responses.

Changes:
- Resolves users by email and formats them as "Firstname Lastname <email>"
- Keeps unknown email addresses unchanged
- Avoids modifying non-email values like 'Users'
- Deduplicates recipients and normalizes casing
- Applies normalization before email delivery
- Applies normalization in Ticket::Article API output (to/cc)

Fixes #4938


HanSolo72:
Hi @DominikKlein,

I’ve addressed the feedback and updated the implementation:

* Removed the N+1 query by preloading users (`User.where(...).index_by`)
* Preserved the original behavior (including case-insensitive matching and deduplication)
* Ensured full fallback to the original input if any unexpected error occurs
* Adjusted and cleaned up the spec accordingly

All Ruby-related checks (RSpec + Rubocop) are passing locally.

However, there is still a failing CI test:

`apps/desktop/pages/ticket/__tests__/ticket-create/ticket-create-user.spec.ts`

The test times out after 30 seconds. Since this PR only touches backend Ruby code, this looks unrelated (possibly flaky or CI-related). I retriggered the pipeline, but it failed again.

Could you please have a look or retrigger the check?

Thanks!

dominikklein:
@mantas @rolfschmidt any thoughts about this one here? :-)

HanSolo72:
Screenshots speak louder than words.

Current version:
image

After my changes:
image

**Important:** The order of the names is not affected by PR 6032. The order "First Name, Last Name" is always used.

mantas:
@dominikklein I think this area is very tricky and prone to potential issues. Some possible issues that come to my mind:

- There's this lovely feature in some cases allowing multiple users with the same email. In such systems there's no way to find which user is which by email. 

- Agents anonymization. Won't this reveal the names of agents? This could be avoided with some extra checks though.

- What if someone wants to enter a custom name different than what is in the system? This would silently override it.

- In an incoming email, this would loose the original metadata?

I'd rather improve this in new ticket form. It looks like even in new tech stack we simply insert email. Instead we could insert both name and email. Possibly with editable name part.

dominikklein:
> There's this lovely feature in some cases, allowing multiple users with the same email. In such systems, there's no way to find which user is which by email.

I think here we have already a general decision on how to go forward in this situation (it's only not implemented in all places and an open story)

--

But yeah, I agree that it's in general a difficult topic. Strange that we have no better relation support here.

Issue #4139, opened 2022-06-21
Lastname in Uppercase is ignored, when first name is optional


ric99cs:
### Infos:

* Used Zammad version: 5.2 and 5.0.1
* Installation method (source, package, ..): Package
* Operating system: Debian 10 & 11
* Database + version:
* Elasticsearch version:
* Browser + version:

I have an workflow, to set the first name as optional, if a lastname is given. Now, when i put an name in upper case in the last name field and left the first name field blank, then the last name is not stored in upper case, only the first character is upper case.

If the first name field is also set, the last name is stored completely in upper case.

### Steps to reproduce the behavior:

* Tested with a fresh install on Debian 11/Zammad 5.2:

Create a workflow, that sets the first name as optional, if the last name is given. Now create a user with the last name in upper case like "TESTUSER", and it will be stored as "Testuser". Setting last name to e.g. "TEST" and first name to "USER" stores both values in upper case.

Yes I'm sure this is a bug and no feature request or a general question.

#### Ticket
- Ticket#10215395

ric99cs:
The same happens, when the name is in lowercase. Also this happens, if you set the lastname as optional if a first name is given. It looks like there is a auto capitalization, that has not to be there. Or there is a auto capitalization that does not work, if both the firstname and the lastname are given.

Martin-Bonn:
I noticed the same issue as capitalization of names synced from LDAP was not being preserved when the user had no first name. In my case: first name = '', last name = "SUPPORT" is synced as first name = '', last name = "Support"

You can also observe this when you try to run `User.find(xy).update(firstname: "", lastname:"SUPPORT")`. This will update to "Support".

I traced this down to app/models/user.rb function check_name.

If first name or last name is empty when calling the function, the function (or rather the called function check_name_apply) will capitalize the name, and names in all caps will be capitalized. 

If both first and last name are present, nothing will be done to them due to 
`return if firstname.present? && lastname.present?` 

I verified this is the cause by temporarily commenting out `self[identifier].capitalize! if self[identifier]&.match? %r{^([[:upper:]]+|[[:lower:]]+)$}` in check_name_apply.

Capitalizing certainly is useful when the code extracts first and last name from an email address like john.doe@example.com.
This need someone with more insight and proper Ruby knowledge to address further. 

[...]

PR #6051, opened 2026-03-28
Fix #5903 disable name guessing for structured user sources

### LINKED ISSUE #5903: LDAP sync updates users on every run when givenName / firstname is empty
### Used Zammad Version

6.5.2

### Environment

- Operating system (if you're unsure: `cat /etc/os-release` ): Debian 12.12
- Database + version: postgresql
- Browser + version: Chrome 143
- LDAP: Microsoft Active Directory
- Authentication source: LDAP
- Mapping: givenName → firstname

### Actual behaviour

- The user is marked as “updated” on every LDAP sync run
- Rails console dump of the user before and after the sync shows identical values
- No visible attribute changes occur
- Setting any non-empty value (e.g. "x") in givenName immediately fixes the issue:
   - first sync updates correctly
   - subsequent syncs are stable

### Expected behaviour

- LDAP sync should be idempotent
- If givenName is empty in LDAP and firstname is empty in Zammad, no update should occur
- Empty / missing LDAP attributes must not trigger repeated updates

1. Use Active Directory as LDAP source
2. Map LDAP attribute givenName → Zammad user field firstname
3. Create or use an AD user where:
4. givenName is not set / empty
5. all other mapped attributes are stable
6. Run LDAP sync
7. Run LDAP sync again

**Technical notes**
- LDAP delivers missing attributes as nil
- Zammad appears to store empty values as ""
- Comparison between nil and empty string causes changed? == true on every run
- This is reproducible with a single user and does not depend on filters or mappings
- Verified via Rails console and repeated sync runs

**Impact**
- Causes constant LDAP updates for affected users
- Pollutes logs and job output
- Makes it hard to detect real changes
- Especially problematic in environments that intentionally use empty givenName to distinguish real users from functional accounts

### Support Ticket

_No response_

yes

HanSolo72:
Fixes #5903

## Summary

`User#check_name` currently guesses and splits names too aggressively when only one of `firstname` or `lastname` is present.

This causes the LDAP issue described in #5903, where repeated updates can happen if `givenName` / `firstname` is empty. It also affects other structured user creation flows where separate `firstname` / `lastname` fields already exist and explicitly provided values should be preserved.

For example, when `firstname` is left blank and `lastname` contains a structured value such as `Zimmermann GmbH`, the value may be split unexpectedly.

## Relation to recent changes / #4139

A recent change already addressed #4139 via `name_from_channel_import` handling for channel-import related name processing.

This PR does not try to replace that logic. Instead, it complements it by covering additional structured user creation flows outside the channel-import context.

In other words: the existing `name_from_channel_import` handling remains responsible for its specific channel-import use case, while this change introduces an explicit `disable_name_guessing` opt-out for structured sources that already provide separate `firstname` / `lastname` fields.

## Changes

- add a transient `disable_name_guessing` accessor to `User`
- update `User#check_name` to skip guessing when that flag is set and either name part is already present
- set `disable_name_guessing = true` for structured imports in the sequencer save path
- set `disable_name_guessing = true` in CSV import
- set `disable_name_guessing = true` in manual user creation via `UsersController`
- set `disable_name_guessing = true` in self signup via `Service::User::Signup`
- add specs covering the disabled-guessing behavior

## Why this approach

A global change of the existing `check_name` behavior would be too broad and would risk changing behavior in flows where guessing is still useful.

Sources like email, web forms with unstructured names, SMS, Facebook, Telegram, or WhatsApp may provide unstructured name data, so the existing guessing behavior should remain intact there.

This change therefore takes an explicit opt-out approach for structured user creation flows that already provide separate `firstname` / `lastname` fields.

## Result

This fixes the repeated LDAP update problem from #5903 and also prevents unintended splitting or rewriting of explicitly provided name values in structured user creation flows such as LDAP/sequencer imports, CSV import, manual user creation, and self signup.

It may also overlap with the underlying symptom described in #4139 for those structured flows, but without changing the existing channel-import specific handling.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -852,6 +852,7 @@ curl http://localhost/api/v1/users/avatar -v -u #{login}:#{password} -H "Content
 
     user = User.new(clean_user_params)
     user.associations_from_param(params)
+    user.disable_name_guessing = true
 
     user.save!
 
--- a/app/models/concerns/can_csv_import.rb
+++ b/app/models/concerns/can_csv_import.rb
@@ -143,6 +143,7 @@ returns
 
                 record = new(param_cleanup(clean_params).reverse_merge(created_by_id: 1, updated_by_id: 1))
                 record.associations_from_param(attributes)
+                record.disable_name_guessing = true if record.is_a?(User)
                 record.save!
               rescue => e
                 errors.push "Line #{i.next}: Unable to create record - #{e.message}"
@@ -158,6 +159,7 @@ returns
                 record.with_lock do
                   record.associations_from_param(attributes)
                   record.assign_attributes(clean_params)
+                  record.disable_name_guessing = true if record.is_a?(User)
                   record.save! if record.changed?
                 end
               rescue => e
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -696,7 +696,7 @@ try to find correct name
     preferences.fetch(:locale) { Locale.default }
   end
 
-  attr_accessor :skip_ensure_uniq_email, :name_from_channel_import
+  attr_accessor :skip_ensure_uniq_email, :name_from_channel_import, :disable_name_guessing
 
   def shared_organizations?
     all_organizations.exists? shared: true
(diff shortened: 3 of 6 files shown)


dominikklein:
I'm not sure, if this is the correct fix. We maybe should do it the other way around, only activate in in special situations, because now we are disabling it in the most situations.

It will also a breaking change (also it's maybe currently not a good behavior), so we need to be careful with changing this.

HanSolo72:
I actually did it the other way around at first. But when I realized there were so many possibilities (Telegram, WhatsApp, etc.) that I wasn't even able to test, it seemed less "breaking" to only implement it where I could assess and test the behavior.

If you'd like me to reverse the logic, let me know, and I'll adjust it accordingly. Or would you and the team prefer to do it yourselves?

PR #6053, opened 2026-03-30
Fix #6033 - backup script issue 

### LINKED ISSUE #6033: Initial backup process fails completely under certain circumstances in Docker context
### Used Zammad Version

7.0.0

### Environment

- Installation method: Docker
- Operating system: ubuntu 22.04, but it doesn't matter
- Database + version:  not relevant
- Elasticsearch version: not relevant
- Browser + version: not relevant

### Actual behaviour

The backup cleanup logic currently uses shell globbing in combination with `find`:

`find "${BACKUP_DIR}"/*_zammad_*.gz -type f -mtime +"${HOLD_DAYS}" -delete`

This causes an issue when no files match the pattern `*_zammad_*.gz`. In that case, the shell does not expand the glob and passes it literally to `find`, which then errors:

`find: ‘/var/tmp/zammad/*_zammad_*.gz’: No such file or directory`

As a result, the script terminates prematurely (especially if the -e option is enabled). And—what's worse—no (initial/new) backups are created.

This affects:

- contrib/docker/backup.sh
- contrib/backup/functions (function delete_old_backups)

### Expected behaviour

The cleanup step should silently do nothing if no matching files exist and continue execution.

Manually empty the backup directory completely (or have it emptied by an external backup process, as was done here). Have a look on `docker-compose logs -f zammad-backup`

### Suggested fix:

Avoid shell globbing and let `find` handle the pattern:

`find "${BACKUP_DIR}" -type f -name "*_zammad_*.gz" -mtime +"${HOLD_DAYS}" -delete`

This approach is robust and avoids errors when no files match.

### Offer:

I’d be happy to provide a pull request that applies this fix consistently across the affected scripts.

Best regards, an thanks for your fantastic work!

### Support Ticket

_No response_

yes

ggruening:
### Summary

This PR fixes an issue in the backup cleanup logic (see #6033) where `find` fails if no files match the glob pattern `*_zammad_*.gz`.

### Problem

The current implementation uses shell globbing:

`find "${BACKUP_DIR}"/*_zammad_*.gz ...`

If no files match, the shell passes the pattern literally to `find`, resulting in:

`find: ‘.../*_zammad_*.gz’: No such file or directory`. 

With `set -e`, this causes the script to exit prematurely. As a result no (initial/new) backups are created.

### Solution

Replace shell globbing with `find -name` so that pattern matching is handled by `find` itself:

`find "${BACKUP_DIR}" -type f -name '*_zammad_*.gz' ...`

### DIFF
--- a/contrib/backup/functions
+++ b/contrib/backup/functions
@@ -60,7 +60,7 @@ function delete_old_backups () {
   # Use -mmin to clean up files as -mtime (days) is too unprecise.
   #   However, add +60 minutes to allow for the backup script run time, so that
   #   backups created a few minutes more than a day ago are not already purged.
-  FILES_TO_DELETE=$(test -d "${BACKUP_DIR}" && find "${BACKUP_DIR}" -type f -name '*_zammad_*.gz' -mmin "+$((60*24*HOLD_DAYS+60))")
+  FILES_TO_DELETE=$(test -d "${BACKUP_DIR}" && find "${BACKUP_DIR}" -maxdepth 1 -type f -name '*_zammad_*.gz' -mmin "+$((60*24*HOLD_DAYS+60))")
 
   if [[ -n ${FILES_TO_DELETE} ]]; then
     if [ "${DEBUG}" == "yes" ]; then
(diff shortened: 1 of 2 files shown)


mgruner:
🚀

fliebe92:
Hi @ggruening and thanks for this pull request.

Due to internal tooling, this pull request will be closed, but don't be worried about it. Your changes are merged to `develop` and will soon be merged to `stable` as well.

Thanks again for your contribution!

PR #6067, opened 2026-04-02
Make OmniAuth provider plugins optional

cryptomilk:
This change allows selective installation of OmniAuth provider gems.

Provider gems are now marked with 'require: false' in the Gemfile and only loaded when needed. The initializer checks if each provider strategy class is defined before registering it.

I'm packaging zammad here: https://build.opensuse.org/package/show/home:gladiac:apps/zammad

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -100,17 +100,47 @@ gem 'webauthn', require: false
 # authentication - third party
 gem 'omniauth-rails_csrf_protection'
 
-# authentication - third party providers
-gem 'omniauth-facebook'
-gem 'omniauth-github'
-gem 'omniauth-gitlab'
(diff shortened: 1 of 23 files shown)


YetAnotherGerrit:
But what are the benefits? This makes installing Zammad much more complex.  Currently, Elasticsearch is the only dependency that needs to be installed separately.

cryptomilk:
Everything is still turned on and installed by default. However I don't want to provide support for them from a packaging standpoint.

dominikklein:
@cryptomilk Thanks for the contribution. Also, it's still installed by default; we currently do not see the need to add more complexity in this direction to our codebase. Maybe we will change our opinion in the long run, but for now I will close the PR.

PR #6069, opened 2026-04-04
Fix Facebook integration for Facebook Login for Business

sherdent:
## Summary

Two issues prevent the Facebook channel from working with newly created Meta developer apps:

1. **Deprecated OAuth scopes** — The currently requested scopes (`pages_manage_posts`, `pages_manage_engagement`, `pages_read_user_content`) are no longer valid in Meta's current Graph API, causing an "Invalid Scopes" error during OAuth authorization. This PR replaces them with the current equivalents: `pages_manage_metadata`, `pages_messaging`, `pages_show_list`, `pages_read_engagement`.

2. **Empty pages with Facebook Login for Business** — Meta now creates apps with "Facebook Login for Business" by default (rather than the legacy "Facebook Login"). This uses [granular permissions](https://developers.facebook.com/docs/permissions/overview/), where page access is granted per-page during the OAuth flow. With granular permissions, the `/me/accounts` endpoint returns an empty array even when the user has authorized specific pages. The fix adds a fallback that inspects the token's `granular_scopes` via the [debug_token](https://developers.facebook.com/docs/graph-api/reference/v21.0/debug_token) endpoint to discover authorized page IDs, then queries each page individually for its name and access token.

## The problem

Without this fix, it is not possible to set up a new Facebook channel in Zammad using a newly created Meta developer app. Users encounter:
- "Invalid Scopes: pages_manage_posts, pages_manage_engagement, pages_read_engagement, pages_read_user_content" during OAuth
- After fixing scopes manually, no Facebook Pages appear in the channel configuration despite the user selecting pages during the OAuth flow

## Changes

- Updated OAuth scopes in `request_account_to_link` to use current Meta Graph API permissions
- Added granular_scopes fallback in `link_account` that activates only when `/me/accounts` returns empty
- Removed stale commented-out scope lines

## Test plan

- [ ] Create a new Meta developer app (type: Business) with "Facebook Login for Business"
- [ ] Add permissions: `pages_manage_metadata`, `pages_messaging`, `pages_show_list`, `pages_read_engagement`
- [ ] Configure Facebook channel in Zammad — OAuth should complete without scope errors
- [ ] Authorized Facebook Pages should appear in the channel configuration
- [ ] Verify backward compatibility: apps using legacy "Facebook Login" (where `/me/accounts` works) should continue to work without hitting the fallback path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/external_credential/facebook.rb
+++ b/lib/external_credential/facebook.rb
@@ -32,9 +32,7 @@ class ExternalCredential::Facebook
     state = SecureRandom.uuid
     {
       request_token: state,
-      # authorize_url: oauth.url_for_oauth_code(permissions: 'publish_pages, manage_pages, user_posts', state: state),
-      # authorize_url: oauth.url_for_oauth_code(permissions: 'publish_pages, manage_pages', state: state),
-      authorize_url: oauth.url_for_oauth_code(permissions: 'pages_manage_posts, pages_manage_engagement, pages_manage_metadata, pages_read_engagement, pages_read_user_content', state: state),
+      authorize_url: oauth.url_for_oauth_code(permissions: 'pages_manage_metadata, pages_messaging, pages_show_list, pages_read_engagement', state: state),
     }
   end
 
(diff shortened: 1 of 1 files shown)


mantas:
Hi @sherdent 

Does this change works with old apps using outdated APIs?

sherdent:
Hi @mantas — the change is designed to be backward compatible:

1. **Scopes**: The new scopes (`pages_manage_metadata`, `pages_messaging`, `pages_show_list`, `pages_read_engagement`) have been available since Graph API v7.0 (2020), so they should work with both old and new apps. The scopes they replace (`pages_manage_posts`, `pages_manage_engagement`, `pages_read_user_content`) are now rejected by Meta for newly created apps.

2. **Granular scopes fallback**: This only activates when `/me/accounts` returns an empty array. For existing apps using legacy "Facebook Login" (not "Facebook Login for Business"), `/me/accounts` should continue to work normally and the fallback would never be reached.

However, I've only tested this with a newly created "Facebook Login for Business" app, which is what prompted the fix. I don't have an old-style app to verify against.

MedPlex98:
Hey, any Updates on this? I can't add Facebook Messenger because of this problem

PR #6090, opened 2026-04-19
Fix PreparedStatementInvalid when multi_tree_select / multiselect condition values contain '?'

### LINKED ISSUE #6091: Invalid object selector conditions if value contains a question mark
## Summary
Saving a Trigger / Overview / Scheduler / Core Workflow with a `contains one` / `contains all` condition on a `multiselect` or `multi_tree_select` attribute fails with HTTP 422 `Invalid object selector conditions` whenever **any of the selected values contains a literal `?` character**. The underlying error is `ActiveRecord::PreparedStatementInvalid: wrong number of bind variables`.

## Used Zammad Version
6.6.x (reproduced on self-hosted Docker deployment on Zammad 7.0.1, railsserver image). `sql_helper.rb` on `develop` still exhibits the same pattern, so this is current.

## Environment
- Installation method: official `zammad-docker-compose`
- Database: PostgreSQL 17
- Ruby 3.4.0, Rails/ActiveRecord 8.0.5
- Browser: any (issue is server-side)

## Actual behaviour
Any attempt to save a Trigger that uses `contains one` on a `multi_tree_select` attribute whose selected values include any string containing `?` fails:

```
ERROR -- : Invalid object selector conditions (Exceptions::InvalidAttribute)
app/models/concerns/checks_condition_validation.rb:12:in 'ChecksConditionValidation#validate_condition'
```

422 Unprocessable Entity is returned. No `Selector::Sql.get->InvalidCondition` line in the log because the error happens inside `Ticket.selectors` (called via `Selector::Sql#valid?`), where the bare `rescue` in `valid?` swallows it.

Running the same validation in a Rails console surfaces the real error:

```ruby
trig = Trigger.find(<id>)
Ticket.selectors(trig.condition, current_user: User.find(1), limit: 1, ticket_id: 1, access: 'ignore')
# => ActiveRecord::PreparedStatementInvalid: wrong number of bind variables (1 for 2) in:
#    "(((("tickets"."id" IN (?)) AND (("tickets"."cs_contact_reason" && ARRAY['...value with ? inside...']::varchar[])))))"
```

## Expected behaviour
A tree-select value that contains a literal `?` character should save and validate just like any other value.

## Steps to reproduce

1. Create a `multi_tree_select` object attribute (e.g. `contact_reason`) on `Ticket` with at least one option whose path contains `?`, for example `Payments::Why was I charged?::Explained T&C`.
2. Migrate the DB so the attribute is active.
3. Create a Trigger with condition: `ticket.contact_reason` â€” operator `contains one` â€” value `['Payments::Why was I charged?::Explained T&C']`, and any simple perform.
4. Save the trigger (UI or API `POST /api/v1/triggers`).
   - Save fails with HTTP 422.
5. Remove the `?` from the value (or switch to a non-`?` sibling) â€” save succeeds.

## Root cause

`lib/sql_helper.rb` methods `array_contains_one` and `array_contains_all` build SQL by inlining the values directly into the `ARRAY[...]::varchar[]` literal, using `quote_string` to escape single quotes but **doing nothing about `?`**:

```ruby
def array_contains_one(attribute, value, negated: false)
  value = [''] if value.blank?
  value = Array(value)
  result = "(#{db_column(attribute)} && ARRAY[#{value.map { |v| "'#{self.class.quote_string(v)}'" }.join(',')}]::varchar[])"
  negated ? "NOT(#{result})" : "(#{result})"
end
```

The resulting fragment is later concatenated into the full selector query and passed to ActiveRecord as `relation.where(query_sql, *bind_params)`. AR 8's `build_bound_sql_literal` walks the query string and treats **every** literal `?` â€” including ones inside single-quoted SQL string literals â€” as a bind placeholder. Because the helper emits no bind param for those, the count mismatches and AR raises `PreparedStatementInvalid: wrong number of bind variables (N for M)`.

`valid?` swallows the exception (bare `rescue` returning `false`), which surfaces to the user as the opaque `Invalid object selector conditions` 422.

## Suggested fix

Split out `?` when building the quoted literal and concatenate with `chr(63)` so the resulting SQL no longer contains a literal `?` that AR can misinterpret:

```ruby
def quote_array_literal(value)
  quoted = self.class.quote_string(value.to_s)
  return "'#{quoted}'" unless quoted.include?('?')
  quoted.split('?', -1).map { |p| "'#{p}'" }.join(' || chr(63) || ')
end

def array_contains_one(attribute, value, negated: false)
  value = [''] if value.blank?
  value = Array(value)
  literal = Array(value).map { |v| quote_array_literal(v) }.join(',')
  result = "(#{db_column(attribute)} && ARRAY[#{literal}]::varchar[])"
  negated ? "NOT(#{result})" : "(#{result})"
end
# same for array_contains_all
```

I'll follow up with a PR including a regression spec.

## Impact / severity
- Blocks admins from saving triggers/overviews/schedulers/core workflows when operational vocabularies naturally include `?` (common in FAQ/contact-reason trees â€” e.g. "Why was I charged?", "Has my Extra Hold been released?", "Customer OK with outcome?").
- Also swallows the real error, so admins see a generic 422 that is very hard to debug without access to Rails logs.
- Workaround is unappealing: rename all tree nodes to remove `?`, losing an established business vocabulary.

GabrAGM:
## Summary

Fixes #6091. Saving a Trigger / Overview / Scheduler / Core Workflow with a `contains one` / `contains all` condition on a `multiselect` or `multi_tree_select` attribute currently fails with HTTP 422 `Invalid object selector conditions` whenever any selected value contains a literal `?` character. Under the hood this surfaces as `ActiveRecord::PreparedStatementInvalid: wrong number of bind variables (N for M)` — `Selector::Sql#valid?`'s bare `rescue` swallows it and returns `false`, which admins then see as the opaque 422.

## Root cause

`lib/sql_helper.rb#array_contains_one` and `array_contains_all` build SQL by inlining values directly into the `ARRAY[…]::varchar[]` literal. The resulting fragment is later passed to `relation.where(query_sql, *bind_params)`. AR 8's `build_bound_sql_literal` scans the entire query string for `?` — including occurrences inside quoted string literals — and treats each as a bind placeholder. Because the helper emits no bind for those, the placeholder count mismatches → `PreparedStatementInvalid`.

This matters in practice because natural-language operational vocabularies commonly use `?` (FAQs, contact-reason trees, satisfaction flags — e.g. "Why was I charged?", "Has my Extra Hold been released?", "Customer OK with outcome?").

## Fix

Add `SqlHelper.quote_array_literal(value)` that splits the value on `?` and joins the segments with `|| chr(63) ||` so the emitted SQL no longer contains a raw `?`. Route both `array_contains_one` and `array_contains_all` through it.

- `chr(63)` is ASCII `?` in PostgreSQL; the concatenation evaluates at query time to the exact original string, so no semantic change for any existing value. Only values that actually contain `?` get the rewrite.
- API is unchanged; this is a purely internal escape strategy.

## Test plan

- [x] Added `spec/lib/sql_helper_spec.rb` covering:
  - `quote_array_literal` for simple values, values with single quotes, single `?`, multiple `?`.
  - `array_contains_one` / `array_contains_all` emit no raw `?` and preserve semantics.
  - Integration: `Ticket.selectors` with a `multi_tree_select` condition whose value contains `?` does not raise `PreparedStatementInvalid`.
- [x] Manual repro on Zammad 7.0.1 / PostgreSQL 17 / Rails 8.0.5: without patch — 422 `Invalid object selector conditions` + `PreparedStatementInvalid (1 for 2)` in Rails console. With patch — trigger save 200 OK, SQL fragment contains `|| chr(63) ||` instead of raw `?`, returned row count is correct.

## Related

- Issue: #6091

### DIFF
--- a/lib/sql_helper.rb
+++ b/lib/sql_helper.rb
@@ -10,6 +10,26 @@ class SqlHelper
     ApplicationModel.sanitize_sql_like(...)
   end
 
+  # Quote a value for inclusion as a literal element inside an ARRAY[...]
+  # SQL expression.
+  #
+  # The resulting fragment is later concatenated into a larger SQL string
+  # and passed to ActiveRecord as `relation.where(query_sql, *bind_params)`.
+  # ActiveRecord scans the query string for `?` bind placeholders across
+  # the entire string — it does not exclude occurrences inside quoted
+  # literals. Any `?` inside a value would therefore be mistaken for an
+  # unfilled bind placeholder and trigger
+  # `ActiveRecord::PreparedStatementInvalid: wrong number of bind
+  # variables`. To avoid this, split the value on `?` and concatenate the
+  # segments with `chr(63)` so the resulting SQL no longer contains a
+  # literal `?` character.
+  def self.quote_array_literal(value)
+    quoted = quote_string(value.to_s)
+    return "'#{quoted}'" unless quoted.include?('?')
+
+    quoted.split('?', -1).map { |segment| "'#{segment}'" }.join(' || chr(63) || ')
+  end
+
   def initialize(object:, table_name: nil)
     @object     = object
     @table_name = table_name
@@ -198,7 +218,7 @@ sql = 'tickets.created_at ASC, tickets.updated_at DESC'
   def array_contains_all(attribute, value, negated: false)
     value = [''] if value.blank?
     value = Array(value)
-    result = "(#{db_column(attribute)} @> ARRAY[#{value.map { |v| "'#{self.class.quote_string(v)}'" }.join(',')}]::varchar[])"
+    result = "(#{db_column(attribute)} @> ARRAY[#{value.map { |v| self.class.quote_array_literal(v) }.join(',')}]::varchar[])"
 
     negated ? "NOT(#{result})" : "(#{result})"
   end
@@ -206,7 +226,7 @@ sql = 'tickets.created_at ASC, tickets.updated_at DESC'
   def array_contains_one(attribute, value, negated: false)
     value = [''] if value.blank?
     value = Array(value)
-    result = "(#{db_column(attribute)} && ARRAY[#{value.map { |v| "'#{self.class.quote_string(v)}'" }.join(',')}]::varchar[])"
+    result = "(#{db_column(attribute)} && ARRAY[#{value.map { |v| self.class.quote_array_literal(v) }.join(',')}]::varchar[])"
 
     negated ? "NOT(#{result})" : "(#{result})"
   end
--- a/spec/lib/sql_helper_spec.rb
+++ b/spec/lib/sql_helper_spec.rb
@@ -1,0 +1,97 @@
+# Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
+
+require 'rails_helper'
+
+RSpec.describe SqlHelper do
+  describe '.quote_array_literal' do
+    it 'wraps a simple value in single quotes' do
+      expect(described_class.quote_array_literal('foo')).to eq("'foo'")
+    end
+
+    it 'doubles embedded single quotes' do
+      expect(described_class.quote_array_literal("foo's")).to eq("'foo''s'")
+    end
+
+    it 'splits on ? so the fragment does not contain a raw ? placeholder (#6089)' do
+      # A literal ? inside the quoted value would otherwise be counted as an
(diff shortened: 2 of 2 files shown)


fliebe92:
Hi @GabrAGM and thanks for your pull request.

After some internal discussion, we will address it in another way, however, thanks for spending your valuable time and for bringing it to our attention.

PR #6100, opened 2026-04-27
Allow same-browser session tabs to coexist

Kian0815:
## Summary

This change allows multiple browser tabs in the same browser profile to stay active for the same user.

Previously, opening a second tab could trigger the legacy `session_takeover` blocker in the first tab because `App.TaskManager.TaskbarId()` generated a random in-memory value per tab. This made tabs from the same browser look like competing sessions.

The taskbar id is now persisted in `localStorage`, so same-browser tabs share the same id. If `localStorage` is unavailable, the existing in-memory fallback is still used.

## Changes

- Persist the legacy task manager tab/session id in `localStorage`.
- Re-read the stored id before comparing takeover messages, so tabs opened close together converge on the same browser id.
- Update the system spec to assert that opening another browser tab does not stop the existing tab.

## Verification

Ran the targeted system spec locally:

```sh
RAILS_ENV=test VITE_TEST_MODE=1 CI_SKIP_ASSETS_PRECOMPILE=1 \
DATABASE_URL=postgresql://test@127.0.0.1:5432/zammad_test \
REDIS_URL=redis://127.0.0.1:6379 \
SELENIUM_BROWSER=chrome SELENIUM_BROWSER_HEADLESS=true \
bundle _2.6.9_ exec rspec spec/system/basic/session_takeover_spec.rb

### DIFF
--- a/app/assets/javascripts/app/lib/app_post/task_manager/singleton.coffee
+++ b/app/assets/javascripts/app/lib/app_post/task_manager/singleton.coffee
@@ -447,8 +447,22 @@ class App.TaskManagerSingleton extends App.Controller
     false
 
   TaskbarId: =>
-    if !@TaskbarIdInt
-      @TaskbarIdInt = Math.floor( Math.random() * 99999999 )
+    storageKey = 'taskbar_id'
+
+    try
+      storedTaskbarId = App.LocalStorage.get(storageKey)
+
+      if storedTaskbarId
+        @TaskbarIdInt = storedTaskbarId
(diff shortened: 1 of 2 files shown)


dominikklein:
@Kian0815 Thanks for the contribution. I think the implementation will not work, when you think about multiple browsers and not only the same browser with multiple tabs (and it has also other side effects which can happen).
In the new desktop-view we are already supporting multiple tabs, so there is no plan to improve this currently in the old desktop app.

Because of that I will close the PR :-)

PR #6106, opened 2026-04-29
feat(editor): add Highlight + TextAlign toolbar actions

GabrAGM:
## Summary

Adds two missing formatting capabilities to the desktop ticket editor toolbar:

- **Highlight** — background-color marker via `@tiptap/extension-highlight` (multicolor enabled)
- **Text alignment** — left / center / right / justify via `@tiptap/extension-text-align`, applied to `heading` and `paragraph` nodes

The current editor already wires Color, Indent/Outdent, Tables, etc., but agents have no way to highlight passages or center/justify a paragraph in a ticket reply. These are common requests for outbound customer communication.

## Changes

- `package.json` — pin `@tiptap/extension-highlight@3.19.0` and `@tiptap/extension-text-align@3.19.0` to match the rest of the `@tiptap/*` stack
- `pnpm-lock.yaml` — regenerated
- `app/frontend/shared/components/Form/fields/FieldEditor/extensions.ts` — register both extensions in `getHtmlExtensions()`
- `app/frontend/apps/desktop/components/Form/fields/FieldEditor/useEditorActions.ts` — add a Highlight button next to the Color button, and a Text-Alignment submenu with four entries (left/center/right/justify)

## Test plan
- [ ] Open a ticket reply, verify the new Highlight button toggles a yellow background mark on the selection
- [ ] Verify the Text-Alignment submenu sets `style="text-align: …"` on the surrounding paragraph/heading
- [ ] Confirm round-trip: save → close → reopen → formatting persists in the article body
- [ ] Plain-text mode unaffected (extensions only registered for `getHtmlExtensions`)

Built and deployed against Zammad 7.0.1 in production at AGM Egypt.


fliebe92:
Hi @GabrAGM. 876 changed files looks like an accident to me, could you please check?

GabrAGM:
Hi @fliebe92, you're right — the original push was branched from the 7.0.1 release tag (which sits on the stable branch and has diverged ~600 commits from develop), so GitHub showed every file that differs between stable and develop. Just rebased the branch onto current upstream/develop and re-applied the commit on top — the diff is now the intended **4 files / +291 / −139** (`package.json`, `pnpm-lock.yaml`, `extensions.ts`, `useEditorActions.ts`). State is now `MERGEABLE`. Thanks for the catch!

fliebe92:
Ping @ivan-abrash + @YetAnotherGerrit. Product decision needed :)

dominikklein:
I think about the highlight color there was already some discussion if it's a good idea, because it will conflict with our "marker" feature.

(About adding some stuff inside the old editor toolbar, I'm not sure in general)

YetAnotherGerrit:
Hi @GabrAGM, thank you very much for your contribution. However, as you may already be aware, we are developing a brand new editor behind the scenes. Compared to the current editor, it has a lot of advanced features. This new editor has a carefully selected colour palette that works in both dark and light modes. The colours are actually slightly different depending on the mode. We decided against introducing background colours to this feature for several reasons. This doesn't mean that we won't introduce them ever; we just don't want them included at the start of the new editor.

Generally speaking, we don't want to include new features in the old editor either at this time. We will consider alignment options in the future.

Please be patient and try the new editor once it is released in beta. Unfortunately, there is no ETA yet.

Issue #581, opened 2017-01-03
orward attachments to the customer/agent per note


ECM80:
Hello,

- It is not possible to forward attachments to the customer per note.
- It is not possible to forward attachments to the agent by email if a customer opens a ticket.
- If an agent on an e-mail replies and an attachment sent with, this is not delivered to the customer.

greetings

ECM80

martinvonwittich:
Forwarding complete tickets would be a useful feature. Our old ticket system Kayako had a feature for this, though it wasn't necessarily implemented very well - Kayako just seems to take the plaintext part of the articles and creates a new plaintext mail that looks like this:

```
test



Ticket History
===================
Martin von Wittich (Staff) Posted On: 20.11.2018 13:51

===============================================================
test



Martin von Wittich (Staff) Posted On: 20.11.2018 13:53

===============================================================
test




Martin von Wittich (Staff) Posted On: 20.11.2018 13:54

===============================================================
test




Martin von Wittich (Staff) Posted On: 20.11.2018 13:54

===============================================================
test








Ticket Details
---------------------------------
Ticket ID: QRW-516-55512
Department: Raw
Type: Issue
Status: Geschlossen
Priority: Normal
```

Attachments of the respective articles are jumbled together and attached to the resulting mail, and it's not obvious which attachment belonged to which article. I'm not sure how it would've handled HTML articles (especially when the original mail doesn't have a plaintext part), but knowing Kayako, it's probably safe to assume that it would handle that badly.

I'm not sure what the best solution would be here. At first I considered attaching the original messages as *.eml files, but a quick test shows that at least GMail doesn't handle that well - it just offers to download these attached emails, and doesn't natively support displaying them properly. Then I considered multipart/digest which sounds like it's designed exactly for this usecase, but GMail seems to handle that even worse, just more or less displaying the mail source code, including MIME boundaries :/

On the plus side: Thunderbird seems to handle both cases well. EML files have the disadvantage that at least in Ubuntu I have to tell Ubuntu to open the attached EML files with Thunderbird before I can view the attached mails. multipart/digest looks very nice in Thunderbird.

MrGeneration:
hey Martin,

thanks for your participation - please don't recycle issues that are not matching the issue you're posting about. Also this is covered by #573 .

Edit: New feature request can be posted at our community: https://community.zammad.org

martinvonwittich:
[...]

PR #6112, opened 2026-05-03
feat(sso): support return_to query param for /auth/sso deep-link redirects

MichaelUray:
## Why

Today `/auth/sso` always redirects to `/#` after successful SSO authentication, which discards any URL hash the browser brought along. That makes deep-linking from an upstream portal into a specific Zammad ticket (e.g. `/#ticket/zoom/57`) effectively impossible — server redirects don't carry hash fragments, and modern browsers (Chromium 121+) block the cross-origin sub-resource cookie tricks that used to work around this. The remaining workaround (a JavaScript popup that establishes the session and a second top-level navigation in the original tab) is fragile and popup-blocker prone.

## What

Adds an optional `return_to` query parameter to `GET/POST /auth/sso`. When present and pointing to a safe relative path, it is used as the redirect target instead of the hardcoded `/#`.

## Security

`return_to` is validated as a strict same-origin relative path:

- Must start with a single `/`
- Must NOT start with `//` (protocol-relative URLs)
- Must NOT contain control characters or backslashes
- Must be at most 1024 chars

Anything that doesn't pass these checks falls back to the previous `/#` default — open-redirect attempts are silently ignored, not echoed back to the user. The validator lives in a small private helper (`safe_return_to_path`) so it can be unit-tested and reused if needed.

## Tests

New request specs in `spec/requests/session_spec.rb` cover:

- default behaviour (no `return_to` → `/#`)
- valid relative path with hash (`/#ticket/zoom/57`)
- preservation of URL fragments
- deeper relative paths without a fragment
- empty `return_to` falls back to default
- rejection of absolute http/https URLs
- rejection of protocol-relative URLs (`//evil.example.com`)
- rejection of `javascript:` and `data:` URLs
- rejection of backslashes and control characters
- rejection of paths longer than 1024 chars
- rejection of paths not starting with `/`

CI on the fork (full test suite, identical workflow as upstream) passes: [run 25288084186](https://github.com/MichaelUray/zammad/actions/runs/25288084186).

## Backwards compatibility

No callers are required to pass `return_to`. Existing integrations (which never set the parameter) continue to land on `/#` exactly as before.

## Use case

External SSO portal (Traefik ForwardAuth + REMOTE_USER) generates Zammad links pointing to specific tickets. Without `return_to`, the portal must use a fragile two-tab popup workaround to make deep-link navigation survive the SSO round-trip. With this change a single 302 to `/auth/sso?return_to=/%23ticket/zoom/<id>` is enough.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -66,7 +66,7 @@ class SessionsController < ApplicationController
 
     initiate_session_for(user, 'SSO')
 
-    redirect_after_omniauth('/#')
+    redirect_after_omniauth(safe_return_to_path(params[:return_to]))
   end
 
   # "Delete" a login, aka "log the user out"
@@ -291,6 +291,28 @@ class SessionsController < ApplicationController
     return if trusted_ips.blank?
 
     raise Exceptions::Forbidden, __('SSO request from untrusted IP address.') if trusted_ips.exclude?(request.remote_ip)
+  end
+
+  SAFE_RETURN_TO_DEFAULT = '/#'.freeze
+  SAFE_RETURN_TO_MAX_LENGTH = 1024
+
+  # Validate the optional `return_to` parameter for the SSO flow.
+  #
+  # Only same-origin, relative paths are accepted to prevent open-redirect attacks:
+  #   * Must start with a single "/" and not "//" (which would be protocol-relative)
+  #   * Must not contain control characters or backslashes
+  #   * Must be at most SAFE_RETURN_TO_MAX_LENGTH characters long
+  #
+  # Anything that does not pass these checks falls back to SAFE_RETURN_TO_DEFAULT.
+  def safe_return_to_path(value)
+    return SAFE_RETURN_TO_DEFAULT if value.blank?
+    return SAFE_RETURN_TO_DEFAULT if !value.is_a?(String)
+    return SAFE_RETURN_TO_DEFAULT if value.length > SAFE_RETURN_TO_MAX_LENGTH
+    return SAFE_RETURN_TO_DEFAULT if !value.start_with?('/')
+    return SAFE_RETURN_TO_DEFAULT if value.start_with?('//')
+    return SAFE_RETURN_TO_DEFAULT if value.match?(%r{[\x00-\x1f\x7f\\]})
+
+    value
   end
 
   def authenticate_with_password
--- a/spec/requests/session_spec.rb
+++ b/spec/requests/session_spec.rb
@@ -277,6 +277,87 @@ RSpec.describe 'Sessions endpoints', type: :request do
       end
     end
 
+    context 'with optional return_to parameter' do
+      let(:user)    { create(:agent) }
+      let(:headers) { { 'X-Forwarded-User' => user.login } }
+
+      it 'redirects to "/#" when no return_to is given' do
+        get '/auth/sso', as: :json, headers: headers
+
+        expect(response).to redirect_to('/#')
+      end
+
+      it 'redirects to the given relative return_to path' do
+        get '/auth/sso', params: { return_to: '/#ticket/zoom/57' }, as: :json, headers: headers
+
+        expect(response).to redirect_to('/#ticket/zoom/57')
+      end
+
+      it 'preserves URL fragments' do
+        get '/auth/sso', params: { return_to: '/#dashboard' }, as: :json, headers: headers
+
+        expect(response).to redirect_to('/#dashboard')
+      end
+
+      it 'accepts deeper relative paths without a fragment' do
+        get '/auth/sso', params: { return_to: '/some/path' }, as: :json, headers: headers
+
+        expect(response).to redirect_to('/some/path')
+      end
+
+      it 'falls back to "/#" when return_to is empty' do
+        get '/auth/sso', params: { return_to: '' }, as: :json, headers: headers
+
+        expect(response).to redirect_to('/#')
+      end
+
+      it 'rejects absolute http/https URLs' do
+        %w[https://evil.example.com/ http://evil.example.com/path].each do |bad|
+          get '/auth/sso', params: { return_to: bad }, as: :json, headers: headers
+
+          expect(response).to redirect_to('/#')
+        end
+      end
+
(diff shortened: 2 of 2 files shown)


fliebe92:
Hi @MichaelUray and thanks for your contribution. Could you please check if #6034 also works for you? Thanks for your feedback.

MichaelUray:
Hi @fliebe92, thanks for the pointer — I've read through #6034.

It addresses a related but distinct problem and unfortunately does **not** cover this PR's use case. A short comparison so we're on the same page:

|                          | #6034 (stephaneberle9)                                                                                | This PR (#6112)                                                                                |
|--------------------------|--------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------|
| Scenario                 | An unauthenticated user is mid-OAuth-flow on `/oauth/authorize` and gets sent through login first      | An external portal generates a one-shot link to a specific ticket and uses `/auth/sso` to log the user in |
| Where the target is set  | Server-side, by Doorkeeper's `resource_owner_authenticator` writing `session[:doorkeeper_return_to]`    | Client-side, by the upstream caller passing `?return_to=…` on `/auth/sso`                       |
| Validation               | Whitelisted to paths starting with `/oauth/authorize`                                                  | Validated as a generic same-origin relative path (must start with a single `/`, no protocol-relative `//…`, no `javascript:`/`data:`, no control chars or backslashes, ≤1024 chars) |
| Login paths it covers    | Password (legacy + desktop SPA), OmniAuth, `REMOTE_USER`/`X-Forwarded-User` SSO                        | Only `/auth/sso` (`REMOTE_USER`/`X-Forwarded-User`), since that's the endpoint with the deep-link problem |
| Trigger                  | Implicit — the user wasn't on a login page when they started; Zammad has to remember their target     | Explicit — the caller knows where it wants the user to land and tells Zammad in the URL          |

In our setup the user clicks a link in an upstream portal that points at e.g. `https://zammad.example/auth/sso?return_to=%2F%23ticket%2Fzoom%2F57`. The reverse proxy (Traefik ForwardAuth) injects `X-Forwarded-User: alice@example.com`, Zammad authenticates the SSO request — and today, regardless of any URL hash on the original link, the response is `redirect_to '/#'` and the deep link is lost. No Doorkeeper / `/oauth/authorize` is involved at any point in this flow.

With #6034 alone in this scenario:

1. There is no preceding `/oauth/authorize` request, so `session[:doorkeeper_return_to]` is never written.
2. `create_sso` falls through to the unchanged `redirect_to '/#'` and the deep link is lost — exactly as today.

The hardcoded `start_with?('/oauth/authorize')` guard in #6034 is also intentionally narrow: relaxing it would defeat its open-redirect protection (fine for that PR's narrow case, but not safe to repurpose for arbitrary deep links).

I see the two PRs as complementary rather than overlapping:

- #6034 = "Zammad knows where the user was going (`/oauth/authorize`) before login; resume that after login."
- #6112 = "An external trusted caller tells Zammad where the user should go after `/auth/sso`; honour that, with strict validation."

### Backwards compatibility

`return_to` is a new optional parameter. Callers that don't pass it land on `/#` exactly as before — no behavior change for any existing integration.

### CI

Full Zammad CI on the fork is green on the current branch tip: <https://github.com/MichaelUray/zammad/actions/runs/25288084186> (run took ~1h 18m, identical workflow to the upstream one).

### Coordination

Happy to do whichever of these makes review easier on your side:

1. **Land both as-is.** They touch the same `redirect_to '/#'` line in `create_sso` but their conditions are mutually exclusive (`session[:doorkeeper_return_to]` vs. `params[:return_to]`). I can rebase on top of #6034 once it's merged so the diff is just the new branch.
2. **Extract a shared helper.** Pull the path validation into something like `Auth::SafeReturnPath` (or a private method on `SessionsController`) and have both PRs use it, with #6034 supplying its `/oauth/authorize` allowlist as an extra check on top of the same-origin-relative-path basics.
3. **Drop this in favour of a follow-up to #6034** that generalises its session-stored target. I'd argue against this option only because it forces every external caller to first hit some Zammad endpoint that writes the session value — vs. a simple URL parameter on `/auth/sso` that any reverse-proxy SSO setup can already produce.

Just let me know which direction you'd like and I'll adjust accordingly.

fliebe92:
Hi @MichaelUray and thanks for the comparison and explanation, really appreciated!

We're in discussing and testing #6034. I would suggest pinging you when your PR could get rebased.

fliebe92:
Hi @MichaelUray. Reviewing + testing of #6034 is done by now, and changes are present in `develop`. Feel free to rebase your pull request.

MichaelUray:
Hi @fliebe92, rebased on top of `develop` (which now contains #6034). The merge conflict in `create_sso` was the expected one at the `redirect_to '/#'` line — resolved by combining both PRs through the new `redirect_after_omniauth` helper:

```ruby
redirect_after_omniauth(safe_return_to_path(params[:return_to]))
```

Doorkeeper's `session[:doorkeeper_return_to]` keeps priority (handled inside `redirect_after_omniauth`); only when no doorkeeper resume is pending does the new `return_to` query param come into play. The two flows stay mutually exclusive, no behavior change for existing integrations.

Full CI on the fork is green for the rebased commit (`a00e4ea9`): https://github.com/MichaelUray/zammad/actions/runs/25577048364 (1h 8m, all 8166 examples pass). Upstream CI here is in `action_required` because of the fork-PR approval policy — happy whenever you have a moment to "Approve and run".

MichaelUray:
@fliebe92 thanks for adressing [#581](https://github.com/zammad/zammad-docker-compose/issues/581), by that, this PR should be mergeable.

MichaelUray:
Hi @fliebe92 — gentle follow-up on this one. I've rebased the branch onto current `develop` (no conflicts), and the previously-red `docker-ci` check now passes since #581 was fixed. Full CI is green on my fork against the latest develop. Whenever you have a moment to approve-and-run the upstream CI and take a look, that'd be appreciated — no urgency, just making sure it hasn't slipped off the radar.

fliebe92:
Hi @MichaelUray. I'm sorry, we were busy with other topics that have a higher priority as this one. I'll try to look into it again as soon as possible. Stay tuned!

PR #6170, opened 2026-05-29
Add separate permissions for accessing the i-doit, GitHub and GitLab integrations as an agent

paranmaum:
This PR adds separate permissions for accessing the i-doit, GitHub and GitLab integrations as an agent. Before, every agent always had access to them.

At our org, Zammad is used by multiple departments in agent roles but only some should have access to the i-doit integration. With this PR, this becomes possible.

**Warning**: While I manually verified the code changes and functionality, I want to point out that this PR was prepared with the help of LLMs.

## Changes

- Added 3 new permissions (integration.idoit, integration.github, integration.gitlab) in the backend (REST controller policies + GraphQL mutations/queries)
- Show/hide the integrations in the ticket sidebar in both the old Coffescript and new Vue frontend according to the new permissions
- Added migration to create the permissions and grant it to all existing agent roles (so the effective permissions on existing installations don't change after updating)
- Regenerated the translation catalog (zammad.pot)

## Tests

Added:
- new migration spec: verify permissions are created and granted to existing agents on upgrade
- New authorization cases on the endpoints (GraphQL and REST)

Updated:
- Existing frontend sidebar specs updated to require new permissions

Ran:
Backend (Rspec)
```
RAILS_ENV=test VITE_TEST_MODE=1 bundle exec rspec \
  spec/db/migrate/add_integration_permissions_spec.rb \
  spec/graphql/gql/mutations/ticket/external_references/ \
  spec/graphql/gql/queries/autocomplete_search/idoit_object_types_spec.rb \
  spec/graphql/gql/queries/ticket/external_references/ \
  spec/requests/integration/idoit_spec.rb
```

Frontend (Vitest)
```
pnpm test --run \
  app/frontend/apps/desktop/pages/ticket/__tests__/ticket-create/ticket-create-github.spec.ts \
  app/frontend/apps/desktop/pages/ticket/__tests__/ticket-create/ticket-create-gitlab.spec.ts \
  app/frontend/apps/desktop/pages/ticket/__tests__/ticket-create/ticket-create-idoit.spec.ts \
  app/frontend/apps/desktop/pages/ticket/__tests__/ticket-detail-view/ticket-detail-view-github.spec.ts \
  app/frontend/apps/desktop/pages/ticket/__tests__/ticket-detail-view/ticket-detail-view-gitlab.spec.ts \
  app/frontend/apps/desktop/pages/ticket/__tests__/ticket-detail-view/ticket-detail-view-idoit.spec.ts
```

rubocop
```
bundle exec rubocop $(git diff --name-only 86a22e48dc...HEAD -- '*.rb')
```

All tests pass. Behaviour was also manually tested.

### DIFF
--- a/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_git_issue.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_git_issue.coffee
@@ -10,6 +10,7 @@ class App.SidebarGitIssue extends App.Controller
 
   sidebarItem: =>
     return if !@Config.get("#{@providerIdentifier}_integration")
+    return if !@permissionCheck("integration.#{@providerIdentifier}")
 
     isAgentTicketZoom   = (@ticket and @ticket.currentView() is 'agent')
     isAgentTicketCreate = (!@ticket and @taskKey and @taskKey.match('TicketCreateScreen-'))
--- a/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_idoit.coffee
+++ b/app/assets/javascripts/app/controllers/ticket_zoom/sidebar_idoit.coffee
@@ -1,6 +1,7 @@
 class SidebarIdoit extends App.Controller
   sidebarItem: =>
     return if !@Config.get('idoit_integration')
+    return if !@permissionCheck('integration.idoit')
 
     isAgentTicketZoom   = (@ticket and @ticket.currentView() is 'agent')
     isAgentTicketCreate = (!@ticket and @taskKey and @taskKey.match('TicketCreateScreen-'))
--- a/app/frontend/apps/desktop/pages/ticket/__tests__/ticket-create/ticket-create-github.spec.ts
+++ b/app/frontend/apps/desktop/pages/ticket/__tests__/ticket-create/ticket-create-github.spec.ts
@@ -27,7 +27,7 @@ import { mockTicketExternalReferencesIssueTrackerItemListQuery } from '../../gra
 
 describe('Ticket create GitHub links', () => {
   it('displays sidebar', async () => {
-    mockPermissions(['ticket.agent'])
+    mockPermissions(['ticket.agent', 'integration.github'])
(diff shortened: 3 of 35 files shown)


dominikklein:
Don't know if this is the way to go. In the end, we should first identify the problem, and for me it's more related to whether you want to use the feature or not.

E.g., on support tickets, you want to have the integration, but on sales tickets, it's not needed. But when a sales user has also access to a support ticket, he should still see the sidebar.

YetAnotherGerrit:
I agree with Dominik. Rather than being solved through permissions, I would love to see this issue resolved based on ticket criteria. Perhaps this could be achieved by extending Core Workflows to hide the tabs?

dominikklein:
Core Workflow is not designed for the UI :-) This would be a massive change, nothing for a PR, and I would also say, not for now (in both views).

What we also need to have in mind: Such a hide mechanism could influence the loading feeling of tickets, because we need to have many more checks (e.g., we want to go in a similar direction for the summary).

mantas:
Maybe it could be an extension to integration UIs by selecting groups. Maybe that'd be cleanest.

paranmaum:
Thank you all for the quick feedback!

In my case, we have multiple departments. Each department has it's own agent role and group; some also have subgroups.

>E.g., on support tickets, you want to have the integration, but on sales tickets, it's not needed. But when a sales user has also access to a support ticket, he should still see the sidebar.

That would be perfect. My issue is that currently agents in the sales department (sticking with the example) are exposed to some data that they normally don't have access to through the i-doit integration, as only the support department uses that system.

I do agree that instead of adding this as another user/role permission, deciding access per-group or through some sort of condition based system would be the better solution. I do think that both of these routes would be much bigger changes that I probably cannot contribute. Should I close this PR and file an issue instead then?

YetAnotherGerrit:
Thank you very much for your contribution, even though we have decided to take a different approach. Please open a feature request in [our community](https://community.zammad.org/).

PR #6196, opened 2026-06-11
Fix checklist template item order

HanSolo72:
## Description

When creating a checklist from a checklist template, the template items were copied via the raw `template.items` association.

This association does not guarantee the configured item order and may return items in database order. As a result, checklist items could be copied to tickets in the wrong order, even though `sorted_item_ids` contained the correct order.

This change uses `template.sorted_items` instead, so the configured checklist template order is preserved when creating a checklist from a template.

## Testing

Manually tested on a local Zammad instance:

* Verified that `template.items.pluck(:id)` returned the unsorted database order.
* Verified that `template.sorted_item_ids` contained the expected item order.
* Verified that `template.sorted_items.pluck(:id)` returned the correct order.
* Created a checklist from the template and confirmed that the checklist items were copied to the ticket in the expected order.

## Issue reference

No existing issue found for this specific problem.

## Notes

The pull request targets the `develop` branch.

### DIFF
--- a/app/models/checklist.rb
+++ b/app/models/checklist.rb
@@ -105,10 +105,12 @@ class Checklist < ApplicationModel
     ActiveRecord::Base.transaction do
       Checklist.create!(name: template.name, ticket:)
         .tap do |checklist|
-          sorted_item_ids = template
-            .items
-            .map { |elem| checklist.items.create!(text: elem.text, initial_clone: true) }
-            .pluck(:id)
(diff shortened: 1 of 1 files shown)


dominikklein:
Thanks for the contribution, maybe you can also report a related bug issue, so that we have an reference for this problem?

mantas:
Hey @HanSolo72 , thank you for your contribution ❤️ It is now merged, packages with the fix will be available soon.

PR #6250, opened 2026-07-16
feat: Ticket Participants — let customers share ticket visibility (fixes #1307)

### LINKED ISSUE #1307: Ticket should be visible for more than one customer
### Infos:

* Used Zammad version: 1.6.1
* Used Zammad installation source: (source, package, ...)
* Operating system: 
* Browser + version: 

### Expected behavior:
In some conditions it would be useful that an additional customer could view a ticket.
So it would be cool if you could either assign a ticket to multiple customers or add a "viewer-field" to the ticket where additional customers who should be also able to view the ticket could be added.

usecases:
- two users are doing desksharing. User of the early shift reports the issue but it can´t be solved until the end of the shift. So the user of the late shift should see the ticket with his account and work with the agent on the solution. Next day the user of the morning shift should be able to view what happened after he went home the day before.
- in some conditions the boss wan´t to view what´s happening with a ticket his employee opened.
- the employee in financial department wants to get some special rights in the IT system. So the leader  of financial department opens a ticket for group usermanagement to give the employee the rights. Then it would be useful to add the affected employee to the ticket so boss and employee get the same information about ticket progress.

### Actual behavior:
not available.

addipoes:
## Description

Implements the feature requested in #1307 (2016): A viewer-field where additional customers can be added to a ticket, giving them read access without being the ticket owner.

### What this does
- Agents can add other customers as **participants** to a ticket via the sidebar (Desktop UI + old jQuery UI)
- Participants get **read-only** access to the ticket (FieldScope, same as ticket customer)
- Participants receive email notifications for ticket updates (standard Subscriber-pipeline)
- **Internal articles** (agent notes) are never sent to participants (internal-article guard)
- Cap of 50 participants per ticket
- Feature-flag gated (`ticket_participants_enabled`, default: off → zero impact when disabled)

### Use Cases (from #1307)
- Desksharing: two users share a ticket across shifts
- Manager wants to see progress on employee ticket
- Affected employee added to IT ticket so they get the same information

### Architecture
- **Backend:** Uses Zammad existing `mentions` table. New `Mention.subscribe!` with policy checks.
- **Authorization:** `Controllers::MentionsControllerPolicy` with `agent_update_access?` — symmetric between REST and GraphQL.
- **Notification:** Standard `Transaction::Notification` Subscriber-pipeline + internal-article guard (`article.internal? && !user.permissions?("ticket.agent")` → skip).
- **Frontend:** Vue Composition API (Desktop UI) + CoffeeScript/jQuery (old UI).
- **No new dependencies.** No database changes beyond a settings migration.

### Changes (5 commits)
1. `feat: Backend` — Policy, Scope, participant?, participant_ids, mentionable?, Cap 50, 23 policy specs
2. `feat: GraphQL` — `ticketParticipantAdd` / `ticketParticipantRemove` mutations with `loads: UserType`
3. `feat: Notification` — `mention_user_eligible?`, internal-article guard, participant-added D5 mail, 11 notification specs
4. `feat: Vue Desktop UI` — Participants Sidebar (Add/Remove, autocomplete via GraphQL, badge)
5. `feat: Old UI + REST` — CoffeeScript widget, REST endpoints, participant badge

### Testing
- **420 RSpec examples** (policy, notification, REST, GraphQL, model) — all green
- **Browser-verified** (Playwright) — full Add/Search/Remove chain + Mailcatcher D5 notification
- **Security-audited:** Self-subscribe guard (G1), internal-article guard (IG1-3), cap enforcement (CAP1-6), cross-ticket auth, active-user filter (AF1-6)

### Backward Compatibility
- Feature flag OFF by default (`ticket_participants_enabled: false`)
- All existing behavior unchanged when flag is off
- No breaking changes to the `mentions` table or its existing usage


dominikklein:
@addipoes Thanks for your contribution. I only did a small look, but I think the direction is not like we would see such a feature in the core product.
It's always difficult to bring such more complex features via PR in the product, because we first need to have a picture of how we see this in the core for the long run. It needs to match for the long run.

addipoes:
Thanks for taking a look, Dominik! Totally understand — strategic fit needs to come before code.

To help us understand the direction: is ticket-sharing between customers something Zammad might want in the product long-term, or is this fundamentally outside the product vision? If it is something you see value in, what would the right approach look like from your side — a design discussion on the forum first, or a different architecture?

Happy to restructure, discuss on community.zammad.org, or rebuild as an addon if that fits better. Just want to make sure we are not investing in a dead end.

dominikklein:
In the end, it's really connected to: https://community.zammad.org/t/allow-multiple-customers-to-be-assigned-to-a-ticket/9549/43

So we know it's one of the more requested features from the community. So it's definitely something that we also see in the core product, but currently there is no "implementation idea" present.
It's also a feature we will no longer implement in the old desktop UI, so it's maybe something that will be postponed after the new desktop UI is ready (at least from the customer/agent perspective, without the admin interface).

PR #6264, opened 2026-07-25
Feature: Ticket approval workflow

jhmk:
Hi, zammad is really nice but we are missing a feature to get approvals.
When a User need Hardware or access they need to get the approval from their Lead. 
At the moment this is missing in zammad, so we can't really use it yet. 

## What it adds

- New ticket type "Approval Request". The approval feature is scoped to this ticket type; it doesn't clutter regular tickets.

- Approver field (User select) — the person asked to approve. Only shown for the "Approval Request" type.
Approval State field — Not requested / Pending approval / Approved / Rejected. Only shown for the "Approval Request" type.

- Approver is scoped to the requester's organization — you can only pick someone from your own organization (via Core Workflow).

- Approval State is editable only by the assigned approver or an agent — everyone else sees it read-only (via Core Workflow).

- Email notification to the approver when an approval-request ticket is created (Trigger + a new custom-field recipient resolver).

- Timeline note added automatically whenever the approval state changes, recording who changed it and to what (Trigger).

- "Waiting for my Approval" overview listing tickets pending the current user's decision.

Request: https://community.zammad.org/t/get-approval-for-tickets-by-lead/20745

Bildschirmfoto 2026-07-25 um 16 45 00

Bildschirmfoto 2026-07-25 um 16 45 45

Bildschirmfoto 2026-07-25 um 16 46 09

Bildschirmfoto 2026-07-25 um 16 54 09

### DIFF
--- a/app/models/core_workflow/attributes/user.rb
+++ b/app/models/core_workflow/attributes/user.rb
@@ -9,7 +9,45 @@ class CoreWorkflow::Attributes::User < CoreWorkflow::Attributes::Base
       return ticket_owner_id
     end
 
+    if @attribute[:name] == 'approver' && @attributes.payload['class_name'] == 'Ticket'
+      return ticket_approver_ids
+    end
+
     []
+  end
+
+  # Approvers are limited to active members of the ticket requester's
+  # organization, so an employee can only request approval from someone within
+  # their own organization.
+  def ticket_approver_ids
+    organization = ticket_approver_organization
+    return [''] if organization.blank?
+
+    user_ids = organization.members.where(active: true).each_with_object([]) do |user, result|
+      result << user.id
+      assets(user)
+    end
+
+    return [''] if user_ids.blank?
+
+    user_ids
+  end
+
+  def ticket_approver_organization
+    requester = ticket_approver_requester
+    requester&.organization
+  end
+
+  def ticket_approver_requester
+    params = @attributes.payload['params'] || {}
+
+    if params['customer_id'].present? && @attributes.user.permissions?('ticket.agent')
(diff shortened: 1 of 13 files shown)


dominikklein:
Thanks for your contribution, but I think this can already done with existing features like custom object attributes, core workflow and trigger/macros without adding it directly in the core.

Maybe in the future we can think about of something like process templates which can be imported.

jhmk:
> Thanks for your contribution, but I think this can already done with existing features like custom object attributes, core workflow and trigger/macros without adding it directly in the core.
> 
> Maybe in the future we can think about of something like process templates which can be imported.

I tired but was unable to get it proper working. If you have good way please let me know. 
My intention was to have a far more easier way to use it.

YetAnotherGerrit:
Approval workflows are a common requirement that our consultants encounter in their workshops. Many of these can be met through the clever use of triggers, custom attributes and core workflows. However, one thing we have learned about approval workflows in recent years is that the way they should be handled varies greatly from one use case to another. This makes it extremely difficult to build a one-size-fits-all solution into the core product.

Rather than incorporating a solution that only caters for some use cases, I would prefer to enhance the capabilities of the existing tools to cover more use cases. We have done this in the past by increasing the capabilities of the available conditions and actions on triggers, filters and core workflows. We have also increased the capabilities of custom ticket attributes. And we will continue to do so.

There may be a dedicated approval workflow engine in Zammad in the future. However, we will need to ensure that it covers a wide variety of use cases and is much easier to set up than manually crafting everything with triggers, custom attributes, and core workflows.

Still, a big thank you for your contribution! <3

PR #6266, opened 2026-07-25
Maintenance: Do not truncate the GraphQL introspection dump on a failed generate.

### LINKED ISSUE #6265: pnpm generate-graphql-api empties the tracked introspection dump when the generator fails
### Used Zammad Version

7.2.x (develop)

### Environment

- Installation method: source (development checkout)
- Operating system: any
- Database + version: n/a — the failure happens before the database is reached
- Elasticsearch version: n/a
- Browser + version: n/a

### Actual behaviour

`package.json` defines:

```
"generate-graphql-api": "bundle exec rails generate zammad:graphql_introspection > app/graphql/graphql_introspection.json && pnpm exec graphql-codegen -c .graphql_code_generator.js"
```

A shell sets up the `>` redirection *before* it runs the command, opening the target with `O_CREAT|O_TRUNC`. `app/graphql/graphql_introspection.json` is therefore emptied before `bundle exec rails` even starts — regardless of whether that command succeeds, or exists at all.

The `&&` does not protect the file. It only stops the following codegen step, long after the truncation has happened.

So any failure of the generator leaves a tracked 1.1 MB file at 0 bytes. Triggers include a checkout where `bundle install` has not run yet, a Ruby version not matching `.ruby-version`, an unreachable database, or a syntax error in a GraphQL type.

Two things make this easy to miss:

- Nothing reports the data loss. The visible output is the generator's own error message.
- If the truncation goes unnoticed, the next symptom is usually an unrelated-looking codegen or type error about an invalid schema, which points at the wrong cause.

A subsequent `git add -A` then commits the empty file.

### Expected behaviour

`app/graphql/graphql_introspection.json` is only replaced once the generator has completed successfully. A failed run leaves the tracked file untouched.

In a fresh checkout, before `bundle install` has been run:

1. `wc -c app/graphql/graphql_introspection.json` → ~1.1 MB
2. `pnpm generate-graphql-api`
3. `wc -c app/graphql/graphql_introspection.json` → **0**

The same effect can be shown without Zammad at all, which isolates the cause to the redirection rather than to anything in the generator:

```
$ echo 'important data' > tracked.json && wc -c < tracked.json
16
$ this-command-does-not-exist > tracked.json ; echo "exit=$?"
exit=127
$ wc -c < tracked.json
0
```

The command never ran — the file is empty anyway.

### Support Ticket

_No response_

yes

### Suggested fix

Stage the dump through a temporary file and only move it into place once the generator has succeeded:

```
"generate-graphql-api": "mkdir -p tmp && bundle exec rails generate zammad:graphql_introspection > tmp/graphql_introspection.json && mv tmp/graphql_introspection.json app/graphql/graphql_introspection.json && pnpm exec graphql-codegen -c .graphql_code_generator.js"
```

`tmp/` is already covered by `.gitignore` (`/tmp/*`), so the truncation now hits a disposable file, and `mv` within the same filesystem is an atomic `rename(2)`.

To be explicit about the limits: this guards against the generator *failing*, not against it exiting 0 while writing invalid output. The latter seemed out of scope for a one-line change.

I have this ready as a single-commit branch and am happy to open a PR.

kiliant:
Fixes #6265.

`generate-graphql-api` redirected straight into the tracked
`app/graphql/graphql_introspection.json`, so the shell truncated it to zero bytes before the
generator ran. Any failure — a checkout without `bundle install`, a mismatched Ruby version,
an unreachable database — left the tracked 1.1 MB file empty, with nothing reporting the loss.

The dump is now staged through `tmp/` and only moved into place once the generator has exited
successfully. `tmp/` is already covered by `.gitignore` (`/tmp/*`), and `mv` within the same
filesystem is an atomic `rename(2)`.

This guards against the generator *failing*, not against it exiting 0 while writing invalid
output — that seemed out of scope here, but happy to extend it if you would rather have a
sanity check on the dump.

No tests: the change is a single line in a package script. Verified by running
`pnpm generate-graphql-api` both in a working environment (dump regenerated, unchanged output)
and in a deliberately broken one (generator fails, tracked file untouched).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "generate-graphql-api": "bundle exec rails generate zammad:graphql_introspection > app/graphql/graphql_introspection.json && pnpm exec graphql-codegen -c .graphql_code_generator.js",
+    "generate-graphql-api": "mkdir -p tmp && bundle exec rails generate zammad:graphql_introspection > tmp/graphql_introspection.json && mv tmp/graphql_introspection.json app/graphql/graphql_introspection.json && pnpm exec graphql-codegen -c .graphql_code_generator.js",
     "generate-setting-types": "bundle exec rails generate zammad:setting_types",
     "dev": "RAILS_ENV=development forego start -r -f Procfile.dev",
     "dev:https": "VITE_RUBY_HOST=0.0.0.0 VITE_RUBY_HTTPS=true RAILS_ENV=development forego start -r -f Procfile.dev-https",


vBenTec:
Hey @kiliant, thanks for your contribution. We just made a small tweak to drop  the `mkdir -p tmp` part as it is always present.

We cherry‑picked your commit into an internal process and merged it to develop.
https://github.com/zammad/zammad/commit/8646981689a57c1fec687dd9d36e18f995c16d1a

We look forward to future contributions 🙂.

Issue #5984, opened 2026-03-05
Ollama models with thinking enabled may return empty response field


bialaps:
### Used Zammad Version

7.0

### Environment

- Installation method: [e.g. source, package]
- Operating system (if you're unsure: `cat /etc/os-release` ): [e.g. debian 10.4, ubuntu 20.04]
- Database + version:  [e.g. postgresql 9.3]
- Elasticsearch version: [e.g. 7.17]
- Browser + version: [e.g. chrome 83, safari 14, firefox 105]

### Actual behaviour

Ollama reasoning models may return `The response could not be processed` because the response field is empty while the generated output appears in the thinking field.

This behavior is described here: https://ollama.com/blog/thinking

### Expected behaviour

The model's generated content should always be present in the response field.

### Steps to reproduce the behaviour

Configure Ollama to use a reasoning/thinking model (e.g. `qwen3`). 
Observe that the API response contains: 
· An empty response field
· The generated JSON output appears in the thinking field

### I'm sure this is a bug and no feature request or a general question.

yes

drstefanmeier:
I can confirm this behaviour for llama4 as well as glm-4.7-flash. For llama4 I also saw this response:

`
{
  "model": "llama4:latest",
  "created_at": "2026-03-04T22:48:04.61165421Z",
  "response": "{
  \"summary\": \"<REDACTED FOR PRIVACY REASONS>\",
  \"description\": \"<REDACTED FOR PRIVACY REASONS>\",
  \"keywords\": [
    \"<REDACTED FOR PRIVACY REASONS>\",
    \"Upgrade\",
    \"Configuration\",
    \"Testing\",
    \"Implementation\"
  ]
},...
`

dvuckovic:
@drstefanmeier That response does not seem to be related to te the reported issue. It looks like in your case LLM decided *not* to follow instructions, since the reponse structure is completely off (hallucination, perhaps?).

Please check your Ollama configuration, especially regarding the allowed context window, etc. As you can imagine, we cannot really give exact guidance for this, please consult vendor documentation for details.

PR #6298, opened 2026-08-13
Fixes #6297 - Anthropic provider fails on Claude models that return thinking blocks

### LINKED ISSUE #6297: AI: Anthropic provider fails on Claude models that return thinking blocks
### Used Zammad Version

7.1.2

### Environment

- Installation method: Kubernetes (zammad Helm chart)
- Operating system: Debian (official zammad container image)
- Database + version: PostgreSQL 15
- Elasticsearch version: 8
- Browser + version: not relevant, this is server side

### Actual behaviour

With the Anthropic provider set to a Claude 5 model such as `claude-sonnet-5`, AI features fail intermittently. The UI shows "The summary could not be generated. Please try again later or contact your administrator. API server error: The response could not be processed."

The HTTP request itself succeeds. The entry in Monitoring / HTTP log with facility `AI::Provider` shows status 200 and a valid body. The problem is in parsing.

Claude 5 models have adaptive thinking enabled by default, so the response contains a `thinking` content block before the `text` block:

```json
{"content": [
  {"type": "thinking", "thinking": "", "signature": "Et..."},
  {"type": "text", "text": "{\"language\": \"en-us\", ...}"}
], "stop_reason": "end_turn"}
```

`lib/ai/provider/anthropic.rb` extracts the answer with:

```ruby
data['content'].first['text']
```

When a thinking block comes first this returns `nil`, which then fails in the service layer. The failure is intermittent because the model does not emit a thinking block on every request. In our staging system 5 of 6 requests to `claude-sonnet-5` contained one, while `claude-sonnet-4-6` and `claude-haiku-4-5` never did.

All AI features are affected, including ticket summaries, text tools and AI agents. Because it is intermittent, a ticket can silently go unclassified or unsummarized at random on an otherwise correctly configured system.

`DEFAULT_OPTIONS[:models_without_temperature]` in the same file already lists `claude-fable-5`, `claude-opus-5` and `claude-sonnet-5`, so these models are expected to be selectable.

This is the same class of problem as #5984 for the Ollama provider, which was solved by sending `think: false`. That approach does not translate to Anthropic: disabling thinking is not uniformly supported across the Claude 5 family. On Claude Fable 5 an explicit `thinking: {"type": "disabled"}` returns 400 at any effort level, and on Claude Opus 5 it is rejected when effort is `xhigh` or `max`. Tolerating the response shape is therefore the more robust option.

### Expected behaviour

The provider extracts the model answer regardless of whether the response contains thinking blocks.

1. Configure Anthropic as the AI provider and set the model to `claude-sonnet-5`.
2. Open a ticket and request an AI summary several times.
3. Some attempts fail with "The response could not be processed".
4. Monitoring / HTTP log, facility `AI::Provider`, shows status 200 with a `thinking` block first in the response body.

### Proposed fix

Select the first content block of type `text` instead of the first block:

```diff
-    data['content'].first['text']
+    data['content'].find { |block| block['type'] == 'text' }&.fetch('text', nil)
```

This is backwards compatible. For a response without thinking blocks the first block is the text block, so behaviour for existing models is unchanged.

Verified against the live API on 7.1.2 by prepending a patched `chat` method in a `rails runner` process and running `AI::Service::TicketSummarize` three times with `claude-sonnet-5`. All three responses contained a thinking block, the current code extracted `nil` in all three, and the patched code returned the summary correctly in all three.

A pull request follows.

yes

bdriton13:
Fixes #6297

Claude 5 models have adaptive thinking enabled by default and return a `thinking` content block before the `text` block. `data['content'].first['text']` then evaluates to `nil` and the service layer reports "The response could not be processed", even though the API call returned 200.

This selects the first content block of type `text` instead of the first block. For responses without thinking blocks the first block is the text block, so behaviour for existing models such as `claude-sonnet-4-6` and `claude-haiku-4-5` is unchanged.

Disabling thinking, as done for the Ollama provider in #5984 with `think: false`, is not a viable equivalent here: on Claude Fable 5 an explicit `thinking: {"type": "disabled"}` returns 400 at any effort level, and on Claude Opus 5 it is rejected at effort `xhigh` or `max`. Both models are already listed in `DEFAULT_OPTIONS[:models_without_temperature]`.

Added a spec that stubs a response containing a thinking block and asserts the text block is returned.

Verified against the live API on 7.1.2 by prepending the patched `chat` method in a `rails runner` process and running `AI::Service::TicketSummarize` three times with `claude-sonnet-5`. All three responses contained a thinking block, the previous code extracted `nil` in all three, and the patched code returned the summary correctly in all three.

### DIFF
--- a/lib/ai/provider/anthropic.rb
+++ b/lib/ai/provider/anthropic.rb
@@ -119,7 +119,9 @@ class AI::Provider::Anthropic < AI::Provider
     data = validate_response!(response)
     extract_response_metadata(data)
 
-    data['content'].first['text']
+    # Models with extended thinking return one or more 'thinking' blocks before
+    # the 'text' block, so the answer is not necessarily the first element.
+    data['content'].find { |block| block['type'] == 'text' }&.fetch('text', nil)
   end
 
   def embeddings(input:)
--- a/spec/lib/ai/provider/anthropic_spec.rb
+++ b/spec/lib/ai/provider/anthropic_spec.rb
@@ -64,6 +64,26 @@ RSpec.describe AI::Provider::Anthropic, integration: true, required_envs: %w[ANT
       end
     end
 
+    context 'when the response contains thinking blocks' do
+      it 'returns the content of the text block' do
+        allow(UserAgent).to receive(:post).and_return(
+          UserAgent::Result.new(
+            success: true,
+            code:    200,
+            data:    {
+              'content' => [
+                { 'type' => 'thinking', 'thinking' => '', 'signature' => 'Et...' },
+                { 'type' => 'text', 'text' => '{ "connected": "true" }' },
+              ],
+              'usage'   => { 'input_tokens' => 1, 'output_tokens' => 1 },
+            },
+          )
+        )
+
+        expect(ai_provider.ask(prompt_system:, prompt_user:)).to match({ 'connected' => 'true' })
+      end
+    end
+
     context 'when metadata is extracted' do
       it 'stores metadata from response' do
         ai_provider.ask(prompt_system:, prompt_user:)


dominikklein:
Thank you for the detailed report and the well-scoped fix, @bdriton13! This has been merged into develop and cherry-picked into our stable branch, so it will be available in the next stable release soon.

PR #6346, opened 2026-09-03
perf: upgrade to vitest 5 and make tests 4 times faster

sheremet-va:
Hello there 👀 

This PR updates vitest to v5 and changes the `pool` to `vmForks`. With this I was able to reduce the time it takes to run tests (on my machine) from 250s to 69s (64s with cache) 👀 

The main requirement for this is that `window.navigation` cannot be overriden anymore - `jsdom` makes it nonconfigurable. To bypass this, I added a `#shared/utils/navigation.ts` module (maybe it's also a good idea to _always_ mock it, but in this PR it's only mocked in tests that require it).

### DIFF
--- a/app/frontend/tests/support/mock-vue-router.ts
+++ b/app/frontend/tests/support/mock-vue-router.ts
@@ -1,13 +1,16 @@
 // Copyright (C) 2012-2026 Zammad Foundation, https://zammad-foundation.org/
 
-export const mockRouterHooks = () => {
-  vi.mock('vue-router', async () => {
-    const module = await vi.importActual<typeof import('vue-router')>('vue-router')
+vi.mock('vue-router', async () => {
+  const module = await vi.importActual<typeof import('vue-router')>('vue-router')
 
-    return {
-      ...module,
-      onBeforeRouteUpdate: vi.fn(),
-      onBeforeRouteLeave: vi.fn(),
-    }
-  })
-}
+  return {
+    ...module,
+    onBeforeRouteUpdate: vi.fn(),
+    onBeforeRouteLeave: vi.fn(),
+  }
+})
+
+// this function never did anything, `vi.mock` is always executed as the first statement,
+// but vitest 5 now fails with a hard error if it sees `vi.mock` being called ouside of the top-level scope.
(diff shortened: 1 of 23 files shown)


pnpm-workspace.yaml
```diff
@@ -4,4 +4,10 @@ allowBuilds:
   esbuild: true
   unrs-resolver: true
   vue-demi: true
+minimumReleaseAgeExclude:
```
sheremet-va, line 7: This is just because vitest was _just_ released

app/frontend/tests/vitest.setup.ts
```diff
@@ -282,13 +282,13 @@ afterEach((context) => {
...
-// There is a problem that sa11y uses still vitest v3
+// @sa11y/vitest still depends on older Vitest APIs, so keep explicit registration
+// until it officially supports Vitest 5.
 // https://github.com/salesforce/sa11y/blob/master/packages/vitest/package.json
-// In vitest v.4 we still need to manually provide the assertion api
 expect.extend({ toBeAccessible })
 
-expect.extend(assertions)
-expect.extend(domMatchers)
+expect.extend(assertions as unknown as MatchersObject)
```
sheremet-va, line 290: Vitest is more strict with what you can pass, it's not shared with the `Matchers` type

app/frontend/tests/support/mock-vue-router.ts
```diff
@@ -1,13 +1,16 @@
...
+    ...module,
+    onBeforeRouteUpdate: vi.fn(),
+    onBeforeRouteLeave: vi.fn(),
+  }
+})
+
+// this function never did anything, `vi.mock` is always executed as the first statement,
+// but vitest 5 now fails with a hard error if it sees `vi.mock` being called ouside of the top-level scope.
+// keeping the function for backwards compatibility, but it is not needed anymore
+export const mockRouterHooks = () => undefined
```
sheremet-va, line 16: Vitest 4 already printed a warning about this

vBenTec:
Thanks for you contribution @sheremet-va ! 
https://github.com/zammad/zammad/commit/4e489121c52fb644506843258c8f693cdcd89cdc

PR #6358, opened 2026-09-12
Fix S/MIME decryption for forwarded emails

### LINKED ISSUE #6357: S/MIME decryption fails for forwarded emails when recipient is only present in Delivered-To
### Used Zammad Version

7.1.3-0011; also reproduced on current `develop` (d9c5ba05f3).

### Environment

Installation method: Docker Compose

The issue was originally observed on Zammad 7.1.3-0011 and independently reproduced using the official Zammad development environment on current `develop` (d9c5ba05f3).

### Actual behaviour

Incoming S/MIME-encrypted messages cannot be decrypted when the address for which the message was encrypted is present in the `Delivered-To` header, but not in `To` or `Cc`.

This can occur with forwarded messages where the original recipient is preserved in the `To` header, for example:

```text
To: forwarding-address@example.com
Delivered-To: smime-recipient@example.com
```

The S/MIME message is encrypted for the certificate belonging to `smime-recipient@example.com`, and the corresponding certificate and private key are available in Zammad.

Zammad nevertheless reports:

```text
The private key for decryption could not be found.
```

The same message can be successfully decrypted by another S/MIME-capable mail client using the same private key.

The issue can be reproduced on current `develop`. `SecureMailing::SMIME::Incoming#decryption_certificates` currently selects candidate private keys only from addresses in `To` and `Cc`. Consequently, no suitable candidate is found when the actual S/MIME recipient is only available in `Delivered-To`.

An isolated regression test on unchanged `develop` reproduces the issue: the encrypted body remains unavailable and the test fails with:

```text
expected "no visible content" to include "Some text"
```

### Expected behaviour

Zammad should successfully decrypt an incoming S/MIME-encrypted message when the required private key is available and the actual recipient is provided in the `Delivered-To` header, even if that address is not present in `To` or `Cc`.

Existing private-key selection based on `To` and `Cc` should continue to work unchanged.

1. Configure an S/MIME certificate and its corresponding private key for `smime-recipient@example.com` in Zammad.

2. Create a valid S/MIME-encrypted message for `smime-recipient@example.com`.

3. After encryption, change the message headers so that the actual encryption recipient is no longer present in `To` or `Cc`, while preserving it in `Delivered-To`, for example:

```text
To: forwarding-address@example.com
Delivered-To: smime-recipient@example.com
```

This represents a forwarding scenario in which the original `To` header is preserved while the message is ultimately delivered to `smime-recipient@example.com`.

4. Process the message as an incoming email in Zammad.

5. Observe that Zammad does not decrypt the message and reports:

```text
The private key for decryption could not be found.
```

The issue is also reproducible using the existing S/MIME test infrastructure in `spec/lib/secure_mailing/smime_spec.rb`.

A regression test for this scenario fails on unchanged current `develop`. After including `Delivered-To` in the candidate certificate lookup, the same test passes, and the complete S/MIME spec passes with 64 examples and 0 failures.

### Support Ticket

_No response_

yes

reger-p:
Fixes #6357

### Summary

Fix S/MIME decryption for forwarded messages where the actual S/MIME recipient is present in `Delivered-To` but not in the `To` or `Cc` headers.

`SecureMailing::SMIME::Incoming#decryption_certificates` currently selects candidate private keys using only `To` and `Cc`. This can fail after forwarding because these headers may retain the original recipient while the final mailbox address is provided in `Delivered-To`.

This change adds `Delivered-To` to the candidate certificate lookup and de-duplicates the resulting certificates.

### Example

A forwarded message may contain:

```text
To: forwarding-address@example.com
Delivered-To: smime-recipient@example.com
```

while the S/MIME envelope itself is encrypted for the certificate belonging to `smime-recipient@example.com`.

Before this change, Zammad reports:

```text
The private key for decryption could not be found.
```

With this change, the corresponding private key is found and the message is successfully decrypted.

### Tests

Added a regression test which:

* creates an S/MIME-encrypted message for the configured recipient;
* replaces `To` with a forwarding address;
* removes `Cc`;
* sets the actual recipient only in `Delivered-To`;
* verifies that the message is successfully decrypted.

The regression test fails on unchanged `develop` and passes with this change.

Validation:

```text
spec/lib/secure_mailing/smime_spec.rb
64 examples, 0 failures

RuboCop
2 files inspected, no offenses detected
```

### DIFF
--- a/lib/secure_mailing/smime/incoming.rb
+++ b/lib/secure_mailing/smime/incoming.rb
@@ -279,6 +279,9 @@ class SecureMailing::SMIME::Incoming < SecureMailing::Backend::HandlerIncoming
       mail[:mail_instance].cc.each { |cc| certs += ::SMIMECertificate.find_by_email_address(cc, filter: { key: 'private', usage: :encryption }) }
     end
 
-    certs
+    delivered_to = mail[:mail_instance]['Delivered-To']&.value
+    certs += ::SMIMECertificate.find_by_email_address(delivered_to, filter: { key: 'private', usage: :encryption }) if delivered_to.present?
+
+    certs.uniq
   end
 end
--- a/spec/lib/secure_mailing/smime_spec.rb
+++ b/spec/lib/secure_mailing/smime_spec.rb
@@ -846,6 +846,27 @@ RSpec.describe SecureMailing::SMIME do
           expect(mail['x-zammad-article-preferences'][:security][:encryption][:comment]).to eq(recipient_certificate_subject)
         end
 
+        context 'when recipient is only present in Delivered-To' do
+          let(:mail) do
+            smime_mail = build_mail
+
+            smime_mail.to = 'forwarding-address@example.com'
+            smime_mail.cc = nil
+            smime_mail['Delivered-To'] = recipient_email_address
+
+            mail = Channel::EmailParser.new.parse(smime_mail.to_s)
+            SecureMailing.incoming(mail)
+
+            mail
+          end
+
+          it 'decrypts' do
+            expect(mail[:body]).to include(raw_body)
+            expect(mail['x-zammad-article-preferences'][:security][:encryption][:success]).to be true
+            expect(mail['x-zammad-article-preferences'][:security][:encryption][:comment]).to eq(recipient_certificate_subject)
+          end
+        end
+
         it_behaves_like 'HttpLog writer', 'success'
 
         context 'expired allowed' do


dvuckovic:
@reger-p Merged in [`08f75cf`](https://github.com/zammad/zammad/commit/08f75cf405df93bb927b43078a0dd73a1bcb554b), thanks for your contribution!

We decided to change the parsing and handling of the `Delivered-To` header slightly, considering it can be included multiple times and contain not just email addreses.

